PR #28, opened 2026-02-15
fix(hud): harden tmux launcher against shell command injection

jaeyeom:
## Summary

The `launchTmuxPane` function built a shell command string via template literal interpolation and passed it to `execSync()`. A `cwd` containing `"`, `$()`, or backticks could escape the double-quoted context, and an `omxBin` path containing `'` could break out of the single-quoted command string — both allowing arbitrary command execution.

## Changes

- Replace `execSync` with `execFileSync` so `cwd` is passed as a literal argv element to tmux (no shell expansion)
- Add `shellEscape()` helper using POSIX single-quote wrapping to neutralise metacharacters in `omxBin`
- Extract `parseHudPreset()` whitelist validator and apply it inside `buildTmuxSplitArgs` as defense-in-depth for the preset parameter
- Add 19 tests: vulnerability reproduction, shellEscape unit tests, and buildTmuxSplitArgs injection hardening tests

## Validation

- [x] `npm run build`
- [x] `npm test` (205 tests pass, including 19 new)

## Checklist

### DIFF
--- a/src/hud/__tests__/hud-tmux-injection.test.ts
+++ b/src/hud/__tests__/hud-tmux-injection.test.ts
@@ -1,0 +1,194 @@
+/**
+ * Tests for shell command injection hardening in the tmux HUD launcher.
+ *
+ * The first describe block reproduces the vulnerability that existed when
+ * launchTmuxPane() built a command string via template-literal interpolation
+ * and passed it to execSync().  The second block verifies the hardened
+ * buildTmuxSplitArgs() + shellEscape() approach is safe.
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Follow-up applied on dev (separate from PR #28): raised DEFAULT_MAX_WORKERS to 20 (commit 1cd735a). This aligns CLI 1-20 worker count with initTeamState default max_workers and avoids surprising workerCount>maxWorkers errors.

PR #541, opened 2026-03-03
fix: security quick wins (#488, #493, #495, #496, #508, #523)

### LINKED ISSUE #488: [Security] code-intel-server: npx --yes auto-installs packages (supply chain risk)
## Summary

`src/mcp/code-intel-server.ts` line ~185 calls `execFileAsync('npx', ['--yes', '@ast-grep/cli', '--version'])` with the `--yes` flag, which auto-installs the package without user confirmation.

## Location

- `src/mcp/code-intel-server.ts:185` — ast-grep availability check

## Details

The `--yes` flag to `npx` automatically confirms installation of packages. If the npm registry is compromised or if a typosquatted package name is used, this could execute malicious code without user consent.

## Risk

- **Severity**: Medium — supply chain attack vector via auto-installed npm packages
- **Prerequisite**: Requires npm registry compromise or package name typosquatting

## Recommendation

1. Remove `--yes` flag and require ast-grep to be pre-installed
2. Or check if the package is already installed before running with `--yes`
3. Or pin the exact version to reduce risk

### LINKED ISSUE #493: [Security] mcp/team-server.ts: Unvalidated sessionName passed to tmux spawn
## Summary

`listLiveSessionPaneIds` passes an unvalidated `sessionName` string directly to `spawn('tmux', ['list-panes', '-t', sessionName, ...])`.

## Location

- `src/mcp/team-server.ts` — lines 109–129

## Code

```typescript
async function listLiveSessionPaneIds(sessionName: string): Promise<string[]> {
  if (!sessionName || !sessionName.trim()) return [];
  return await new Promise((resolve) => {
    const child = spawn('tmux', ['list-panes', '-t', sessionName, '-F', '#{pane_id}'], {
```

## Problem

`sessionName` is read from the team config file (`config.tmux_session`). While `spawn` (not `exec`) is used so direct shell injection is not possible, tmux's `-t` flag accepts complex target expressions that could target arbitrary sessions. A crafted session name like `omx-team-foo:1.0` could reference specific windows/panes in other sessions.

If an attacker can write to the team config (e.g., via a path traversal in `state_write`), they can inject a crafted session name to target or inspect other tmux sessions.

## Recommended Fix

Validate `sessionName` against a strict pattern (`/^[a-zA-Z0-9_-]+$/`) before passing to tmux.

## Severity

**High** — Requires config write access, but allows targeting arbitrary tmux sessions.

### LINKED ISSUE #495: [Security] mcp/state-paths.ts: OMX_TEAM_STATE_ROOT env var bypasses working directory allowlist
## Summary

The `OMX_TEAM_STATE_ROOT` environment variable can be used to redirect all state file reads/writes to an arbitrary filesystem location, bypassing the `OMX_MCP_WORKDIR_ROOTS` allowlist.

## Location

- `src/mcp/state-paths.ts` — lines 141–142

## Code

```typescript
export function getBaseStateDir(workingDirectory?: string): string {
  if ((workingDirectory == null || workingDirectory === '') &&
      typeof process.env.OMX_TEAM_STATE_ROOT === 'string' &&
      process.env.OMX_TEAM_STATE_ROOT.trim() !== '') {
    return resolveWorkingDirectoryForState(process.env.OMX_TEAM_STATE_ROOT.trim());
  }
  return join(resolveWorkingDirectoryForState(workingDirectory), '.omx', 'state');
}
```

## Problem

When `workingDirectory` is empty/null, `OMX_TEAM_STATE_ROOT` is used directly. While `resolveWorkingDirectoryForState` calls `enforceWorkingDirectoryPolicy` (which checks against `OMX_MCP_WORKDIR_ROOTS`), `OMX_TEAM_STATE_ROOT` itself is an environment variable. If an attacker can set it (e.g., via a hook that modifies the environment, a `.env` file, or CI/CD config), they can point state reads/writes to arbitrary filesystem locations.

This could be exploited to:
1. Read sensitive files by pointing state dir to a directory containing them
2. Write state files to sensitive locations (e.g., overwriting configs)
3. Bypass audit trails by redirecting state to a location the attacker controls

## Recommended Fix

Either:
1. Apply the allowlist check to `OMX_TEAM_STATE_ROOT` as well
2. Document that `OMX_TEAM_STATE_ROOT` is a trusted-admin-only variable and validate it more strictly
3. Remove the env var override and require explicit `workingDirectory` parameter

## Severity

**High** — Requires environment control, but bypasses existing security boundaries.

### LINKED ISSUE #496: [Security] cli/doctor.ts: execSync with shell string form for codex --version
## Summary

`execSync('codex --version 2>/dev/null')` uses shell string form instead of `execFileSync`, creating a shell injection surface.

## Location

- `src/cli/doctor.ts` — line 386

## Code

```typescript
const version = execSync('codex --version 2>/dev/null', {
  encoding: 'utf-8'
}).trim();
```

## Problem

`execSync` with a string argument invokes `/bin/sh -c`. The `2>/dev/null` shell redirection requires this form. If `PATH` is attacker-controlled (e.g., a malicious `codex` binary earlier in PATH), arbitrary code runs. More critically, the shell string form means any future change that interpolates a variable here would be immediately injectable.

## Recommended Fix

Use `execFileSync('codex', ['--version'], { stdio: ['pipe', 'pipe', 'pipe'] })` and handle stderr separately. For the redirect, simply ignore stderr in the caller.

## Severity

**High** — Low exploitability in practice (requires PATH control), but this is a fundamentally unsafe pattern that should be replaced.

### LINKED ISSUE #508: [Code Quality] team/orchestrator.ts: Switch statements without default — implicit undefined return
## Summary

`getPhaseAgents` and `getPhaseInstructions` have switch statements without `default` cases, causing implicit `undefined` returns when `TeamPhase` gains new values.

## Location

- `src/team/orchestrator.ts` — lines 135, 153

## Code

```typescript
export function getPhaseAgents(phase: TeamPhase): string[] {
  switch (phase) {
    case 'team-plan': return [...];
    case 'team-prd':  return [...];
    case 'team-exec': return [...];
    case 'team-verify': return [...];
    case 'team-fix':  return [...];
    // NO DEFAULT — returns undefined if TeamPhase gains a new value
  }
}
```

## Problem

TypeScript exhaustiveness checking only works with `never` in the default case. Without a `default`, if `TeamPhase` is extended (e.g., `'team-review'`), the function returns `undefined` at runtime while the declared return type is `string[]`. Callers will crash when iterating the result.

## Recommended Fix

```typescript
default:
  throw new Error(`Unknown team phase: ${phase satisfies never}`);
```

This provides both compile-time exhaustiveness checking (`satisfies never`) and runtime safety.

### LINKED ISSUE #523: [Security] package.json: @modelcontextprotocol/sdk version floor allows vulnerable 1.x versions
## Summary

The MCP SDK dependency is declared as `^1.0.0`, which allows resolution to vulnerable versions below 1.26.0. Three known CVEs affect earlier 1.x versions.

## Location

- `package.json` — line 49: `"@modelcontextprotocol/sdk": "^1.0.0"`

## Known Vulnerabilities in @modelcontextprotocol/sdk

| CVE | Advisory | Issue | Fixed In |
|-----|----------|-------|----------|
| CVE-2025-66414 | GHSA-w48q-cv73-mx4w | DNS rebinding protection default | 1.24.0 |
| CVE-2026-0621 | GHSA-cqwc-fm46-7fff | UriTemplate ReDoS | 1.25.2 |
| CVE-2026-25536 | GHSA-345p-7cg4-v4c7 | Cross-client leakage via server/transport reuse | 1.26.0 |

## Problem

While the current `package-lock.json` resolves to `1.26.0` (patched), the `^1.0.0` floor means:
1. Fresh `npm install` without a lockfile could resolve to a vulnerable version
2. `npm update` respects the `^1.0.0` range — not a direct risk but poor signaling
3. Downstream consumers of this package inherit the permissive range

## Recommended Fix

```json
"@modelcontextprotocol/sdk": "^1.26.0"
```

This ensures the minimum installed version includes all three security patches.

junhoyeo:
## Summary
Fixes 6 security and code quality issues with minimal, targeted changes.

## Issues Fixed
- **#523**: Bump `@modelcontextprotocol/sdk` from `^1.0.0` to `^1.26.0` (covers CVE-2025-66414, CVE-2026-0621, CVE-2026-25536)
- **#488**: Remove `--yes` flag from `npx` ast-grep availability check (supply chain risk)
- **#493**: Validate tmux `sessionName` against strict alphanumeric pattern before passing to `spawn`
- **#495**: Wrap `OMX_TEAM_STATE_ROOT` env var in try/catch so it falls back to default path if validation fails
- **#496**: Replace `execSync('codex --version 2>/dev/null')` with `execFileSync('codex', ['--version'])` to eliminate shell injection surface
- **#508**: Add exhaustive `default` branches to `getPhaseAgents`/`getPhaseInstructions` switch statements

## Verification
- `npm run build` passes

Fixes #488, Fixes #493, Fixes #495, Fixes #496, Fixes #508, Fixes #523

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.7.6",
       "license": "MIT",
       "dependencies": {
-        "@modelcontextprotocol/sdk": "^1.0.0"
+        "@modelcontextprotocol/sdk": "^1.26.0"
       },
       "bin": {
         "omx": "bin/omx.js"
@@ -866,7 +866,6 @@
       "resolved": "https://registry.npmjs.org/express/-/express-5.2.1.tgz",
       "integrity": "sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "accepts": "^2.0.0",
         "body-parser": "^2.2.1",
@@ -1180,7 +1179,6 @@
       "resolved": "https://registry.npmjs.org/hono/-/hono-4.12.1.tgz",
       "integrity": "sha512-hi9afu8g0lfJVLolxElAZGANCTTl6bewIdsRNhaywfP9K8BPf++F2z6OLrYGIinUwpRKzbZHMhPwvc0ZEpAwGw==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=16.9.0"
       }
@@ -2332,7 +2330,6 @@
       "resolved": "https://registry.npmjs.org/zod/-/zod-4.3.6.tgz",
       "integrity": "sha512-rftlrkhHZOcjDwkGlnUtZZkvaPHCsDATp4pGpuOOMDaTdDDXF91wuVDJoWoPsKX/3YPQ5fHuF3STjcYyKr+Qhg==",
       "license": "MIT",
-      "peer": true,
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
       }
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.0.0"
+    "@modelcontextprotocol/sdk": "^1.26.0"
   },
   "devDependencies": {
     "@biomejs/biome": "^2.4.4",
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -5,7 +5,7 @@
 import { existsSync } from 'fs';
 import { readdir, readFile } from 'fs/promises';
 import { join } from 'path';
-import { execSync, spawnSync } from 'child_process';
+import { execFileSync, spawnSync } from 'child_process';
 import {
   codexHome, codexConfigPath, codexPromptsDir,
   userSkillsDir, omxStateDir,
@@ -383,7 +383,7 @@ function listTeamTmuxSessions(): Set<string> | null {
 
 function checkCodexCli(): Check {
   try {
-    const version = execSync('codex --version 2>/dev/null', { encoding: 'utf-8' }).trim();
+    const version = execFileSync('codex', ['--version'], { encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe'] }).trim();
     return { name: 'Codex CLI', status: 'pass', message: `installed (${version})` };
   } catch {
     return { name: 'Codex CLI', status: 'fail', message: 'not found - install from https://github.com/openai/codex' };
--- a/src/mcp/code-intel-server.ts
+++ b/src/mcp/code-intel-server.ts
@@ -182,7 +182,7 @@ async function findSgBinary(): Promise<string | null> {
   }
   // Try npx
   try {
-    await execFileAsync('npx', ['--yes', '@ast-grep/cli', '--version'], { timeout: 15000 });
+    await execFileAsync('npx', ['@ast-grep/cli', '--version'], { timeout: 15000 });
     return 'npx-ast-grep';
   } catch { /* not available */ }
   return null;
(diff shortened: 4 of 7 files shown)


Yeachan-Heo (approved):
Follow-up review: required checks are now green on this revision, and no additional blocking issues were found in this pass. LGTM.

PR #542, opened 2026-03-03
fix(security): replace shell injection vectors with safe alternatives (#492, #494)

### LINKED ISSUE #492: [Security] notifications/tmux.ts: Command injection via execSync template literal in PID walking loop
## Summary

`detectTmuxSessionByPid` and `detectTmuxPaneByPid` use `execSync` with a template literal that embeds a loop variable derived from previous `ps` output.

## Location

- `src/notifications/tmux.ts` — lines 92, 238

## Code

```typescript
// Line 92 (detectTmuxSessionByPid) and Line 238 (detectTmuxPaneByPid):
const ppidStr = execSync(`ps -o ppid= -p ${currentPid}`, {
  encoding: "utf-8",
  timeout: 1000,
  stdio: ["pipe", "pipe", "pipe"],
}).trim();
```

## Problem

`currentPid` starts as `process.pid` but is reassigned to `parseInt(ppidStr, 10)` from the previous iteration's `ps` output on each loop pass. The string is passed to `execSync` with a shell string form (not `execFileSync`), so shell metacharacters are interpreted.

If `ps` output is tampered (e.g., via a malicious `/proc` implementation, a fake `ps` binary earlier in `PATH`, or a corrupted PID file), a crafted `ppidStr` could inject shell commands.

## Recommended Fix

Use `execFileSync('ps', ['-o', 'ppid=', '-p', String(currentPid)])` — no shell interpolation, arguments passed as an array.

## Severity

**Critical** — Direct shell injection surface, though exploitation requires ability to influence `ps` output or `PATH`.

### LINKED ISSUE #494: [Security] cli/index.ts: process.env.SHELL embedded in tmux command without allowlist validation
## Summary

`buildTmuxPaneCommand` uses `process.env.SHELL` as the shell binary path for tmux pane commands without validating it against known-safe shell paths.

## Location

- `src/cli/index.ts` — lines 1165–1176

## Code

```typescript
export function buildTmuxPaneCommand(
  command: string,
  args: string[],
  shellPath: string | undefined = process.env.SHELL
): string {
  const shellBin = shellPath && shellPath.trim() !== '' ? shellPath : '/bin/sh';
  const inner = `${rcSource}exec ${bareCmd}`;
  return `${quoteShellArg(shellBin)} -lc ${quoteShellArg(inner)}`;
}
```

## Problem

If an attacker can set `SHELL=/tmp/evil-shell` (e.g., via a compromised parent process, environment injection, or `.env` file), the tmux pane will execute the attacker's binary. The value is passed through `quoteShellArg` which prevents shell word-splitting, but does not prevent execution of a malicious binary.

## Recommended Fix

Validate `SHELL` against an allowlist of known shell paths (`/bin/sh`, `/bin/bash`, `/bin/zsh`, `/usr/bin/zsh`, `/usr/bin/bash`, etc.). Fall back to `/bin/sh` if the value doesn't match.

## Severity

**High** — Requires environment control, but environment injection is a realistic attack vector in CI/CD and shared-server environments.

junhoyeo:
## Summary
Eliminates two shell injection attack surfaces in tmux PID detection and shell binary selection.

## Issues Fixed
- **#492**: Replace `execSync` template literal interpolation in `detectTmuxSessionByPid` and `detectTmuxPaneByPid` with `execFileSync('ps', [...args])` — no shell interpretation
- **#494**: Add `ALLOWED_SHELLS` allowlist for `process.env.SHELL` in `buildTmuxPaneCommand` — falls back to `/bin/sh` if shell path is not in the allowlist

## Verification
- `npm run build` passes

Fixes #492, Fixes #494

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -123,6 +123,11 @@ const REASONING_MODES = ['low', 'medium', 'high', 'xhigh'] as const;
 type ReasoningMode = typeof REASONING_MODES[number];
 const REASONING_MODE_SET = new Set<string>(REASONING_MODES);
 const REASONING_USAGE = 'Usage: omx reasoning <low|medium|high|xhigh>';
+const ALLOWED_SHELLS = new Set([
+  '/bin/sh', '/bin/bash', '/bin/zsh', '/bin/dash', '/bin/fish',
+  '/usr/bin/sh', '/usr/bin/bash', '/usr/bin/zsh', '/usr/bin/dash', '/usr/bin/fish',
+  '/usr/local/bin/bash', '/usr/local/bin/zsh', '/usr/local/bin/fish',
+]);
 
 type CliCommand = 'launch' | 'setup' | 'uninstall' | 'doctor' | 'team' | 'version' | 'tmux-hook' | 'hooks' | 'hud' | 'status' | 'cancel' | 'help' | 'reasoning' | string;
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (commented):
Completed initial review pass. No blocking code comments from this pass; holding final LGTM until remaining check(s) finish: Coverage Gate (Team Critical).

PR #545, opened 2026-03-03
fix(perf): unify locks with backoff, cache team summary, serialize event log (#500, #501, #511, #517)

### LINKED ISSUE #500: [Performance] team/state.ts: Lock polling uses fixed interval without backoff — thundering herd under contention
## Summary

Four lock implementations (`withTeamLock`, `withTaskClaimLock`, `withMailboxLock`, `withScalingLock`) all use fixed-interval polling (25-50ms) without exponential backoff, causing thundering herd behavior under high concurrency.

## Location

- `src/team/state.ts` — lines 1136, 1184, 1249 (25ms polling)
- `src/team/scaling.ts` — 50ms polling

## Code

```typescript
// withTeamLock, withTaskClaimLock, withMailboxLock — all have:
await new Promise((resolve) => setTimeout(resolve, 25));  // fixed 25ms poll

// withScalingLock:
await new Promise((resolve) => setTimeout(resolve, 50));  // fixed 50ms poll
```

## Problem

Under contention with many workers (e.g., 10-20 workers all claiming tasks), all waiters wake up simultaneously every 25ms, all try `mkdir(lockDir)`, all but one fail, all sleep again. This is O(n) wasted syscalls per poll interval.

Notably, `withDispatchLock` already correctly implements exponential backoff with jitter (lines 1750-1752), but the other four lock implementations do not.

## Impact

CPU waste and increased lock acquisition latency under high worker concurrency. Scales quadratically with worker count.

## Recommended Fix

Apply the same exponential backoff + jitter pattern from `withDispatchLock` to all four lock implementations.

### LINKED ISSUE #501: [Performance] team/state.ts: getTeamSummary reads all task files + 2 files per worker on every call
## Summary

`getTeamSummary` performs O(tasks + 2×workers) file reads on every call, with no caching. It is called on every monitor poll cycle.

## Location

- `src/team/state.ts` — lines 2044–2136

## Code

```typescript
export async function getTeamSummary(...): Promise<TeamSummary | null> {
  const tasks = await listTasks(teamName, cwd);  // reads ALL task files
  const workerSignals = await Promise.all(
    workers.map(async (worker) => {
      const [hb, status] = await Promise.all([
        readWorkerHeartbeat(teamName, worker.name, cwd),  // file read per worker
        readWorkerStatus(teamName, worker.name, cwd),     // file read per worker
      ]);
    })
  );
}
```

## Problem

With 50 tasks and 10 workers, that's 70 file reads per call. Called on every monitor poll cycle at ~5-second intervals, this is 14 file reads/second sustained. Under load, this dominates I/O.

The existing `performance` timing in the return value suggests this was already identified as a performance concern.

## Impact

Sustained high I/O load proportional to task count × worker count. Scales poorly with team size.

## Recommended Fix

1. Cache task list with a short TTL (1-2s) and invalidate on write
2. Use `fs.watch` on the tasks directory to invalidate only on change
3. Consider an in-memory task index that syncs lazily from disk

### LINKED ISSUE #511: [Architecture] 5 duplicated file-based lock implementations should be unified
## Summary

Five nearly identical `mkdir`-based file lock implementations exist in `team/state.ts`, each with slightly different timeout, stale-lock handling, and backoff behavior.

## Locations

1. `src/team/state.ts:1104` — `withTeamLock` (25ms poll, no backoff)
2. `src/team/state.ts:1160` — `withTaskClaimLock` (25ms poll, no backoff)
3. `src/team/state.ts:1218` — `withMailboxLock` (25ms poll, no backoff)
4. `src/team/state.ts:~line` — `withScalingLock` (50ms poll, no backoff)
5. `src/team/state.ts:1750` — `withDispatchLock` (exponential backoff with jitter ✓)

## Pattern

```typescript
async function with*Lock<T>(teamName: string, cwd: string, fn: () => Promise<T>): Promise<T> {
  const lockDir = join(stateDir, '.lock-*');
  const timeout = *_MS;
  while (true) {
    try {
      await mkdir(lockDir);
      // acquired
      break;
    } catch {
      // check stale, check timeout
      await new Promise(r => setTimeout(r, POLL_MS)); // fixed interval
    }
  }
  try { return await fn(); }
  finally { await rm(lockDir, { recursive: true }).catch(() => {}); }
}
```

## Problem

- **5 copies** of nearly identical code with subtle differences
- Only `withDispatchLock` has exponential backoff — the other 4 use fixed polling (thundering herd)
- Each has slightly different stale-lock timeout detection logic
- Bug fixes must be applied 5 times
- The differences are accidental, not intentional

## Recommended Fix

Extract a single `withFileLock(lockDir, options, fn)` utility:

```typescript
interface FileLockOptions {
  timeoutMs: number;
  staleMs: number;
  initialPollMs?: number;   // default 25
  maxPollMs?: number;       // default 500
  backoffFactor?: number;   // default 1.5
  jitter?: boolean;         // default true
}

async function withFileLock<T>(
  lockDir: string,
  options: FileLockOptions,
  fn: () => Promise<T>
): Promise<T> { ... }
```

All 5 callers become one-liners with different `options`.

### LINKED ISSUE #517: [Concurrency] team/state.ts: appendTeamEvent uses non-atomic appendFile — concurrent writers corrupt NDJSON log
## Summary

`appendTeamEvent` uses `appendFile` without locking to write to the team event log. Concurrent workers interleave writes and corrupt the NDJSON format.

## Location

- `src/team/state.ts` — lines 1604–1614

## Code

```typescript
export async function appendTeamEvent(...): Promise<TeamEvent> {
  const full: TeamEvent = { ... };
  const p = eventLogPath(teamName, cwd);
  await mkdir(dirname(p), { recursive: true });
  await appendFile(p, `${JSON.stringify(full)}\n`, 'utf8');  // no lock
  return full;
}
```

## Problem

Multiple workers and the leader call `appendTeamEvent` concurrently. `appendFile` is not atomic at the application level — two concurrent calls can interleave their writes, producing corrupted NDJSON:

```
{"event_id":"a","type":"task_completed"...}{"event_id":"b","type":"worker_stopped"...}
// ↑ Two events merged on one line — JSON parse fails
```

This is especially likely during shutdown when multiple `task_completed`, `worker_stopped`, and `shutdown_gate` events fire simultaneously.

**Contrast:** Task files correctly use `writeAtomic` (tmp+rename). The event log does not.

## Recommended Fix

Either:
1. Wrap `appendFile` with a per-team file lock (reuse `withTeamLock`)
2. Use a dedicated event log writer with a queue that serializes writes
3. Write individual event files (like tasks) instead of appending to a shared log

junhoyeo:
## Summary
Unifies 5 duplicated file-based lock implementations into a single `withFileLock` utility with exponential backoff and jitter, eliminating thundering herd behavior.

## Issues Fixed
- **#511**: Extract shared `withFileLock(lockDir, options, fn)` utility with configurable timeout, stale detection, exponential backoff, and jitter. All 5 lock wrappers (`withTeamLock`, `withTaskClaimLock`, `withMailboxLock`, `withDispatchLock`, `withScalingLock`) now delegate to it.
- **#500**: All locks now use exponential backoff (25ms → 500ms, factor 1.5) with jitter, matching the pattern previously only in `withDispatchLock`.
- **#501**: Add 2-second TTL cache for `getTeamSummary` to avoid O(tasks + 2×workers) file reads per monitor poll.
- **#517**: Wrap `appendTeamEvent`'s `appendFile` in `withTeamLock` to serialize concurrent NDJSON writes.

## Verification
- `npm run build` passes

Fixes #500, Fixes #501, Fixes #511, Fixes #517

### DIFF
--- a/src/team/scaling.ts
+++ b/src/team/scaling.ts
@@ -12,6 +12,7 @@
 
 import { join, resolve } from 'path';
 import { mkdir } from 'fs/promises';
+import { withFileLock } from './state.js';
 import {
   sanitizeTeamName,
   isTmuxAvailable,
@@ -33,7 +34,6 @@ import {
   teamNormalizePolicy as normalizeTeamPolicy,
   teamReadWorkerStatus as readWorkerStatus,
   teamWriteWorkerStatus as writeWorkerStatus,
-  teamWithScalingLock as withScalingLock,
   teamAppendEvent as appendTeamEvent,
   teamMarkDispatchRequestNotified as markDispatchRequestNotified,
   teamReadDispatchRequest as readDispatchRequest,
@@ -96,6 +96,11 @@ export interface ScaleDownResult {
 export interface ScaleError {
   ok: false;
   error: string;
+}
+
+async function withScalingLock<T>(teamName: string, cwd: string, fn: () => Promise<T>): Promise<T> {
+  const lockDir = join(resolveCanonicalTeamStateRoot(cwd), 'team', teamName, '.lock.scaling');
+  return await withFileLock(lockDir, { timeoutMs: 10_000, staleMs: 5 * 60 * 1000, initialPollMs: 50 }, fn);
 }
 
 async function notifyWorkerPaneOutcome(
--- a/src/team/state.ts
+++ b/src/team/state.ts
@@ -29,7 +29,6 @@ import {
 } from './state/dispatch.js';
 import {
   resolveDispatchLockTimeoutMs as resolveDispatchLockTimeoutMsImpl,
-  withDispatchLock as withDispatchLockImpl,
 } from './state/dispatch-lock.js';
 import {
   writeTaskApproval as writeTaskApprovalImpl,
@@ -329,6 +328,9 @@ const LOCK_STALE_MS = 5 * 60 * 1000;
 const DEFAULT_DISPATCH_ACK_TIMEOUT_MS = 800;
 const MIN_DISPATCH_ACK_TIMEOUT_MS = 100;
 const MAX_DISPATCH_ACK_TIMEOUT_MS = 10_000;
+const SUMMARY_CACHE_TTL = 2000;
+
+let teamSummaryCache: { key: string; data: TeamSummary | null; ts: number } | null = null;
 
 type TeamTaskStatus = TeamTask['status'];
 
@@ -362,10 +364,6 @@ function validateTaskId(taskId: string): void {
       `Invalid task ID: "${taskId}". Must be a positive integer (digits only, max 20 digits).`
     );
   }
-}
-
-async function writeTaskClaimLockOwnerToken(ownerPath: string, ownerToken: string): Promise<void> {
-  await writeFile(ownerPath, ownerToken, 'utf8');
 }
 
 function defaultLeader(): TeamLeader {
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (changes requested):
Blocking for now: CI is failing on this revision (Test Node 20/22 and CI Status). Please fix or retrigger and resolve failures, then re-request review.

PR #588, opened 2026-03-06
feat(omx): add posture-aware agent routing metadata

HaD0Yun:
## Summary
- add posture-aware routing metadata to agent definitions so roles can distinguish frontier-orchestrator, deep-worker, and fast-lane operating styles
- extend native agent config generation to append posture/model-class overlays into generated TOML instructions
- document the experiment in README and agent tier docs, including local test steps for validating generated configs

## Testing notes
- build the branch with `npm run build`
- run `node bin/omx.js setup` to regenerate native agent configs
- inspect `~/.omx/agents/*.toml` for the new posture and model-class overlay sections
- run `node --test dist/agents/__tests__/definitions.test.js dist/agents/__tests__/native-config.test.js`

## Verification
- npm run build
- node --test dist/agents/__tests__/definitions.test.js dist/agents/__tests__/native-config.test.js


Yeachan-Heo:
Follow-up after local smoke testing on the merged branch:

I installed the updated `dev` branch locally, ran `omx setup`, and confirmed the generated native agent configs now include the new posture/model-class overlays as intended.

What worked:
- posture metadata is present in `~/.omx/agents/*.toml`
- planner/executor/etc. now get explicit posture/model-class/routing metadata in generated configs
- setup/config generation path is working end-to-end

What still looks weak in real session behavior:
- planner posture is visible in config/prompt text, but the session UX does not yet feel strongly differentiated in practice
- executor still tends to do the usual upfront exploration/read-pattern before narrow execution, so the behavioral split is not yet very crisp
- in other words: config injection is working, but runtime behavior differentiation still feels more like a soft prompt overlay than a strong routing contract

I think this was still worth merging to `dev`, but the next follow-up should probably tighten one or more of:
- initial role dispatch behavior
- slash/prompt routing into planner vs executor flows
- stronger posture-specific startup constraints or execution policy

@HaD0Yun tagging you here since this looks like the natural next step if we want posture to be felt clearly in real sessions, not just in generated config artifacts.

—

PR #606, opened 2026-03-06
Promote analyze/code-review as public entry points while keeping critic first-class

Yeachan-Heo:
## Summary
This PR advances the catalog consolidation toward a cleaner public review/analysis surface for the OMX 1.0 beta line.

Publicly, this lane is now centered on:
- `analyze`
- `code-review`
- `critic`

Where:
- `analyze` is the public entry point for investigation/diagnosis
- `code-review` is the public entry point for code review
- `critic` remains a **first-class public critique agent**

Internal expert prompts such as:
- `architect`
- `debugger`
- `code-reviewer`
- `security-reviewer`

remain available, but are pushed further behind the routing layer instead of dominating the primary public menu.

### Public review/analysis surface
The primary public surface is now centered on:
- `analyze`
- `code-review`
- `critic`

### Internal expert layer
These remain available internally / for direct expert invocation, but are no longer the primary public menu for this lane:
- `architect`
- `debugger`
- `code-reviewer`
- `security-reviewer`

### Compatibility / deprecation direction
- `review` is deprecated from the public story
- `security-review` is treated as a specialist compatibility lane rather than a headline public entry
- reviewer-family prompts remain compatibility/internal-facing rather than core public discovery points

## Why
The previous cleanup still left too much internal ontology visible in the public UX.

Users naturally think in:
- “analyze this”
- “review this code”

But `critic` is different: it is not just another alias or wrapper — it is a distinct public concept for critique.

So the target shape is:
- **task-intent public entries** for `analyze` and `code-review`
- a **first-class public critique agent** for `critic`

This gives a simpler public surface without flattening away a genuinely important reasoning mode.

## Main implementation areas
- Reframed AGENTS/docs/README around `analyze`, `code-review`, and `critic`
- Updated routing/skill contracts for `analyze`, `code-review`, `review`, and `security-review`
- Updated catalog/metadata/tests to enforce the new public-vs-internal distinction

## Guardrail
A hard constraint for this phase was preserved:
- `prompts/executor.md` remains untouched

This PR does **not** redesign the execution lane.

## Verification
- `npm run build` ✅
- targeted contract/catalog/routing tests ✅
- `node scripts/generate-catalog-docs.js --check` ✅
- `npm test` ✅

Full suite result:
- **1950 tests**
- **416 suites**
- **0 failures**

Executor guardrail check:
- `git diff --name-only -- prompts/executor.md` → empty ✅

## Notes
One team-execution edge case occurred during the run:
- a worker completed its slice and reported evidence, but one task lease expired before the final task-state transition
- the task state was repaired minimally after completion evidence was already present
- this did **not** affect the product-code changes in the branch

## Follow-ups
Not required for this PR, but likely next steps:
1. continue reducing public visibility of internal expert prompts where they still leak into docs
2. decide which compatibility aliases should remain long-term vs be removed later
3. keep `critic` explicitly public and first-class while further simplifying adjacent routing surfaces


Yeachan-Heo:
Following up: this is the expected concrete execution of the public-surface cleanup direction if it truly makes `analyze` / `code-review` clearer as user-facing entry points while keeping the underlying routing model disciplined.

What I care about here is:
- public naming gets clearer, not noisier
- `critic` remains legible instead of turning into a weird shadow API
- the docs/examples align with the actual intended surface

I'll review it with that lens.

—

PR #1148, opened 2026-04-02
fix: restore intrinsic notifications for project-scope sessions

Yeachan-Heo:
## Summary
- fix intrinsic lifecycle notifications to respect project-scope `CODEX_HOME` during `omx` launch/exit dispatch
- add regression coverage for temporary `CODEX_HOME` override behavior
- document the actual intrinsic notification trigger model and effective config path

## Root cause
`session-start` and `session-end` were dispatched from the leader process without temporarily applying the resolved project-scope `CODEX_HOME`. That made intrinsic lifecycle notifications read `~/.codex/.omx-config.json` instead of `./.codex/.omx-config.json` in project-scope launches. Manual Telegram tests could still pass because the transport itself was healthy, and the watcher path already propagated `CODEX_HOME` correctly.

## Actual trigger model
Intrinsic runtime events today:
- `session-start`
- `session-idle`
- `session-end`

Schema/template-only for now:
- `ask-user-question`
- `stop` / `session-stop`

## Testing
- `npm run build`
- `node --test dist/cli/__tests__/index.test.js dist/notifications/__tests__/verbosity.test.js`
- `npx biome lint src/cli/index.ts src/cli/__tests__/index.test.ts docs/openclaw-integration.md skills/configure-notifications/SKILL.md`

### DIFF
--- a/docs/openclaw-integration.md
+++ b/docs/openclaw-integration.md
@@ -22,6 +22,20 @@ export OMX_OPENCLAW_COMMAND=1
 export OMX_OPENCLAW_COMMAND_TIMEOUT_MS=120000
 ```
 
+## Actual intrinsic trigger model
+
+As of April 2, 2026, OMX intrinsically emits these notification events:
+
+- `session-start` — at `omx` session launch
+- `session-idle` — after completed turns when the idle cooldown allows it
+- `session-end` — when the `omx` session exits
+
+Important caveats:
+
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
This PR targets `main`, but oh-my-codex PRs should target `dev`.

The notification root-cause writeup and fix direction look valid, but the PR still needs to come through `dev` first. Please retarget or reopen the same change against `dev`, and we'll continue review there.

—

PR #1150, opened 2026-04-02
fix: harden OMX tmux routing to managed sessions only

deanpress:
## Summary
- harden notify-hook and auto-nudge tmux targeting so OMX only drives managed session panes
- remove unmanaged heuristic fallback behavior from tmux injection paths
- gate code-simplifier and auto-nudge sends behind managed-session ownership checks
- add a shared managed-tmux helper and update affected tests/fixtures
- restore internal team-dispatch/team-worker compatibility after the resolver refactor

## What changed
- added `managed-tmux.ts` for managed session identity + pane ownership checks
- updated auto-nudge to resolve panes from managed session context only
- updated tmux injection to prefer managed mode/explicit targets and avoid unmanaged routing
- removed the `resolveCodexPane()` readiness bypass from `team-tmux-guard`
- switched code-simplifier sends to managed-session-gated `sendPaneInput`
- fixed resolver call sites used by team dispatch / team worker flows

## Review / validation
- then performed an AI diff review and fixed the concrete issues it surfaced
- validation run after fixes:
  - `npm run build`
  - `node --test dist/hooks/__tests__/notify-hook-auto-nudge.test.js dist/hooks/__tests__/notify-hook-tmux-heal.test.js dist/hooks/__tests__/notify-hook-team-tmux-guard.test.js dist/hooks/__tests__/notify-hook-team-dispatch.test.js dist/hooks/__tests__/notify-hook-team-worker.test.js`

## Notes
- this keeps explicit internal OMX/team pane flows working while preventing unmanaged tmux Codex sessions from being targeted by leader-session notify-hook automation


Yeachan-Heo (changes requested):
Full-suite CI is still red, so this is not merge-ready yet.

Blocking issue:
- failing area: `src/hooks/__tests__/notify-fallback-watcher.test.ts`
- reproduced failing test: `auto-nudges stalled session output even when no active mode state exists`
- likely cause: managed-session-only routing is valid, but the fallback watcher fixture still appears to simulate the old tmux session ownership/naming, so the pane is rejected and the expected send-keys path never fires

Please fix the fixture/regression and get the full suite green before this is re-reviewed.

—

baekenough (approved):
Approved as requested by Sangyi.

PR #1175, opened 2026-04-03
Fix duplicate post-turn idle blocked alerts for unchanged sessions

Yeachan-Heo:
## Summary
- fingerprint post-turn idle notifications before the blocked hook event is emitted
- persist the idle fingerprint in idle cooldown state so unchanged idle sessions stay suppressed
- add regression coverage for unchanged idle repeats and real state transitions

## Root cause
`src/scripts/notify-hook.ts` re-entered the `session-idle` emit path after every lead-session turn, but `src/notifications/idle-cooldown.ts` only tracked `lastSentAt`. Once the cooldown elapsed, OMX could re-announce the same unchanged idle state and emit another `post_turn_idle_notification` blocked event without any transition gate.

## Verification
- `npm run build`
- `npx biome lint src/scripts/notify-hook.ts src/notifications/idle-cooldown.ts src/notifications/__tests__/idle-cooldown.test.ts`
- `npm run check:no-unused`
- `node --test dist/notifications/__tests__/idle-cooldown.test.js dist/cli/__tests__/lifecycle-notifications.test.js`

### DIFF
--- a/src/notifications/__tests__/idle-cooldown.test.ts
+++ b/src/notifications/__tests__/idle-cooldown.test.ts
@@ -69,6 +69,16 @@ describe('shouldSendIdleNotification', () => {
     assert.equal(shouldSendIdleNotification(stateDir, 'sess1'), true);
   });
 
+  it('returns true when cooldown is 0 even for unchanged fingerprints', () => {
+    process.env.OMX_IDLE_COOLDOWN_SECONDS = '0';
+    const sessionId = 'test-session-disabled-fingerprint';
+    const fingerprint = '{"phase":"idle","summary":"Waiting for input"}';
+
+    recordIdleNotificationSent(stateDir, sessionId, fingerprint);
+
+    assert.equal(shouldSendIdleNotification(stateDir, sessionId, fingerprint), true);
+  });
+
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
OMX review session result: request-changes.

Blocker summary:
- `src/notifications/idle-cooldown.ts` regresses the documented `OMX_IDLE_COOLDOWN_SECONDS=0` contract once fingerprints are present.
- `src/scripts/notify-hook.ts` now always passes a fingerprint, but the new zero-cooldown fast path only returns `true` when there is no fingerprint.
- Result: repeated identical idle states stay suppressed even when cooldown is explicitly set to 0, so the opt-out path no longer works in the real hook flow.

Repro used in review:
- `npm run build`
- `node --test dist/notifications/__tests__/idle-cooldown.test.js dist/cli/__tests__/lifecycle-notifications.test.js`
- `npm run check:no-unused`
- direct probe against `dist/notifications/idle-cooldown.js` with `OMX_IDLE_COOLDOWN_SECONDS=0` returned `{"shouldSendAgain":false}` for the same fingerprint

Please restore the zero-cooldown opt-out semantics in the fingerprinted path, then I'll re-check quickly.

—

PR #1199, opened 2026-04-03
fix: ignore repo-local .codex setup output

### LINKED ISSUE #1197: omx setup creates repo-local .codex files that can leak auth data and user paths
### Summary
Running `omx setup` creates repo-local files under `.codex/` that are easy to commit by accident.

Two concrete examples from my machine:
- `.codex/auth.json` is created in the project workspace and appears to contain authentication data.
- `.codex/config.toml` is created in the project workspace and contains user-specific absolute paths such as the home directory and local Node install paths.

In a fresh project, these files were **not automatically ignored**, so they can be staged and pushed accidentally.

### Why this is a problem
- `auth.json` looks like sensitive local credential material and should not live in a repo by default.
- `config.toml` leaks machine-specific paths and personal directory structure.
- Users may reasonably assume `omx setup` only generates safe project files, but it currently creates a mix of reusable config and local/private state inside the repository tree.

### Reproduction
1. Open a git repository that does not already ignore `.codex/`.
2. Run `omx setup`.
3. Inspect the generated files under `.codex/`.
4. Run `git status`.

### Observed
`omx setup` creates files such as:
- `.codex/auth.json`
- `.codex/config.toml`
- sqlite/log/cache/state files under `.codex/`

These are project-local and may be committed unless the user manually updates `.gitignore`.

### Expected
One of these should happen by default:
1. Sensitive and machine-local files should be stored outside the repository in a user config/state directory.
2. If repo-local files are required, `omx setup` should add safe ignore rules automatically (or at minimum warn loudly and print the exact `.gitignore` entries needed).
3. Generated output should be split more clearly between shareable project config and private local state.

### Suggested fix
- Move auth/state/cache/log/db artifacts out of the repository tree.
- Keep only intentionally shareable project configuration in the repo.
- If `.codex/config.toml` must remain project-local, avoid embedding absolute user paths when possible.
- Add default ignore coverage for private/generated `.codex` artifacts.

### Version
Observed on `oh-my-codex v0.11.12`.

Yeachan-Heo:
## Summary
- reproduce issue #1197 in a fresh git repo and confirm project-scoped `omx setup` leaves `.codex/` unignored while `.codex/config.toml` contains absolute local paths
- auto-ignore both `.omx/` and `.codex/` during project-scoped setup as an immediate containment fix
- add regression coverage for `.gitignore` creation/update and a fresh-repo git-status containment check

## Repro
Before this change, a fresh repo after `node dist/cli/omx.js setup --scope project` showed:
- `.gitignore` containing only `.omx/`
- `git status --short` reporting `?? .codex/`
- `.codex/config.toml` embedding absolute local paths to OMX scripts/MCP servers

## Verification
- `npm run build`
- `npm run lint -- src/cli/setup.ts src/cli/__tests__/setup-refresh.test.ts`
- `node --test dist/cli/__tests__/setup-refresh.test.js`
- manual fresh-repo repro confirming `git status --short --ignored` now shows `!! .codex/`

Closes #1197.

### DIFF
--- a/src/cli/setup.ts
+++ b/src/cli/setup.ts
@@ -122,7 +122,7 @@ export interface SkillFrontmatterMetadata {
   description: string;
 }
 
-const PROJECT_OMX_GITIGNORE_ENTRY = ".omx/";
+const PROJECT_GITIGNORE_ENTRIES = [".omx/", ".codex/"] as const;
 
 function applyScopePathRewritesToAgentsTemplate(
   content: string,
@@ -550,7 +550,7 @@ function hasGitignoreEntry(content: string, entry: string): boolean {
     .some((line) => line === entry);
 }
 
-async function ensureProjectOmxGitignore(
+async function ensureProjectGitignore(
   projectRoot: string,
   backupContext: SetupBackupContext,
   options: Pick<SetupOptions, "dryRun" | "verbose">,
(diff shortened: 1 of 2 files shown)


src/cli/setup.ts
```diff
@@ -122,7 +122,7 @@ export interface SkillFrontmatterMetadata {
   description: string;
 }
 
-const PROJECT_OMX_GITIGNORE_ENTRY = ".omx/";
+const PROJECT_GITIGNORE_ENTRIES = [".omx/", ".codex/"] as const;
```
kgrg, line 125: should we really ignore the entire .codex/ folder by default? 

From what I understand, .codex/ isn’t just runtime/cache. It can also contain user-defined things like AGENTS.md or custom skills (skills/*/SKILL.md), which are actually part of how the agent behaves in a project. Those files are more like config/instructions than temporary data, and Codex relies on them to guide execution

instead you can do something like 

```suggestion
const PROJECT_GITIGNORE_ENTRIES = [
  ".omx/",
  ".codex/*",
  "!.codex/AGENTS.md",
  "!.codex/skills/",
  "!.codex/skills/**/SKILL.md",
] as const;
```

Issue #1242, opened 2026-04-04
omx team live sessions can lose mailbox persistence and overwrite task-1 during create-task


Yeachan-Heo:
## Summary

During a real `omx team` session on `0.11.13`, the live runtime surfaced multiple coordination bugs:

1. `omx team api send-message` can fail with:
   - `send-message could not locate persisted mailbox message for worker-2 -> leader-fixed`
   - `send-message could not locate persisted mailbox message for leader-fixed -> worker-2`
2. `omx team api create-task` reused `task-1` and overwrote an existing task instead of allocating a new id.
3. Initial generated task ownership put all startup tasks on `worker-1`, which blocked `worker-2` from claiming its intended verification lane.

These issues caused the team run to stall and required `omx team shutdown --force`.

## Version

Observed against `0.11.13` on April 4, 2026.

## Repro

1. Start from a clean repo snapshot (the main repo was dirty, so I used a clean snapshot because `omx team` now requires a clean leader workspace for worktree mode).
2. Launch a real team session:
   ```bash
   omx team 2:executor "Validate 0.11.13 team runtime fixes ..."
   ```
3. Confirm the team starts and both workers ACK through the mailbox.
4. Send leader/worker follow-up messages with:
   ```bash
   omx team api send-message --input '{...}' --json
   ```
5. Create a dedicated verification task for `worker-2` with:
   ```bash
   omx team api create-task --input '{..."owner":"worker-2"...}' --json
   ```
6. Observe mailbox/task state and attempt normal shutdown.

## Expected

- `send-message` persists the mailbox entry and dispatch metadata reliably for both:
  - `worker -> leader-fixed`
  - `leader-fixed -> worker`
- `create-task` allocates a fresh task id and never overwrites an existing `task-<id>.json`.
- Initial task ownership should not strand a worker behind claim conflicts when the task description explicitly targets that worker.
- A normal shutdown should succeed once the team is done; forced shutdown should not be needed due to runtime bookkeeping failures.

### Mailbox persistence failure

`send-message` sometimes fails even though the session is live:

```text
send-message could not locate persisted mailbox message for worker-2 -> leader-fixed
send-message could not locate persisted mailbox message for leader-fixed -> worker-2
```

### Task overwrite on create-task

Creating a new task for `worker-2` returned id `1` and replaced the prior `task-1.json` instead of creating a new task file.

### Worker blocked by ownership mismatch

`worker-2` reported:

```text
[...]

PR #1248, opened 2026-04-04
Keep team startup task metadata in sync after bootstrap

Yeachan-Heo:
## Summary

Follow up on the #1242 investigation by keeping startup task metadata aligned with persisted team state during `startTeam`.

## Changes

- reload persisted team config after seeding initial tasks so runtime state picks up the advanced `next_task_id`
- persist `assigned_tasks` onto worker config rows during startup, matching the worker identity metadata already written to disk
- add runtime regression coverage for both `next_task_id` and `assigned_tasks` after routed startup task creation

## Validation

- [x] `npm run build`
- [ ] `npm test`

## Checklist

## Related

Related to #1242

### DIFF
--- a/src/team/__tests__/api-interop.test.ts
+++ b/src/team/__tests__/api-interop.test.ts
@@ -1,6 +1,6 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
-import { mkdtemp, rm, mkdir, readFile, writeFile } from 'node:fs/promises';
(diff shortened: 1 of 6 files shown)


harshbhx (approved):

Yeachan-Heo:
CI is green, but this stays on hold for now.

Reason: this is still inside the runtime surface, and the current owner directive is to avoid advancing runtime follow-up PRs independently while the launcher/tmux failure path is being stabilized and confirmed. So green CI here is noted, but not treated as merge-ready by itself.

—

PR #1267, opened 2026-04-05
feat(setup): add portable Bash project config style for migrations

gaofeng21cn:
## Summary
- add an opt-in `--project-config-style portable-bash` mode for `omx setup --scope project`
- generate OMX-managed notify and MCP launchers through `bash` + `npm root -g` instead of repo-local absolute paths
- persist project setup scope/style in `.omx/setup-scope.json` so reruns stay consistent
- document the portable setup flow in `README.md` and cover setup, refresh, uninstall, and idempotency with tests

## Why
Project-scoped setup currently embeds repo-local absolute paths for OMX-managed scripts. That breaks when the same repository is shared across macOS, Linux, and WSL machines, or when the repository/home path changes during migration.

This PR keeps the current behavior as the default and adds a Bash-friendly opt-in project config style for multi-machine workspaces and migrations.

## Verification
- `npm run build`
- `npm run lint`
- `npm run check:no-unused`
- `node --test dist/cli/__tests__/index.test.js dist/config/__tests__/generator-notify.test.js dist/config/__tests__/generator-idempotent.test.js dist/cli/__tests__/setup-refresh.test.js dist/cli/__tests__/setup-scope.test.js dist/cli/__tests__/uninstall.test.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,16 @@ Most users should think of OMX as **better task routing + better workflow + bett
 4. Use `$ralplan "..."` to approve the plan and review tradeoffs
 5. Choose `$team` for coordinated parallel execution or `$ralph` for persistent completion loops
 
+### Portable project setup for Bash environments
+
+If you want a project-scoped OMX install that is easier to move between macOS, Linux, and WSL machines that have Bash available, use:
+
+```bash
+omx setup --scope project --project-config-style portable-bash
+```
+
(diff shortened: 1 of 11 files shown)


gaofeng21cn:
Thanks for the thoughtful explanation. That makes sense to me, and I agree this kind of setup/migration contract should be driven from the maintainer side rather than through a broad external feature PR.

I appreciate you taking the use case seriously and for the feedback on the tests and coverage. If there's a maintainer-defined direction for this later, I'd be happy to revisit it in that form.

PR #1273, opened 2026-04-05
Fix Shift+Enter falling through to submit in tmux-backed OMX sessions

### LINKED ISSUE #1271: [Bug] shift+enter is sending message instead of creating a new line
## Summary

With omx, shift+enter will send the message straight away. This is not happening with a clean `codex` session

## Environment

oh-my-codex v0.11.13
Node.js v22.22.0
Platform: darwin arm64
codex-cli 0.118.0

## Steps to reproduce

1. fresh start `omx`
2. type message and shift+enter 

## Expected behavior

it should create a new line

## Actual behavior

message sent

Yeachan-Heo:
## Summary
- enable tmux extended keys while OMX-owned Codex TUI sessions are active
- restore the prior tmux `extended-keys` mode after in-tmux launches and on detached-session shutdown
- add focused launch-path regression tests for enable/restore behavior

## Root cause
OMX launches Codex through tmux whenever tmux is available. tmux's default `extended-keys` mode collapses `Shift+Enter` back to plain `Enter`, so Codex only sees the submit keypath instead of a multiline newline key event.

## Testing
- npm run build
- ./node_modules/.bin/biome lint src/cli/index.ts src/cli/__tests__/index.test.ts
- node --test dist/cli/__tests__/index.test.js

Fixes #1271

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -42,6 +42,8 @@ import {
   resolveNotifyFallbackWatcherScript,
   resolveHookDerivedWatcherScript,
   resolveNotifyHookScript,
+  acquireTmuxExtendedKeysLease,
+  releaseTmuxExtendedKeysLease,
   withTmuxExtendedKeys,
 } from "../index.js";
 import { HUD_TMUX_HEIGHT_LINES } from "../../hud/constants.js";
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Found one blocker before merge.

The root fix direction looks right, but the current implementation appears to toggle tmux `extended-keys` through a **server-scoped** option. That creates a concurrency/restore-risk when multiple OMX/tmux-backed sessions overlap: one launch can restore the old value while another still depends on extended key forwarding.

So this should not merge in the current shape yet. Please tighten it to a scope that cannot race across concurrent sessions, or make the restore semantics provably safe under overlapping launches.

—

PR #1313, opened 2026-04-06
[codex] Create portable Codex starter pack

lkb6925:
## Summary
- carve the repo down into a self-contained portable Codex starter pack
- add portable Codex agents, skills, Codespaces, Copilot hooks, workflows, and MCP templates
- split quickstart usage from longer operational docs and harden install, doctor, and verify flows

## Why
This turns the fork into a reusable personal starter that can be copied into new Codespaces-based projects without depending on the old OMX runtime.

## Validation
- `git diff --check`
- `cd portable-codex-starter && npm run verify:kit`
- repeated install smoke test
- script syntax checks with `bash -n` and `node --check`


Yeachan-Heo:
Thanks for the draft, but this can't be reviewed as-is.

Two blockers:
1. PRs for this repo must target `dev`, not `main`.
2. The scope rewrites oh-my-codex into a different portable starter-pack direction, which is much broader than what we can accept through this PR.

If you want to continue, please open a much narrower proposal against `dev` with repo-aligned scope instead of a large product-direction fork.

—

PR #1325, opened 2026-04-06
docs: narrow deep-interview readiness split

### LINKED ISSUE #1318: Proposal: strengthen deep-interview spec-quality scoring with ontology-first validation
First, I want to say that I am in genuine admiration of the platform, and especially of the way principles of classical philosophy are being applied to the `deep-interview` process. The combination of Socratic clarification, staged ambiguity reduction, and execution handoff is already unusually strong.

For quite a while now, I have been using the [`greek-scholastic` skill](https://github.com/strato-space/greek-scholastic/blob/main/greek-scholastic/SKILL.md) to analyze specifications for:

- ambiguity reduction,
- ontology validation,
- logic analysis,
- modality separation,
- hidden assumptions,
- and logical/fallacy detection.

What I am seeing in practice is this:

- I run `deep-interview`,
- I get a much better spec,
- and then I still run `greek-scholastic` on top of that spec,
- and I still keep finding structural issues.

That makes me think the two methods have highly complementary strengths, and that there may be real value in combining them more tightly so that the second pass becomes unnecessary in most cases.

I allowed myself to apply one method to the other, and I got the following analysis.

<details>
<summary>Ontology-first critique of <code>deep-interview</code></summary>

## I. Definitions and normalization

- **Ambiguity**: epistemic underdetermination in the task brief.
- **Execution readiness**: a composite state including
  1. epistemic clarity,
  2. practical feasibility, and
  3. delegated authority.
- **Decision Boundaries**: deontic permissions; what OMX may decide without further consent.
- **Non-goals**: explicit exclusions from scope.
- **Deep interview**: a dialogical clarification method before planning/execution.

**Normalization:** the file uses “ambiguity” for both underdetermination and go/no-go readiness. I will separate them, because they are not the same kind of thing.

### Verdict
The top-level ontology is mostly sound: this is a **pre-execution inquiry mode**.

But its **“mathematical ambiguity gating”** layer is not sound as stated. It commits a **categorical failure** by treating heterogeneous realities as one measurable substance.

### Why
The score mixes:

- intent clarity,
- outcome clarity,
- scope clarity,
- constraints,
- success criteria,
- sometimes codebase context,

and then uses that scalar as a threshold for action.

But these are not one genus.

- **Intent clarity** is teleological.
- **Constraints/context/success criteria** are operational.
- **Decision boundaries / non-goals** are deontic.

The file itself admits this fracture: `Non-goals` and `Decision Boundaries` are not in the score and must be handled as separate gates. That is evidence the scalar is **not** the thing it claims to measure.

### Concrete counterexample
A user says:

> Rename function `foo` to `bar` in `src/x.ts`; keep behavior identical; tests must pass.

This can be execution-ready even if the user’s deeper motive is underexplained.

But under the current formula, low “intent clarity” can keep “ambiguity” artificially high. So the score does not measure execution-readiness; it measures a mixed construct.

**Failure label:** categorical, with empirical consequences.

### Premises
1. A hard gate should track a coherent kind.
2. The file’s score aggregates unlike kinds.
3. The workflow also uses non-scored hard gates.
4. Therefore the score is not sufficient as “ambiguity gating.”

### Hidden assumptions
- That **motive clarity** is usually necessary for execution readiness.
- That 0–1 scores across dimensions are **commensurable** and **additive**.
- That the chosen weights are meaningful without calibration.
- That one-question rounds plus pressure passes minimize user effort in general.
- That forming a **“probable intent hypothesis”** before questioning will not anchor the inquiry.

### Inconsistencies
1. Above-threshold handoff is disallowed unless explicitly accepted, but max-round behavior can still “proceed with warning”.
2. Early crystallization can be triggered by high dimension scores even though some gating conditions are explicitly outside the score.
3. “Reduce user effort” and “trigger challenge modes at thresholds” can sometimes pull in opposite directions if challenge passes become too procedural.

### Salvage by trivialization
If one rescues the framework by saying:

> “The math is just a heuristic.”

then the strong language of thresholds, gating, and quantitative rigor collapses into: “ask questions until it feels clear enough.”

That saves the framework only by trivializing its central claim.

## IV. Modal analysis

- **Epistemic modality**: “probable intent hypothesis”, “known facts/evidence”, “brownfield evidence vs inference”, clarity scores.
- **Deontic modality**: ask one question, always preflight, do not implement directly, do not hand off until gates are explicit.
- **Instrumental/prudential modality**: prefer `explore`, prefer `request_user_input`, use challenge modes, recommend `$ralplan`.
- **Temporal/procedural modality**: phase order, round thresholds, soft warning / hard cap.

### Main modal confusion
**Decision boundaries** are deontic, not epistemic.
The file partly knows this, but still presents the overall workflow as if “ambiguity” were the master variable. It is not.

## V. What is sound

These parts are strong:
- intent-first ordering,
- brownfield evidence-backed confirmation,
- explicit non-goals and decision-boundary gates,
- no direct implementation in this mode,
- handoff contracts to later execution/planning modes.

So the workflow is not bad; its **measurement theory** is.

## VI. Minimal repair

1. **Rename the scalar**
   - from `ambiguity`
   - to `clarity_index` or `interview_confidence_heuristic`

2. **Split readiness into three gates**
   - **Epistemic**: intent, outcome, scope, constraints, success, context
   - **Deontic**: non-goals, decision boundaries, explicit opt-in where required
   - **Dialectical**: at least one real pressure-pass completed

3. **Add a bounded-task bypass**
   - If file/symbol target, bounded scope, and testable acceptance criteria already exist, low motive clarity alone should not force deep interview.

4. **Fix the hard-cap rule**
   - At max rounds: crystallize unresolved issues and stop.
   - Do **not** auto-handoff without explicit user choice.

5. **Demote the intent hypothesis**
   - Keep it as a defeasible note only.
   - It should not count toward scoring before user confirmation.

6. **Make challenge modes pattern-triggered, not round-triggered**
   - Use them when the answer type warrants them, not because the calendar says so.

## VII. Re-run under repaired ontology

Under repair, the skill becomes:

> a deontically-gated inquiry workflow with an epistemic heuristic, not a mathematically grounded measurement of ambiguity.

That ontology is sound.

## Conclusion

**Final judgment:**
Good procedural discipline, weak metaphysics.

The skill works best if read as:
- a structured clarification workflow,
- with hard authority gates,
- plus a loose heuristic for clarity.

It fails when read literally as **mathematical ambiguity gating**.

</details>

## Proposed direction

I would love to help combine the strongest parts of both methods.

My current thought is **not** to replace the existing `deep-interview` workflow, and probably **not** even to rename the key user-facing concepts that people may already be used to.

Instead, I think the skill could be strengthened by making the scoring / readiness model more rigorous internally, for example by distinguishing:

- **epistemic clarity** — do we understand the task?
- **deontic clarity** — do we know what OMX may or may not decide?
- **dialectical robustness** — has the spec survived at least one serious pressure-pass?

That would preserve the spirit and ergonomics of the existing skill while making the resulting spec more resistant to a second-pass critique from a stricter ontology/logical-analysis skill.

## Contribution offer

If this direction seems useful, I would be happy to prepare a PR that tries to:

- preserve the spirit of the original skill,
- preserve familiar user-facing terminology where practical,
- strengthen the internal scoring model,
- improve ontology/logic validation,
- and reduce the need for a second pass with another skill after `deep-interview`.

The goal would be: after `deep-interview`, the produced spec should already be strong enough that a separate `greek-scholastic` pass is no longer routinely needed.

If that sounds aligned, I can put together a concrete proposal / PR draft.

iqdoctor:
## Summary

Narrow, contract-focused follow-up to #1318.

This keeps the existing `deep-interview` workflow shape and user-facing ambiguity terminology, but makes the internal readiness contract more explicit by:

- clarifying that the weighted ambiguity score is an epistemic signal rather than the whole readiness truth,
- splitting readiness language across epistemic / deontic / robustness lanes,
- adding lightweight normalization / consistency / anti-trivialization checks,
- and extending the contract test so the change stays reviewable.

## Changes

- update `skills/deep-interview/SKILL.md` to:
  - keep `ambiguity` as the external term,
  - describe readiness as split across epistemic, deontic, and robustness checks,
  - add key-term normalization guidance,
  - add lightweight consistency + `salvage by trivialization` checks,
  - make Ontologist mode explicitly check for category mistakes / epistemic-deontic-operational collapse,
  - add compact artifact requirements for readiness split, modal summary, and minimal-repair/failure-label output
- extend `src/hooks/__tests__/deep-interview-contract.test.ts` to lock the new contract language

## Validation

- [x] `npm run build`
- [ ] `npm test` *(currently fails in unrelated existing `dist/team` tests: `startTeam preserves routed task roles into team state and worker launch args`, and `claimTask owner write failure cleans up claim lock without orphan lock dir`)*
- [x] `npm run lint`
- [x] `node --test dist/hooks/__tests__/deep-interview-contract.test.js`

## Checklist

## Related

Closes #1318

### DIFF
--- a/skills/deep-interview/SKILL.md
+++ b/skills/deep-interview/SKILL.md
@@ -43,7 +43,7 @@ If no flag is provided, use **Standard**.
 - Ask ONE question per round (never batch)
 - Ask about intent and boundaries before implementation detail
 - Target the weakest clarity dimension each round after applying the stage-priority rules below
-- When a key term is overloaded, shifts meaning, or mixes multiple categories, spend the round normalizing that term explicitly before scoring it as clear
+- Define key terms when they are used inconsistently; state the normalization before scoring them as clear
 - Treat every answer as a claim to pressure-test before moving on: the next question should usually demand evidence or examples, expose a hidden assumption, force a tradeoff or boundary, or reframe root cause vs symptom
 - Do not rotate to a new clarity dimension just for coverage when the current answer is still vague; stay on the same thread until one layer deeper, one assumption clearer, or one boundary tighter
 - Before crystallizing, complete at least one explicit pressure pass that revisits an earlier answer with a deeper, assumption-focused, or tradeoff-focused follow-up
@@ -54,10 +54,10 @@ If no flag is provided, use **Standard**.
 - For brownfield work, prefer evidence-backed confirmation questions such as "I found X in Y. Should this change follow that pattern?"
 - In Codex CLI, prefer `request_user_input` when available; if unavailable, fall back to concise plain-text one-question turns
 - Re-score ambiguity after each answer and show progress transparently
-- Treat the weighted ambiguity score as an epistemic readiness signal, not the whole truth: final readiness also requires deontic authority clarity and robustness checks
+- Treat the weighted ambiguity score as epistemic only; final readiness also requires deontic clarity and a completed pressure pass
 - Do not hand off to execution while ambiguity remains above threshold unless user explicitly opts to proceed with warning
 - Do not crystallize or hand off while `Non-goals` or `Decision Boundaries` remain unresolved, even if the weighted ambiguity threshold is met
-- If ontology or consistency fails, label the failure (`categorical`, `empirical`, `authority/deontic`, or `logical inconsistency`), state the minimal repair, and keep the workflow narrow rather than broadening into a rewrite
+- If ontology fails, label it (`categorical` or `empirical`), give a concrete counterexample, and state the minimal repair. If logic fails, state the inconsistency or `salvage by trivialization` explicitly
 - Treat early exit as a safety valve, not the default success path
 - Persist mode state for resume safety (`state_write` / `state_read`)
 </Execution_Policy>
@@ -129,13 +129,13 @@ Target the lowest-scoring dimension, but respect stage priority:
 - **Stage 2 — Feasibility:** Constraints, Success Criteria
 - **Stage 3 — Brownfield grounding:** Context Clarity (brownfield only)
 
-If a key term (`scope`, `done`, `success`, `authority`, etc.) is being used loosely, inconsistently, or across multiple categories, use the next round to normalize that term before returning to ordinary scoring.
+If a key term (`scope`, `done`, `success`, `authority`, etc.) is being used inconsistently, use the next round to define it and state the normalization before returning to ordinary scoring.
 
 Follow-up pressure ladder after each answer:
 1. Ask for a concrete example, counterexample, or evidence signal behind the latest claim
 2. Probe the hidden assumption, dependency, or belief that makes the claim true
 3. Force a boundary or tradeoff: what would you explicitly not do, defer, or reject?
-4. Separate mixed kinds before accepting the answer as clear: epistemic uncertainty, deontic authority, operational feasibility, and evaluative success should not be collapsed into one answer
+4. Separate modalities before accepting the answer as clear: epistemic uncertainty, deontic authority, operational feasibility, and evaluative success should not be collapsed into one answer
 5. If the answer still describes symptoms, reframe toward essence / root cause before moving on
 
 Prefer staying on the same thread for multiple rounds when it has the highest leverage. Breadth without pressure is not progress.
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR.

I don't think this is merge-ready as written. The overall direction is understandable, but this change pushes the deep-interview contract toward a more opinionated readiness framework (`epistemic` / `deontic` / `robustness`, modality summaries, ontology labels, etc.) and that is a bigger contract shift than this narrow PR title suggests.

For external PRs, we need a tighter change than this. If you want to retry, please reopen with a smaller diff that proves one concrete contract improvement without broadening the vocabulary and artifact requirements this much.

—

PR #1337, opened 2026-04-07
Reduce native subagent OMX MCP startup churn

cagedbird043:
## Summary
- make generated native agent role files disable leader-oriented OMX MCP servers only when those servers exist in the current config
- keep omx_code_intel untouched so coding-focused subagents still retain code-intel support
- add regression tests covering conditional injection and no-op behavior when OMX MCP entries are absent

## Why
Native Codex subagents inherit full Codex config, so they can redundantly start OMX session-scoped MCP servers (state, memory, trace, team-run). In practice this can create extra startup churn and misleading `Starting MCP servers` UI even when the subagent is already running.

## Testing
- npm run build
- npm run check:no-unused
- node --test dist/agents/__tests__/native-config.test.js

### DIFF
--- a/src/agents/__tests__/native-config.test.ts
+++ b/src/agents/__tests__/native-config.test.ts
@@ -86,6 +86,64 @@ describe("agents/native-config", () => {
     assert.doesNotMatch(tunedToml, /exact gpt-5\.4-mini model/);
   });
 
+  it("disables leader-oriented OMX MCP servers for generated native subagents only when present in config", () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
Thanks for the PR. The change itself is in a reviewable area, but this repository only takes normal contribution PRs against `dev`, not `main`. Please retarget and reopen from a branch based on `dev` if you want this considered for merge. Closing this instead of leaving it in request-changes limbo.

—

PR #1346, opened 2026-04-07
Fix issue #1342: reap leaked MCP servers after launch exit

### LINKED ISSUE #1342: MCP server processes leak on unclean Codex exit — postLaunch missing orphan cleanup
## Bug

When Codex exits uncleanly (crash, force quit, `kill -9`), the four MCP server child processes (`state-server.js`, `memory-server.js`, `code-intel-server.js`, `trace-server.js`) get reparented to PID 1 (launchd/init) and are never cleaned up. Each session leaks 4 processes (~30 MB each). Over time this compounds into severe memory exhaustion.

I hit this with **232 orphaned `oh-my-codex` processes consuming ~7 GB of RAM**, triggering macOS's "Your system has run out of application memory" dialog.

## Reproduction

1. Launch a session via `omx launch`
2. Force quit the Codex process (or let it crash)
3. Repeat a few times
4. `ps aux | grep oh-my-codex | grep -v grep | wc -l` — count grows by ~4 each time
5. Eventually: `ps aux | grep "oh-my-codex" | grep -v grep | awk '{sum += $6} END {print sum/1024, "MB"}'` shows GB-scale leakage

## Root Cause

Two issues in the cleanup logic:

### 1. `postLaunch()` does not reap orphaned MCP processes (`dist/cli/index.js`)

`preLaunch()` calls `cleanupLaunchOrphanedMcpProcesses()` before each session, but `postLaunch()` does not. The `finally` block in `launchWithHud()` calls `postLaunch()` after `runCodex()` returns — this is the natural place to clean up MCP processes left behind by the session that just died. Without this, orphans only get cleaned up when the *next* session starts, and if no new session is launched, they persist indefinitely.

### 2. `findLaunchSafeCleanupCandidates()` is too conservative with ppid=1 processes (`dist/cli/cleanup.js`)

The launch-safe filter skips any MCP process that has a live Codex or OMX launch process anywhere in its ancestor chain. But when a process is reparented to PID 1, its original parent is dead by definition — it's an orphan regardless of what other Codex instances exist in the system. The `hasAncestorMatching` check happens to return `false` for ppid=1 today (because PID 1 isn't in the process map as a Codex process), but this is fragile and the intent should be explicit.

## Suggested Fix

**Fix 1** — Add orphan cleanup to `postLaunch()`:

```js
// In postLaunch(), before the existing step 0:
try {
    const cleanup = await cleanupLaunchOrphanedMcpProcesses();
    if (cleanup.terminatedCount > 0) {
        console.log(`[omx] postLaunch: reaped ${cleanup.terminatedCount} orphaned OMX MCP process(es).`);
    }
} catch (err) {
    process.stderr.write(`[cli/index] postLaunch MCP cleanup failed: ${err}\n`);
    // Non-fatal
}
```

**Fix 2** — Make `findLaunchSafeCleanupCandidates()` always reap ppid≤1 processes:

```js
export function findLaunchSafeCleanupCandidates(processes, currentPid) {
    const processByPid = new Map(processes.map((processEntry) => [processEntry.pid, processEntry]));
    return findCleanupCandidates(processes, currentPid)
        .filter((candidate) => {
            // Processes reparented to PID 1 (launchd/init) are definitively
            // orphaned — their original parent is dead. Always reap these.
            if (candidate.ppid <= 1) return true;
            return !hasAncestorMatching(processByPid, candidate.pid, isCodexSessionProcess)
                && !hasAncestorMatching(processByPid, candidate.pid, isOmxLaunchProcess);
        });
}
```

Both patches pass the existing test suite (`node --test dist/cli/__tests__/cleanup.test.js` — 12/12, `node --test dist/mcp/__tests__/server-lifecycle.test.js` — 12/12).

## Environment

- oh-my-codex v0.12.0
- macOS (Darwin 25.2.0)
- Node.js v24.14.0

Yeachan-Heo:
## Summary
- run launch-safe MCP orphan cleanup during `postLaunch()` so crash leftovers are reaped even if no later session starts
- make launch-safe candidate selection explicitly preserve `ppid <= 1` orphan cleanup before ancestor checks
- add regression coverage for the new post-launch cleanup helper and the explicit `ppid=1` launch-safe path

## Testing
- npm run build
- node --test dist/cli/__tests__/cleanup.test.js dist/cli/__tests__/index.test.js
- npx biome lint src/cli/cleanup.ts src/cli/index.ts src/cli/__tests__/cleanup.test.ts src/cli/__tests__/index.test.ts
- npx tsc --noEmit --pretty false --project tsconfig.json

Closes #1342

### DIFF
--- a/src/cli/__tests__/cleanup.test.ts
+++ b/src/cli/__tests__/cleanup.test.ts
@@ -130,6 +130,28 @@ describe('findCleanupCandidates', () => {
       },
     ]);
   });
+
+  it('always keeps ppid=1 candidates launch-safe even if pid 1 looks like a Codex ancestor', () => {
+    const pidOneCodexProcesses: ProcessEntry[] = [
+      { pid: 1, ppid: 0, command: 'codex' },
+      { pid: 700, ppid: 500, command: 'codex' },
+      { pid: 701, ppid: 700, command: 'node /repo/bin/omx.js cleanup --dry-run' },
+      {
+        pid: 800,
+        ppid: 1,
+        command: 'node /tmp/oh-my-codex/dist/mcp/memory-server.js',
+      },
+    ];
+
+    assert.deepEqual(findLaunchSafeCleanupCandidates(pidOneCodexProcesses, 701), [
+      {
+        pid: 800,
+        ppid: 1,
+        command: 'node /tmp/oh-my-codex/dist/mcp/memory-server.js',
+        reason: 'ppid=1',
+      },
+    ]);
+  });
 });
 
 describe('listOmxProcesses', () => {
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -39,6 +39,7 @@ import {
   shouldEnableNotifyFallbackWatcher,
   reapStaleNotifyFallbackWatcher,
   cleanupLaunchOrphanedMcpProcesses,
+  reapPostLaunchOrphanedMcpProcesses,
   resolveBackgroundHelperLaunchMode,
   shouldDetachBackgroundHelper,
   resolveNotifyFallbackWatcherScript,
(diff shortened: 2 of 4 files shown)


Yeachan-Heo:
Closing this duplicate fix PR after #1345 merged the MCP orphan-cleanup path to dev.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #1383, opened 2026-04-08
Fix managed OMX hook and MCP commands for Nix-packaged installs

johnrichardrinehart:
## Summary

This fixes OMX-managed hook/MCP command generation for packaged installs, especially Nix-based ones.

Previously, `omx setup` wrote managed entries that invoked a bare `node` binary:
- `notify = ["node", ".../notify-hook.js"]`
- `[mcp_servers.omx_*].command = "node"`
- `.codex/hooks.json` commands starting with `node ...`

That worked in dev shells but could fail when Codex launched those commands outside the `omx` wrapper PATH, causing MCP startup failures and hook `127` errors.

## Changes

- generate managed `notify` and `mcp_servers.omx_*` entries with `process.execPath`
- generate native hook commands in `.codex/hooks.json` with `process.execPath`
- update config/native-hook tests to assert the active interpreter path rather than the literal string `node`
- update uninstall detection so it still recognizes managed OMX entries after interpreter pinning
- update uninstall fixtures to match the modern generated config shape
- refresh `npmDepsHash` so `nix build .` succeeds again

## Verification

- `nix build . --keep-going`

## Commits

- `f855d2a` Keep OMX-managed tool paths bound to the active Node runtime
- `e35a0c4` Recognize managed OMX tool hooks after interpreter pinning

## Notes

This branch intentionally excludes unrelated worktree changes in:
- `README.md`
- `src/cli/index.ts`
- `src/cli/__tests__/index.test.ts`

PR #1384, opened 2026-04-08

johnrichardrinehart:
Supersedes #1383 with the requested `dev`-first target and a follow-up fix for the review note about over-broad notify cleanup.

## Summary

This fixes OMX-managed hook/MCP command generation for packaged installs, especially Nix-based ones.

Previously, `omx setup` wrote managed entries that invoked a bare `node` binary:
- `notify = ["node", ".../notify-hook.js"]`
- `[mcp_servers.omx_*].command = "node"`
- `.codex/hooks.json` commands starting with `node ...`

That worked in dev shells but could fail when Codex launched those commands outside the `omx` wrapper PATH, causing MCP startup failures and hook `127` errors.

This version also tightens legacy notify cleanup so it only removes OMX-managed `notify` entries and does not strip unrelated arrays that merely reference `notify-hook.js`.

## Changes

- generate managed `notify` and `mcp_servers.omx_*` entries with `process.execPath`
- generate native hook commands in `.codex/hooks.json` with `process.execPath`
- update config/native-hook tests to assert the active interpreter path rather than the literal string `node`
- update uninstall detection so it still recognizes managed OMX entries after interpreter pinning
- update uninstall fixtures to match the modern generated config shape
- replace the broad orphan-notify regex cleanup with notify-specific line-aware cleanup
- add a regression test proving non-`notify` arrays that reference `notify-hook.js` are preserved
- refresh `npmDepsHash` so `nix build .` succeeds again

## Verification

- `nix build . --keep-going`

## Commits

- `f855d2a` Keep OMX-managed tool paths bound to the active Node runtime
- `e35a0c4` Recognize managed OMX tool hooks after interpreter pinning
- `3358d94` Keep managed notify cleanup scoped to notify entries

## Notes

This branch intentionally excludes unrelated local launch-path work in `src/cli/index.ts` and `src/cli/__tests__/index.test.ts`.


Yeachan-Heo:
Thanks for the follow-up. We're not taking this PR as submitted.

This area needs to stay tightly controlled in the active core review flow, and we're not accepting this version into the current queue.

Closing this out here.

—

PR #1387, opened 2026-04-08
Resolve Windows prompt launch commands via platform command resolution

johnrichardrinehart:
## Summary

This keeps detached launch helpers aligned on Windows by resolving the command path before encoding the PowerShell prompt wrapper.

After the recent launch-path tightening, tmux shell helpers could resolve launcher paths through the platform-command utilities, but `buildWindowsPromptCommand` still encoded the raw command token (`codex`, `node`, etc.) directly into the PowerShell wrapper. That left Windows detached prompt launches as a special-case weaker path.

## Changes

- thread a resolver through `buildWindowsPromptCommand`
- encode the resolved command path instead of the raw command token
- update the Windows prompt unit test to assert the encoded payload uses the resolved path

## Verification

- Diff review in clean `dev`-based worktree

## Commit

- `187db1a` Resolve Windows prompt launches through the platform command path

## Notes

I did not run the full build/test suite for this branch in the clean `dev` worktree because `dev` does not carry the flake packaging surface used by the packaged-install PR and this sandbox does not have the Node toolchain on PATH.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -1732,19 +1732,23 @@ describe("buildTmuxPaneCommand", () => {
 
 describe("buildWindowsPromptCommand", () => {
   it("encodes detached Windows commands for safe PowerShell prompt injection", () => {
-    const result = buildWindowsPromptCommand("codex", [
-      "--dangerously-bypass-approvals-and-sandbox",
-      "-c",
-      'model_reasoning_effort="high"',
-      "it's",
-    ]);
+    const result = buildWindowsPromptCommand(
+      "codex",
+      [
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR. We're not taking this version into the current queue.

This area needs to stay tightly controlled in the active core review flow, and this PR is not one we're accepting as submitted.

Closing this out here.

—

johnrichardrinehart:
Understood. I don't want to keep sending patches into the wrong queue. For these fixes, what is the preferred contribution path for the active core review flow: restack onto an existing maintainer branch/PR, open an issue first, or retarget against a specific branch owned by the core team? If you point me at the expected lane, I can repackage the changes to match it.

PR #1388, opened 2026-04-08
Fix managed OMX hook and MCP commands for packaged installs

johnrichardrinehart:
Successor to #1384 after GitHub refused reopen. This branch contains the packaged-install hook/MCP path fixes plus the notify-cleanup follow-up commits, including the top-level orphan-fragment fix.

PR #1391, opened 2026-04-08

johnrichardrinehart:
Successor to #1388. This version carries all prior technical feedback forward on the packaged-install branch.

Addressed feedback:
- scoped notify cleanup so non-`notify` arrays mentioning `notify-hook.js` are preserved
- removed header-stripped top-level managed notify fragments left behind after root-key stripping
- kept the runtime-path fix set otherwise unchanged

Branch tip commits:
- `f855d2a` Keep OMX-managed tool paths bound to the active Node runtime
- `e35a0c4` Recognize managed OMX tool hooks after interpreter pinning
- `3358d94` Keep managed notify cleanup scoped to notify entries
- `9e5fab6` Clean top-level managed notify fragments after root-key stripping

Verification:
- `nix build . --keep-going`


Yeachan-Heo:

Closing this out here.

—

PR #1393, opened 2026-04-08

johnrichardrinehart:
Fresh successor for the packaged-install/runtime-path fix branch. This branch already folds in the prior technical review suggestions: scoped notify cleanup, preserved non-notify arrays, and removal of header-stripped top-level managed notify fragments.

If this still needs to be routed differently for the active core review flow, please point me at the preferred intake lane and I can restack it there without changing the underlying fix set.

Branch tip commits:
- `f855d2a`
- `e35a0c4`
- `3358d94`
- `9e5fab6`

Verification:
- `nix build . --keep-going`


Yeachan-Heo:

Closing this out here.

—

PR #1395, opened 2026-04-08

johnrichardrinehart:
Fresh successor for the packaged-install/runtime-path fix branch. This branch already includes the prior technical review follow-ups: scoped notify cleanup, preservation of non-notify arrays, and cleanup of header-stripped top-level managed notify fragments.

If a different intake lane is required for the active core review flow, please point me at it and I can restack this exact branch there.


Yeachan-Heo:

Closing this out here.

—

PR #1396, opened 2026-04-08

johnrichardrinehart:
Fresh successor for the packaged-install/runtime-path fix branch with all prior technical feedback already integrated. If this still needs to move through a different active core review lane, please point me at it and I can restack the unchanged branch there.


Yeachan-Heo:

Closing this out here.

—

Issue #1373, opened 2026-04-08
[Bug] Linux + tmux: team panes/worktrees start, but workers never claim tasks and shutdown can warn 'Mode team not found'


Orinks:
## Summary

On Linux with `tmux`, `omx team` can create the leader/HUD/worker panes and worktrees successfully, but workers never actually claim tasks. All tasks remain `pending`, worker turn counts stay at `0`, and a manual nudge can leave the run in a broken state. During shutdown, OMX may also warn with `Mode team not found`.

This does **not** look like a basic tmux failure. The panes and worker processes existed, but the team runtime/state handoff never progressed the tasks.

## Environment

- OS: Linux (`Linux 6.12.74+deb13+1-cloud-amd64`)
- Multiplexer: `tmux`
- Shell: `bash`
- `oh-my-codex`: observed via `omx --version` as `v0.12.1`
- Date reproduced: April 8, 2026

## Steps to reproduce

1. In the repo, start a team run such as:
   - `omx team 3:executor "Fix AccessiWeather alert sound playback so sound plays only for genuinely new alerts. Reproduce, fix, add regression tests, verify."`
2. Wait for the team to initialize.
3. Check status:
   - `omx team status <team-name> --json`
4. Observe that panes/worktrees exist and workers are alive, but tasks remain pending.
5. If you send a manual tmux nudge to worker panes, the workers can die and the run becomes unrecoverable.
6. Attempt shutdown:
   - `omx team shutdown <team-name> --force`

## Expected behavior

- Workers should claim their assigned tasks.
- Task state should progress from `pending` to `in_progress`.
- Worker turn counts should increment.
- Shutdown should complete cleanly without team mode state warnings.

## Actual behavior

Initial stuck state:
- `status: ok`
- `phase: team-exec`
- `dead_workers: []`
- `workers.total: 3`
- `tasks.pending: 3`
- `tasks.in_progress: 0`
- worker turn counts were `0`
- worker task ids were empty
- panes existed for leader/HUD/workers

After a single manual nudge to the worker panes:
- all workers became dead
- tasks still remained `pending`
- pane lookup could fail with `can't find pane: %1`

During/after shutdown on failed runs:
- warning emitted:
  - `[omx] warning: failed to persist team mode shutdown state { team: '...', error: 'Mode team not found' }`

Additional evidence:
- `git worktree list --porcelain` showed prunable stale OMX worker worktrees after failed runs.

## Concrete status samples

[...]

PR #1409, opened 2026-04-09
Fix Linux tmux team startup handoff and shutdown state persistence

Yeachan-Heo:
## Summary\n- require observed startup evidence after direct tmux fallback before interactive worker bootstrap is treated as successful\n- allow a slightly longer startup-evidence window during team launch so Codex/Claude workers can claim work before fallback is considered settled\n- persist cancelled team mode state during shutdown even when a wedged run never created team mode metadata\n- add regressions for startup false-success and shutdown mode-state persistence\n\n## Testing\n- npm run build\n- node --test --test-name-pattern "startTeam rejects interactive startup when tmux fallback never produces worker startup evidence|persists cancelled team mode state on shutdown even when no team mode state existed beforehand" dist/team/__tests__/runtime.test.js dist/cli/__tests__/team.test.js\n\nCloses #1373

### DIFF
--- a/src/cli/__tests__/team.test.ts
+++ b/src/cli/__tests__/team.test.ts
@@ -1,1674 +1,2485 @@
-import { describe, it } from 'node:test';
-import assert from 'node:assert/strict';
-import { chmod, mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
-import { join } from 'node:path';
-import { tmpdir } from 'node:os';
-import { buildLeaderMonitoringHints, parseTeamStartArgs, teamCommand } from '../team.js';
-import { readModeState } from '../../modes/base.js';
-import { DEFAULT_MAX_WORKERS } from '../../team/state.js';
+import assert from "node:assert/strict";
 import {
-  appendTeamEvent,
-  createTask,
-  initTeamState,
-  updateWorkerHeartbeat,
-  writeMonitorSnapshot,
-  writeTaskApproval,
-  writeWorkerStatus,
-} from '../../team/state.js';
-
+	chmod,
+	mkdir,
+	mkdtemp,
+	readFile,
+	rm,
+	writeFile,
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Follow-up update:

- Reproduced the pre-fix symptom against the installed CLI (`v0.12.3`) on this Linux/tmux box: team panes/worktrees came up, `omx team status ... --json` stayed at `pending=3` / `in_progress=0`, and worker panes showed ready Codex prompts without claimed work.
- Pushed `389d52c6`, which hardens the source-order regression so the startup-evidence change does not leave the targeted team/runtime suite with a brittle formatting-only failure.
- Re-verified with:
  - `npm run build`
  - `node --test --test-name-pattern "startTeam rejects interactive startup when tmux fallback never produces worker startup evidence|startTeam captures interactive worker pid from the resolved pane id|startTeam saves interactive pane ids before readiness waits in source order|persists cancelled team mode state on shutdown even when no team mode state existed beforehand" dist/team/__tests__/runtime.test.js dist/cli/__tests__/team.test.js`

The PR branch now contains the startup-handoff fix, shutdown-mode-state persistence, and the regression/test hardening needed for the targeted suite to stay green.

Yeachan-Heo:
Clean reland is up on current `origin/dev` and opened as replacement PR #1438.

Exact reland path:
- base: `origin/dev` at `dafa7d45`
- branch: `fix/issue-1409-conflict-reland`
- reland commit: `d4e419a1`
- replacement PR: #1438

I kept scope tight to the original CLI/team/runtime intent: startup evidence after tmux fallback, shutdown team mode-state persistence, and the matching regressions.

Yeachan-Heo:
Superseded by #1438, which re-landed the same fix intent on top of current `dev` after the original branch drifted into conflict. Closing this older conflicted branch to keep the queue clean.

—

Issue #1407, opened 2026-04-09
Support current-task baseline and branch/PR guardrails for multi-session workflows


minislively:
## Problem

When working across multiple concurrent sessions, the same task can easily split into duplicate branches and PRs.

Common failure modes:
- creating another branch for the same task
- opening a new PR even though a related PR is already open
- branching from an in-progress working branch instead of the intended base branch, which pollutes diffs and commit history

Using git worktrees helps with workspace isolation, but it does not solve the deeper problem: sessions do not share a reliable **current-task baseline**.

## Why branch/PR sync alone is not enough

Even if branch and PR lists are visible, sessions still do not know:
- whether the current work should continue an existing PR
- whether creating a new branch should be blocked
- whether a task is already finished
- which branch / PR is the canonical baseline for this task

In practice, we seem to need a **current-task baseline**, not just a branch/PR inventory.

## Desired behavior

1. On session start, show the repo's open / active tasks first.
2. Before creating a branch, strongly encourage or block creating a duplicate branch when an existing task/branch/PR should be reused.
3. When a PR is merged or closed, automatically reconcile and close the corresponding task state.
4. In parallel sessions, task state should be shared, while session execution state should remain isolated.

## Conceptual model (example only)

I am not trying to prescribe the exact implementation, but conceptually something like this seems useful:

- **global**: engine for current-task state, branch/PR guardrails, lifecycle cleanup
- **project**: repo-level open-task registry
- **task**: task name, goal, scope, base branch, working branch, PR number, status
- **session**: session id, attached task id, worktree path, current intent

The key idea is:
- **task state is shared** across sessions
- **session state is isolated** per session

## Lifecycle recommendation

A practical lifecycle model might be:
- **primary signal**: PR merge / close updates task status automatically
- **secondary signal**: branch deletion / missing branch detection reconciles stale tasks

## Why this feels like an OMX concern

This is not just a Git habit issue.
It feels like an orchestration/runtime concern because the failure happens when multiple sessions operate without a shared task baseline.

## Summary

In multi-session workflows, branch and PR metadata alone are not enough.
A shared **current-task baseline** plus branch/PR guardrails and lifecycle cleanup would make repeated duplicate-branch / duplicate-PR failures much less likely.

This is meant as a problem statement + requirements proposal from real usage, not as a fixed implementation design.

Yeachan-Heo:
[...]

PR #1419, opened 2026-04-09
feat: add current-task baseline branch guardrails

Yeachan-Heo:
## Summary
- add a repo-local current-task baseline registry under `.omx/state/current-task-baseline.json`
- register named launch worktrees into that shared baseline and block duplicate branch creation when the active branch is already reserved by another worktree
- update baseline lifecycle metadata from notify-hook operational context when PR/issue info is observed

## Why this slice
This is a practical first-stage guardrail for issue #1407: it establishes a shared current-task baseline for named branch launches and uses it to prevent obvious duplicate branch/worktree drift without attempting the full multi-session orchestration system in one PR.

## Verification
- `npm run build`
- `node --test dist/team/__tests__/current-task-baseline.test.js dist/team/__tests__/worktree.test.js dist/hooks/__tests__/notify-hook-modules.test.js`

### DIFF
--- a/src/hooks/__tests__/notify-hook-modules.test.ts
+++ b/src/hooks/__tests__/notify-hook-modules.test.ts
@@ -172,6 +172,41 @@ describe('notify-hook/operational-events – buildOperationalContext', () => {
       else process.env.TMUX = originalTmux;
     }
   });
+
+  it('writes PR lifecycle details into the repo baseline registry as best effort state', async () => {
+    const { buildOperationalContext } = await loadModule('notify-hook/operational-events.js');
+    const { mkdtemp, rm, writeFile, readFile } = await import('node:fs/promises');
+    const { tmpdir } = await import('node:os');
+    const { join } = await import('node:path');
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
I like the stage-one direction here, but this should not merge as-is.

The main issue is that the notify-hook side currently broadens the baseline registry beyond the intended named-launch/worktree guardrail slice. In `buildOperationalContext()`, ordinary branch-scoped operational events can create a brand-new active baseline entry for whatever branch the session happens to be on. That makes the registry implicitly reserve branches that were never explicitly registered through the launch/worktree path.

That can lead to false duplicate-branch guard hits later.

Please tighten the notify-hook write path so it only updates an already-tracked baseline entry, or only writes when there is explicit lifecycle metadata that should mutate an existing task baseline. I also want a regression test proving ordinary operational context on an untracked branch does **not** create a fresh active baseline entry.

Once that overreach is removed, the first-stage guardrail slice looks much closer to mergeable.

—

PR #1427, opened 2026-04-09
Support approved multi-workflow overlaps in canonical state

Yeachan-Heo:
## Summary
- add a canonical workflow-transition model for approved overlaps and invalid-transition denial guidance
- preserve root/session canonical state correctly for `team + ralph` and `team + ultrawork`
- update keyword detection, state sync, HUD/overlay readers, stop handling, docs, and regression coverage around the new multi-skill rules

## Details
This treats the feature as a state-transition model change rather than a HUD-only fix.

Included changes:
- new `workflow-transition` rules and tests
- canonical skill-active/state-server updates for additive approved overlaps
- denial guidance that tells operators how to clear incompatible state via `omx state` or `omx_state.*`
- hook/native-stop/HUD/overlay coverage updates
- contract/review doc updates for the multi-state behavior

## Verification
- `npm run build`
- `node --test dist/state/__tests__/workflow-transition.test.js dist/state/__tests__/skill-active.test.js dist/hooks/__tests__/keyword-detector.test.js dist/mcp/__tests__/state-server.test.js dist/modes/__tests__/base-autoresearch-contract.test.js dist/hud/__tests__/state.test.js dist/hooks/__tests__/agents-overlay.test.js dist/scripts/__tests__/codex-native-hook.test.js`

PR #1451, opened 2026-04-10
Fix cross-session stale root team Stop blocking

HaD0Yun:
## Summary
- prevent native `Stop` from honoring another session's stale root `team-state.json`
- keep the existing same-session root fallback behavior intact
- add a regression test for the cross-session stale-root case

## Problem
A stale root `.omx/state/team-state.json` could survive from an older OMX session and still be treated as active by the native `Stop` hook. When the current session had no scoped team state, that stale root state could produce one extra blocked Stop continuation even though the active team belonged to a different session.

In practice, this looked like a Stop hook warning followed by one more unexpected continuation/nudge.

## Root cause
The native `Stop` path looked up team mode state with the generic mode reader and would still accept root `team-state.json` as a compatibility fallback. That fallback was not gated by the root state's recorded `session_id`, so a stale root state from another session could leak into the current session's Stop decision.

## What changed
- added a dedicated `readTeamModeStateForStop()` helper in `src/scripts/codex-native-hook.ts`
- for `Stop`, prefer the current session-scoped `team-state.json`
- if no scoped team state exists, only honor root `team-state.json` when:
  - it has no `session_id`, or
  - its `session_id` matches the current session
- otherwise ignore the stale root state and continue searching canonical team state as before
- added a regression test covering a stale root team state owned by another session

## Why this design
This is the smallest compatibility-preserving fix:
- it does **not** remove root fallback entirely
- it fixes the incorrect cross-session ownership leak
- it leaves valid same-session coarse-state compatibility behavior unchanged

## Overlap assessment
Adjacent but distinct.

Related prior work in this area:
- #1446 `Fix flaky hook and HUD state scope handling`
- #1427 `Support approved multi-workflow overlaps in canonical state`
- #773 `fix: avoid phantom team skill-active reporting after completion`
- #9 `fix(hooks): fallback watcher + tmux hook test`

Those changes improved hook/state behavior broadly, but they did not close this specific Stop-path bug where another session's stale root team state could still block the current session.

## Validation
- `npm run build`
- `node --test dist/scripts/__tests__/codex-native-hook.test.js dist/cli/__tests__/session-scoped-runtime.test.js dist/hooks/__tests__/notify-hook-team-leader-nudge.test.js`

## Risks / compatibility
- narrow Stop-path behavior change only
- root fallback still works for same-session compatibility and session-less coarse state
- not validated with a live tmux/Codex runtime reproduction in this worktree

## Changed files

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -2273,4 +2273,46 @@ esac
       await rm(cwd, { recursive: true, force: true });
     }
   });
+
+  it("does not fall back to active root team state when the current scoped team state is inactive", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-stop-inactive-scoped-team-"));
+    try {
+      const stateDir = join(cwd, ".omx", "state");
+      await mkdir(join(stateDir, "sessions", "sess-current"), { recursive: true });
+      await writeJson(join(stateDir, "session.json"), { session_id: "sess-current" });
+      await writeJson(join(stateDir, "sessions", "sess-current", "team-state.json"), {
+        active: false,
+        current_phase: "complete",
+        team_name: "scoped-finished-team",
+        session_id: "sess-current",
+      });
+      await writeJson(join(stateDir, "team-state.json"), {
+        active: true,
+        current_phase: "starting",
+        team_name: "root-fallback-team",
+        session_id: "sess-current",
+      });
+      await writeJson(join(stateDir, "team", "root-fallback-team", "phase.json"), {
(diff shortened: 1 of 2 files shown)


HaD0Yun:
Addressed the scoped-precedence regression.

Update summary:
- session-scoped `team-state.json` is now honored even when it is inactive
- root fallback is only used when the scoped file is absent
- cross-session root-state guard from the original fix remains in place
- added a regression test for the inactive-scoped + active-root case

Verification rerun:
- `npm run build`
- `node --test dist/scripts/__tests__/codex-native-hook.test.js dist/cli/__tests__/session-scoped-runtime.test.js dist/hooks/__tests__/notify-hook-team-leader-nudge.test.js`

PR #1454, opened 2026-04-10
fix(notifications): tolerate missing ps on Windows

Gujiassh:
## Summary
- route reply-listener process inspection through the shared platform command helper
- return `false` instead of blowing up when Windows environments do not have `ps` available
- add regression coverage for the missing-`ps` Windows path

## Validation
- `npm run build && node --test dist/notifications/__tests__/reply-listener.test.js`


Yeachan-Heo (approved):
LGTM. Local verification passed: npm run build; npx biome lint src/notifications/reply-listener.ts src/notifications/__tests__/reply-listener.test.ts; node --test dist/notifications/__tests__/reply-listener.test.js.

Issue #1461, opened 2026-04-10
Stop hook falsely reports "OMX Ralph is still active" after session-scoped cancel/clear


fitz-s:
Stop hook falsely reports “OMX Ralph is still active” after session-scoped cancel/clear

## Summary
After `/cancel` successfully clears Ralph/Ultrawork/Ralplan state for the current session, Codex Stop hooks can still inject:

> OMX Ralph is still active (phase: starting); continue the task and gather fresh verification evidence before stopping.

This can happen even when `state_list_active` returns `active_modes = []` for the workspace.

## Expected
- After session-scoped cancel/clear, Stop hook should not block on Ralph for that same session.
- If another session still has active Ralph state, it should not affect the current session’s Stop hook.
- Old hook text already injected into the conversation is a UX concern, but future Stop decisions should reflect current session truth only.

## Actual
- Current session has no active modes.
- Stop hook still emits “OMX Ralph is still active …” in this conversation.
- Investigation found the hook logic could fall back to scanning other sessions for active Ralph state.
- A stale `.omx/state/session.json` pointer also increased the chance of reading the wrong session scope.

## Repro / Evidence
1. Run Ralph in one session, then cancel/clear it.
2. Ensure `state_list_active` for the workspace reports `active_modes = []` (or at least none for the current session).
3. Trigger a Stop hook in the same Codex conversation.
4. Observe stale block message: `OMX Ralph is still active ...`

Observed local evidence during investigation:
- `state_list_active` returned `active_modes = []`
- Yet stale Stop-hook messages still appeared in-thread
- There were stale session-scoped state files such as old `ralplan-state.json`
- `.omx/state/session.json` pointed to an old session id

## Suspected root cause
`codex-native-hook` Stop handling used a helper that could:
- prefer current session first, but
- then fall back to other sessions and treat another session’s active Ralph state as globally blocking

This is incorrect for session-scoped cancellation semantics.

## Impact
- Users think `/cancel` failed even when current session state is cleared
- Stop hook becomes noisy and mistrusted
- Causes repeated false-positive “continue” loops after task completion

## Suggested fix
- Scope Ralph Stop blocking strictly to the requested/current session first.
- Do not scan unrelated sessions when a concrete `session_id` is present and that session has no active Ralph state.
- Consider clearing or invalidating stale `.omx/state/session.json` pointers during cancel or session rollover.
- Optional UX improvement: label already-injected stale hook text as historical/non-authoritative.

## Environment
[...]

PR #1465, opened 2026-04-10
Fix Stop-hook Ralph session scoping for issue #1461

Yeachan-Heo:
## Summary
- pin Ralph Stop-hook blocking to the requested/current session
- avoid root and cross-session fallback when Stop already names a session
- add regression coverage for cross-session and stale session.json cases

## Testing
- npm run build
- node --test --test-name-pattern='Ralph|root Ralph fallback|session-scoped Ralph state' dist/scripts/__tests__/codex-native-hook.test.js\n- npx biome lint src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@
 
 OMX is a workflow layer for [OpenAI Codex CLI](https://github.com/openai/codex).
 
+<table>
+<tr>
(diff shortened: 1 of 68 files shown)


Yeachan-Heo:
Closing this PR because it was opened against  and pulled in a huge unrelated branch history instead of the narrow issue #1461 fix. The actual fix needs to reland cleanly onto  as an isolated session-scoping patch, not as a main-target omnibus diff.\n\nA clean dev-based reland is being reopened from the narrow fix commit/worktree instead of trying to salvage this wrong-base PR.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #1467, opened 2026-04-11
Fix detached tmux launch cwd loss

### LINKED ISSUE #1374: Detached tmux launch can lose the project cwd and trigger Codex trust prompts in ancestor directories
## Summary

`omx` detached launches can start Codex from an ancestor directory instead of the requested project cwd. When that ancestor is untrusted, Codex shows a trust prompt in the hidden tmux session and `omx` appears to exit immediately.

## Reproduction

1. Use `omx` outside tmux on macOS/Linux so it chooses the default detached tmux path.
2. Start from a project directory under an untrusted ancestor (for example `~/Documents/...`).
3. Run `omx`.

## Observed

The detached Codex pane opens in an ancestor directory (`~` in my repro) instead of the project directory. The hidden pane shows a trust prompt like:

```text
You are in /Users/<user>
Do you trust the contents of this directory?
```

From the original terminal, `omx` just looks like it immediately exits.

## Expected

Detached launches should preserve the tmux pane cwd (`tmux new-session -c <project>`) so Codex starts in the same project directory that `omx` was launched from.

## Root cause

`src/cli/index.ts` builds the detached leader shell and pane command wrappers with login shells (`-lc`). On affected setups, that login-shell hop resets the cwd away from the tmux pane's `-c` target.

## Notes

- This is easy to misdiagnose as a trust-config, shell-config, or Node-version problem.
- Trusting the ancestor directory only masks the bug; it does not preserve the correct cwd.

I have a fix ready that switches these wrappers to `-c` while keeping explicit rc sourcing for zsh/bash pane commands.

Yeachan-Heo:
## Summary
- stop using login shells in detached tmux leader and pane wrappers so tmux's requested project cwd is preserved
- keep explicit zsh/bash rc sourcing for pane launches without reintroducing cwd resets
- add a focused regression that simulates a login profile changing directories and verifies detached startup still launches Codex in the project cwd

Closes #1374

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -1250,7 +1250,7 @@ describe("detached tmux new-session sequencing", () => {
     );
     const leaderCmd = steps[0]?.args.at(-1);
     assert.equal(typeof leaderCmd, "string");
-    assert.match(leaderCmd!, /^\/bin\/sh -lc '/);
+    assert.match(leaderCmd!, /^\/bin\/sh -c '/);
     assert.match(leaderCmd!, /acquireTmuxExtendedKeysLease/);
     assert.match(leaderCmd!, /omx_detached_session_cleanup\(\)/);
     assert.match(leaderCmd!, /trap omx_detached_session_cleanup 0 INT TERM HUP/);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Closing this PR because it was opened against . The detached tmux cwd-loss fix needs to land on  as a narrow reland, not as a wrong-base PR.\n\nI am reopening the same fix on a clean  base instead of salvaging the wrong target branch.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #1477, opened 2026-04-11
Keep session-scoped team state from surviving a finished shutdown

Yeachan-Heo:
## Summary
- persist completed/cancelled team mode state in the current session scope during `omx team shutdown`
- prevent stale session-scoped `team-state.json` from making stop hooks think a finished team is still active
- add a CLI shutdown regression test covering the session-scoped team-state path

## Testing
- npm run build
- node --test dist/cli/__tests__/team.test.js --test-name-pattern="persists cancelled session-scoped team mode state on shutdown|persists cancelled team mode state on shutdown even when no team mode state existed beforehand"
- node --test dist/cli/__tests__/session-scoped-runtime.test.js

### DIFF
--- a/src/cli/__tests__/team.test.ts
+++ b/src/cli/__tests__/team.test.ts
@@ -334,6 +334,53 @@ describe('teamCommand shutdown --force parsing', () => {
     }
   });
 
+  it('persists cancelled session-scoped team mode state on shutdown', async () => {
+    const wd = await mkdtemp(join(tmpdir(), 'omx-team-shutdown-session-mode-state-'));
+    const previousCwd = process.cwd();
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
OMX review surfaced a concrete follow-up-fix before merge: shutdown can still create/write session-scoped team state on the second path when only root state exists or a stale `session.json` is hanging around. Moving this from review stall to an active internal fix lane now.

—

Yeachan-Heo:
Follow-up check: the PR is still on the same single commit/CI-green state, and the active fix lane has not yet produced a new branch update. Keeping this out of merge for now and continuing the targeted shutdown/session-state follow-up instead of papering over the earlier review blocker.

—

PR #1491, opened 2026-04-11
Fix local Codex team workers getting stuck behind queued startup drafts

### LINKED ISSUE #1490: [Bug] Team is blocked by the local OMX worker runtime
## Environment

- Node.js: `v24.14.0`
- Codex CLI: `0.120.0`
- oh-my-codex: `v0.12.5`
- Platform: `darwin arm64 - Tahoe 26.3`

## Summary

`omx team` launches the leader pane, worker panes, and team state successfully, but workers never begin execution. They do not produce heartbeats, do not claim tasks, and the team remains stuck with all tasks pending.

## What happened

I launched `omx team` for a 5-worker executor run from a clean git worktree.

Observed behavior:

- leader pane started
- 5 worker panes started
- HUD pane started
- team state directory was created
- tasks were created and assigned

But after launch:

- team status remained `pending=5`, `in_progress=0`, `completed=0`
- workers never wrote heartbeat files
- workers never claimed tasks
- event log showed `ready_prompt_timeout`
- pane captures showed Codex sitting at startup UI instead of processing the worker inbox

## Notes

I retried after:

- installing `tmux`
- relaunching from a clean sibling worktree
- upgrading Codex CLI from `0.118.0` to `0.120.0`

The behavior was unchanged.

## Expected behavior

Workers should ACK, claim their assigned tasks, start execution, and update heartbeat/status files.

Yeachan-Heo:
## Summary
- detect Codex's queued-next-tool-call banner as an unsubmitted worker trigger
- keep nudging interactive Codex panes until the queued draft clears or the transport fails explicitly
- add tmux-session regression coverage for the queued-draft capture shape

## Root cause
Interactive Codex workers could accept the tmux-injected startup trigger into the "Messages to be submitted after next tool call" queue without actually starting execution. Team startup treated that hidden draft as a successful submit, so workers stayed pending until someone pressed Enter manually.

## Verification
- `npm run build`
- `node --test dist/team/__tests__/tmux-session.test.js`
- `node --test dist/team/__tests__/runtime.test.js`
- Manual local OMX runtime repro from a clean detached verification worktree: `./dist/cli/omx.js team 1:executor "issue 1490 repro verify"` reached `in_progress=1` on `omx team status issue-1490-repro-verify --json` without a manual pane nudge

Closes #1490.

### DIFF
--- a/src/team/__tests__/tmux-session.test.ts
+++ b/src/team/__tests__/tmux-session.test.ts
@@ -110,6 +110,11 @@ const VIEWPORT_SCROLLBACK_READY_CAPTURE = `${VIEWPORT_WITHOUT_VISIBLE_PROMPT_CAP
 
 › support lane on multi-image attach`;
 
+const QUEUED_AFTER_TOOL_CALL_CAPTURE = `• Messages to be submitted after next tool call (press esc to interrupt and send immediately)
+  ↳ Read $OMX_TEAM_STATE_ROOT/team/demo/workers/worker-1/inbox.md, work now, report progress
+
+› Write tests for @filename`;
+
 async function withMockTmuxFixture<T>(
   dirPrefix: string,
   tmuxScript: (tmuxLogPath: string) => string,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

The queued-draft startup fix is headed in the right direction, but I still see one blocking correctness gap: `paneHasQueuedCodexSubmission` matches the queue-banner phrases anywhere in the last captured tmux history, while `capturePaneAsync` reads scrollback. That means stale transcript text containing an older queue banner can be mistaken for an active queued submission.

Because `attemptSubmitRounds` / the final verification path treat that broad match as proof the trigger is still queued, this can cause redundant `C-m` retries or `submit_queued_after_tool_call` failures on panes that are actually ready.

Please constrain the detection to the live status/banner region (or otherwise harden it against scrollback false positives) before merge.

PR #1493, opened 2026-04-11
Fix team dispatch lock wedges during slow tmux hook drains

### LINKED ISSUE #1492: omx team workers stall on startup and dispatch lock wedges mid-run
## Summary

`omx team` workers in tmux intermittently fail to consume their startup inbox, remain claimed/in-progress without producing outputs, and later mailbox delivery can deadlock on the dispatch lock. The result is a stalled team run that still reports active tasks.

## Environment

- OS: macOS 26.3 (build 25D125), darwin arm64
- Node.js version: v24.9.0
- `oh-my-codex` version: v0.12.5
- Codex CLI version: codex-cli 0.120.0

## Steps to reproduce

1. Launch a team run in a non-git workspace from inside tmux, for example:
   `OMX_TEAM_READY_TIMEOUT_MS=90000 omx team 3:executor "<multi-file documentation/audit task>"`
2. Monitor with `omx team status <team-name> --json`.
3. Observe that some workers open generic Codex prompts instead of processing team inbox instructions, for example:
   - `Write tests for @filename`
   - `Use /skills to list available skills`
   - `Summarize recent commits`
4. Manually nudge panes with `tmux send-keys` as a fallback.
5. Continue monitoring while workers claim tasks and try to report progress.
6. Observe that later `omx team api send-message` / mailbox activity can fail on the dispatch lock while tasks remain `in_progress`.

## Expected behavior

- Workers reliably consume their startup inbox and claim tasks without manual tmux intervention.
- Leader-to-worker and worker-to-leader mailbox delivery does not wedge on the dispatch lock during normal execution.
- `omx team status` reflects real progress, and stalled panes do not remain indefinitely `in_progress` without output.

## Actual behavior

- Multiple startup dispatch requests failed with `codex_startup_no_evidence_after_fallback:tmux_send_keys_sent`.
- Worker panes often showed generic startup prompts instead of acting on the inbox.
- Manual `tmux send-keys ... C-m` / `C-c` interventions were required to get workers moving.
- Later mailbox sends failed with `Timed out acquiring dispatch lock for <team-name> after 15000ms`.
- Team status continued to show `pending=0, in_progress=3` even when panes were effectively stalled and no new output files were being produced.
- Forced shutdown was eventually required.

## Logs / screenshots

```text
oh-my-codex v0.12.5
Node.js v24.9.0
Platform: darwin arm64
```

```text
codex-cli 0.120.0
```

```text
"last_reason": "codex_startup_no_evidence_after_fallback:tmux_send_keys_sent"
```

```text
Error: Timed out acquiring dispatch lock for relaunch-the-client-doc-consis after 15000ms.
Set OMX_DISPATCH_LOCK_TIMEOUT_MS to increase (current: 15000ms, max: 120000ms).
```

```text
{"tasks":{"total":3,"pending":0,"blocked":0,"in_progress":3,"completed":0,"failed":0}}
```

Yeachan-Heo:
## Summary
- stop `drainPendingTeamDispatch` from holding the global dispatch lock across slow tmux injection work
- claim per-request leases under the lock, inject outside it, then reacquire the lock only to persist results
- add a hook regression proving concurrent mailbox sends no longer time out while a slow injector is running

## Root cause
Issue #1492 turned out to have **separate but interacting causes**:
1. startup stalls were the queued-Codex-draft bug already fixed in #1491
2. mid-run dispatch-lock wedges came from the notify hook monopolizing the global dispatch lock while it performed slow tmux readiness checks / retries

This PR fixes the second root cause without duplicating the startup-draft patch from #1491.

## Verification
- `npm run build`
- `node --test dist/hooks/__tests__/notify-hook-team-dispatch.test.js`
- `node --test dist/team/__tests__/state.test.js`
- `npx biome lint src/scripts/notify-hook/team-dispatch.ts src/hooks/__tests__/notify-hook-team-dispatch.test.ts`
- realistic local tmux runtime repro on stacked verification worktree (`#1491` + this commit):
  - `omx team 3:executor "1492 smoke runtime"` from a non-git workspace inside tmux
  - `omx team api send-message ...` succeeded twice mid-run
  - final status reached `in_progress=1` with `startup_failure_count=0` and `lock_timeout_count=0`

Closes #1492
Related: #1491

### DIFF
--- a/src/hooks/__tests__/notify-hook-team-dispatch.test.ts
+++ b/src/hooks/__tests__/notify-hook-team-dispatch.test.ts
@@ -1,6 +1,6 @@
 import { after, before, describe, it } from 'node:test';
 import assert from 'node:assert/strict';
-import { chmod, mkdtemp, mkdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { chmod, mkdtemp, mkdir, readFile, readdir, rm, writeFile } from 'node:fs/promises';
 import { setTimeout as sleep } from 'node:timers/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
@@ -102,6 +102,13 @@ async function readTeamDeliveryLog(cwd: string): Promise<Array<Record<string, un
     .map((line) => line.trim())
     .filter(Boolean)
     .map((line) => JSON.parse(line) as Record<string, unknown>);
+}
+
+async function listDispatchProcessingLeases(cwd: string, teamName: string): Promise<string[]> {
+  const dispatchDir = join(cwd, '.omx', 'state', 'team', teamName, 'dispatch');
+  return (await readdir(dispatchDir).catch(() => []))
+    .filter((entry) => entry.startsWith('.processing-'))
+    .sort();
 }
 
 async function writeCompatRuntimeFixture(runtimePath: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

Thanks for attacking the mid-run lock wedge — the lock-shortening direction looks right, but I still see two blocking concurrency gaps before merge:

1. requests are preclaimed before tmux injection while issue/trigger cooldown state is only persisted later during finalization, so another concurrent drain can still observe stale cooldown files and claim a second request for the same issue/trigger in that gap
2. if one claimed request throws during injection/finalization, later preclaimed requests are never visited and their lease dirs remain until stale-timeout cleanup, which can temporarily wedge those queued dispatches after a single failure

Please reserve cooldown keys atomically with the claim (or otherwise make that reservation visible before releasing the global lock) and ensure all preclaimed leases are cleaned up on early aborts before merge.

PR #1484, opened 2026-04-11
fix(notify): Stop Ralph steer after stale starting phase

Gujiassh:
## Summary
- treat stale Ralph `starting` phases as terminal in the fallback watcher
- stop periodic continue-steer sends once a handoff is stuck in `starting` past the steer timeout
- add a regression test that covers the stale-starting handoff path

Fixes #1483

## Validation
- `npm install`
- `npm run build`
- `node dist/scripts/run-test-files.js dist/hooks/__tests__/notify-fallback-watcher.test.js`

Issue #1483, opened 2026-04-11
ralph_continue_steer fires repeatedly after session handoff when current_phase stuck at 'starting'


chumicat:
## Summary

After handing off from an OMX Ralph session to Claude Code, the \`notify-fallback-watcher\` continues injecting \`Ralph loop active continue\` into the tmux pane every 60 seconds — persisting across \`/clear\` and new sessions because the watcher is a background OS process independent of Claude Code.

## Reproduction steps

1. Invoke Ralph in an OMX session (e.g. via Telegram bot)
2. Before Ralph progresses past the \`starting\` phase, hand off to Claude Code (or open a new CC session on the same tmux pane)
3. Observe: \`Ralph loop active continue\` typed into Claude Code chat every 60 seconds

## Why it persists across /clear and new sessions

The \`notify-fallback-watcher\` is a background OS process polling filesystem state files every 250ms and firing steers via \`tmux send-keys\`. It is unaffected by Claude Code session lifecycle — \`/clear\`, new sessions, browser refresh. It stops only when \`ralph-state.json\` gets \`active: false\`, the tmux pane is destroyed, or the watcher process is killed.

## Root cause

In \`notify-fallback-watcher.js\`, \`resolveActiveModeState('ralph')\`:

- Session-scoped \`ralph-state.json\` has \`active: true, current_phase: "starting"\`
- \`"starting"\` is not in \`RALPH_TERMINAL_PHASES\` (complete/failed/cancelled)
- \`completed_at\` not yet written (only set when /cancel runs)
- \`hasRalphTerminalState()\` returns \`false\` → steer fires every 60s

The guard that short-circuits the global stateDir (lines 366–372) does **not** protect against session-scoped state — that dir is checked first and the active state is found there.

## Evidence

From \`.omx/logs/notify-fallback-2026-04-11.jsonl\`:
- \`ralph_continue_steer\` events with \`current_phase: "starting"\` from 08:24–09:42 (~80 min)
- \`active: false\` + \`completed_at\` only written at 09:43 when cancel ran

State file during the issue:
```json
{
  "active": true,
  "mode": "ralph",
  "current_phase": "starting",
  "started_at": "2026-04-11T08:33:52.746Z"
}
```

## Suggested fixes

1. **Timeout the starting phase** — if Ralph has been in \`starting\` longer than the steer cadence (e.g. 2 min) with no \`last_turn_at\` update, treat as stale and block the steer
[...]

PR #1494, opened 2026-04-11
fix(notify): suppress stale starting-phase Ralph steer after handoff

Gujiassh:
## Summary
- treat long-lived starting-phase Ralph sessions as stale in the fallback watcher
- block fallback Ralph loop active continue injection when that stale starting guard is hit
- add regression coverage for session-scoped stale starting state

## Validation
- npm run build
- node --test dist/hooks/__tests__/notify-fallback-watcher.test.js

### DIFF
--- a/src/hooks/__tests__/notify-fallback-watcher.test.ts
+++ b/src/hooks/__tests__/notify-fallback-watcher.test.ts
@@ -1920,6 +1920,60 @@ exit 0
     }
   });
 
+  it('suppresses Ralph continue steer when session-scoped Ralph is stuck in stale starting phase', async () => {
+    const wd = await mkdtemp(join(tmpdir(), 'omx-fallback-ralph-starting-stale-'));
+    const fakeBinDir = join(wd, 'fake-bin');
+    const tmuxLogPath = join(wd, 'tmux.log');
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:

I’m leaving this one unmerged for three reasons:

1. this PR targets `main`, and for this review pass I’m not merging anything targeting `main`
2. the underlying `#1483` bug is already closed as fixed in `#1484`, which was merged to `dev`
3. this branch is currently red in CI (`Test (Node 20 / team-state-runtime)` and `CI Status`)

If you still want this carried forward, please retarget/restack it on the intended branch and rerun CI; otherwise it currently looks superseded by the merged dev fix.

Yeachan-Heo:
Closing this PR for now because it targets `main` instead of `dev`, and the current branch shape/CI state is not merge-ready for a direct salvage path. A clean dev-based follow-up is the right path if the change should continue.

—

Yeachan-Heo:
Replacement is now open as #1496. This one stayed closed because the original fork-owned head was based on `main` and could not be force-updated from this workspace; #1496 carries the same fix rewritten onto `dev` with local green verification.

PR #1505, opened 2026-04-12
Handle stale Ralph cleanup when scoped state already terminated

chengjon:
## Summary
- add `omx cancel ralph --stale` so half-started Ralph sessions can be terminalized safely instead of blocking Stop hooks forever
- clear the matching canonical `skill-active` state and keep stale cleanup session-scoped by default
- allow root stale fallback only when the visible scoped Ralph state is absent or already terminal

## Test Plan
- npm run build
- node --test dist/cli/__tests__/session-scoped-runtime.test.js dist/scripts/__tests__/codex-native-hook.test.js
- replay `omx cancel ralph --stale` against stale startup state and confirm the Stop hook no longer re-blocks

### DIFF
--- a/src/cli/__tests__/session-scoped-runtime.test.ts
+++ b/src/cli/__tests__/session-scoped-runtime.test.ts
@@ -285,4 +285,97 @@ describe('CLI session-scoped state parity', () => {
       await rm(wd, { recursive: true, force: true });
     }
   });
+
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the contribution. We only take repository changes through PRs that target `dev`, and this one is opened against `main`, so I'm closing it instead of reviewing it in the current form. If you still want this considered, please reopen the change as a fresh PR against `dev` with the same scope only.

—

PR #1550, opened 2026-04-13
Recognize psmux on native Windows tmux call sites

Yeachan-Heo:
## Summary
- resolve native-Windows tmux invocation points through the shared platform command layer so `psmux` is accepted when `tmux` is absent
- update detached Windows bootstrap and helper tmux callers to use the resolved tmux-compatible binary without changing launch policy
- add focused tests for command resolution/bootstrap behavior and document the narrow native-Windows compatibility path

## Testing
- npm run build
- node --test dist/utils/__tests__/platform-command.test.js dist/notifications/__tests__/tmux-detector.test.js dist/hooks/__tests__/tmux-hook-engine.test.js dist/cli/__tests__/index.test.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ omx wiki refresh --json
 
 `omx team` works best on macOS/Linux with `tmux`.
 Native Windows remains a secondary path, and WSL2 is generally the better choice if you want a Windows-hosted setup.
+On native Windows, OMX accepts `psmux` as the tmux-compatible binary for the existing tmux-backed paths it already uses.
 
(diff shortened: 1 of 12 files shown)


Yeachan-Heo:
Final review found a real blocker, so I did not merge this yet.

- `src/cli/index.ts` still has native-Windows session/launch lookups that hardcode `tmux`, so a `psmux`-only install can still hit ENOENT/logged failures on the advertised compatibility path.

I opened a same-branch follow-up fix session for that exact gap now.

—

PR #1566, opened 2026-04-13
Fix false PostToolUse failures on successful Bash output

snaplyze:
Retargeting the same fix from #1564 to `dev` per repository contribution flow.

## Summary

This fixes a false-positive `PostToolUse` failure path for Bash commands.

Before this change, successful commands could still be blocked if their `stdout` happened to contain phrases such as:

- `command not found`
- `permission denied`
- `no such file or directory`

That made the native hook misclassify benign output from search, debug, or print commands as a hard setup failure.

## Repro

Examples that can succeed but still contain those phrases in `stdout`:

```bash
printf 'command not found\n'
python3 -c "print('permission denied')"
rg -n "command not found" some/file.js
```

## Root cause

`buildNativePostToolUseOutput()` was checking `containsHardFailure()` against the combined `stderr + stdout` text.

That means a successful command with `exit_code = 0` could still be blocked if `stdout` merely mentioned one of the hard-failure phrases.

## Fix

- Return early for successful Bash tool results (`exitCode === 0`)
- Limit hard-failure detection to actual failure-oriented channels:
  - `stderr`
  - structured error fields (`error`, `message`, `details`)
- Preserve the existing review/block guidance for informative non-zero outputs

## Tests

Added regression coverage for:

- successful `stdout` containing `command not found`
- successful `stdout` containing `permission denied`

These now stay silent as expected, while real stderr failures still block.

## Verification

Passed:

```bash
node --test --test-name-pattern='PostToolUse' dist/scripts/__tests__/codex-native-hook.test.js
npx biome lint src/scripts/codex-native-pre-post.ts src/scripts/__tests__/codex-native-hook.test.ts
```

Note: running the full `dist/scripts/__tests__/codex-native-hook.test.js` suite in this fresh clone still shows two unrelated baseline failures outside this diff.


Yeachan-Heo:
Thanks for the follow-up PR. This now targets `dev`, but it's currently in a conflicting state, so I'm reviewing the exact diff and conflict surface before deciding whether to take it as-is or ask for a refresh.

—

Yeachan-Heo (changes requested):
Thanks for the follow-up. I can't take this as-is because it's now conflicting against current `dev` (dirty / 4-file conflict surface), so the branch needs a refresh before the actual code change can be judged cleanly.

Please reopen this from current `dev` and keep the scope narrow to the PostToolUse false-positive fix itself — either cherry-pick the relevant change onto a fresh branch or reimplement the same narrow behavior on top of current `dev`.

—

PR #1629, opened 2026-04-16
fix: run Codex in foreground for detached tmux launch

### LINKED ISSUE #1628: omx detached tmux launch can leave a blank Codex pane when Codex is backgrounded
## Summary

Running bare `omx` could open a detached tmux session and then show a blank Codex pane, even though the Codex process itself was still alive and consuming CPU.

This was reproducible in a real interactive iTerm2 + tmux environment.

## Environment

- oh-my-codex: 0.13.0
- codex-cli: 0.121.0
- platform: macOS arm64
- terminal: iTerm2 3.6.9
- shell: zsh
- tmux: 3.6a
- TERM: `xterm-256color`

## Symptoms

From a real iTerm2 session:

```bash
command omx
```

OMX created and attached the detached tmux session, HUD showed up, but the Codex pane stayed blank.

Important detail: the `codex` process was **not** dead. It stayed alive, held tty-backed stdio, and continued consuming CPU.

## Investigation

This turned out **not** to be a generic tmux problem and **not** a generic Codex-in-tmux problem.

These worked correctly:

```bash
tmux new-session -d -s probe "codex"
```

and

```bash
tmux new-session -d -s probe "/bin/zsh -c 'if [ -f ~/.zshrc ]; then source ~/.zshrc; fi; exec codex'"
```

The breakage was specific to OMX's detached leader shell wrapper.

It currently launches Codex like this:

```sh
${codexCmd} &
omx_codex_pid=$!
wait "$omx_codex_pid"
```

So Codex is started as a **background job** under the detached leader shell.

In real interactive use, that leaves the process alive but can break TUI rendering inside the pane. In my repro, the pane stayed blank even though:

- the process was running
- stdin/stdout/stderr were tty-backed
- manual foreground launch under the same tmux/zshrc conditions rendered correctly

## Root cause

The detached leader shell backgrounds Codex, and that backgrounded launch path appears to break Codex TUI rendering in the detached tmux pane.

Foreground execution under the same conditions renders correctly.

## Proposed fix

Run Codex in the foreground inside `buildDetachedSessionLeaderCommand()` instead of backgrounding it.

In other words, change the detached leader wrapper from:

```sh
${codexCmd} &
omx_codex_pid=$!
wait "$omx_codex_pid"
```

to a direct foreground execution path.

## Verification

Manual verification in a real iTerm2 session showed:

- backgrounded detached leader launch -> blank pane
- foreground detached leader launch -> normal Codex UI rendering

## Notes

My initial hypothesis involved tty reattachment via `/dev/tty`, but after deeper repro work the more direct root cause is simpler: **backgrounding Codex in the detached leader shell breaks the rendered UI path**.

Arsture:
## Why

Detached `omx` launch currently starts Codex from the detached leader shell as a background job:

```sh
${codexCmd} &
omx_codex_pid=$!
wait "$omx_codex_pid"
```

In real interactive iTerm2 + tmux use, that leaves the Codex process alive but can produce a blank pane instead of a rendered TUI.

This turned out not to be a generic tmux issue and not a generic `codex`-inside-tmux issue. A manual detached tmux launch with the same zshrc sourcing path renders correctly when Codex is run in the foreground.

## What changed

- updated the detached leader shell wrapper to run Codex in the foreground instead of backgrounding it
- updated detached-launch regression assertions to match the new leader-shell behavior
- kept the touched tmp-path-sensitive assertions resilient to macOS `/var` vs `/private/var` differences

## Root cause

The blank pane came from the detached leader shell's **background launch** of Codex, not from tmux attach itself.

## Verification

Ran:

```bash
npm run build
npm run lint
node --test --test-name-pattern='buildDetachedSessionBootstrapSteps kills detached tmux session on normal shell exit|detached leader command' dist/cli/__tests__/index.test.js
```

Also manually reproduced and verified in a real iTerm2 session:

- backgrounded detached leader launch -> blank Codex pane
- foreground detached leader launch -> normal Codex UI rendering

Closes #1628


Yeachan-Heo:
Thanks — the detached-launch regression has already been fixed and shipped through the narrower hotfix path (#1631 -> #1633 / v0.13.1), so I'm closing this overlapping path to avoid parallel drift.

—

PR #1644, opened 2026-04-16
fix: quote processSpec.command in worker shell startup

### LINKED ISSUE #1640: [Security] buildWorkerStartupCommand interpolates processSpec.command into shell string without quoting
## Summary

`buildWorkerStartupCommand` in `src/team/tmux-session.ts:828` interpolates `processSpec.command` directly into a POSIX shell string without `shellQuoteSingle()`, while `processSpec.args` on the preceding line is correctly quoted. This inconsistency means that if the resolved binary path contains spaces or shell metacharacters, the resulting `exec` invocation can be split or injected.

## Impact

- A `processSpec.command` that resolves to a path with embedded spaces (e.g. `/Users/John Smith/.nvm/…/codex`) will fail to exec the intended binary.
- If `OMX_LEADER_NODE_PATH` is set to a value containing shell metacharacters, the unquoted interpolation into the `exec` line allows command injection within the worker shell context.
- Same file line 826: `leaderNodeDir` uses manual single-quoting (`'${leaderNodeDir}'`) that does not escape embedded single-quote characters. A path containing `'` would break out of the quote boundary.

## Observed behavior

```typescript
// line 827 — args are correctly quoted
const quotedArgs = processSpec.args.map(shellQuoteSingle).join(' ');
// line 828 — command is NOT quoted
const cliInvocation = quotedArgs.length > 0
  ? `exec ${processSpec.command} ${quotedArgs}`
  : `exec ${processSpec.command}`;
```

The assembled string is later passed through `shellQuoteSingle` as part of the outer `sh -c` wrapper (line 833), but by that point the unquoted command is already inside the inner shell string.

## Expected behavior

`processSpec.command` should receive the same `shellQuoteSingle()` treatment as `processSpec.args`:

```typescript
const cliInvocation = quotedArgs.length > 0
  ? `exec ${shellQuoteSingle(processSpec.command)} ${quotedArgs}`
  : `exec ${shellQuoteSingle(processSpec.command)}`;
```

And `leaderNodeDir` should use `shellQuoteSingle` instead of manual template quoting:

```typescript
const pathPrefix = leaderNodeDir
  ? `export PATH=${shellQuoteSingle(leaderNodeDir)}:$PATH; `
  : '';
```

## Environment

- oh-my-codex: v0.13.0
- Verified via source analysis and PoC against `main` (commit 8a11d12)

## Suspected area

`src/team/tmux-session.ts` — `buildWorkerStartupCommand`, lines 826–833. The `shellQuoteSingle` utility is already imported and used for args; it just needs to be applied to the command and PATH prefix as well.

shaun0927:
## Summary
- apply `shellQuoteSingle()` to `processSpec.command` in `buildWorkerStartupCommand`, matching the quoting already applied to `processSpec.args` on the preceding line
- replace the manual single-quote template for `leaderNodeDir` with `shellQuoteSingle()` so paths containing embedded single-quote characters are escaped correctly
- update the POSIX leader path regression test to expect the quoted command format

## Root cause

`buildWorkerStartupCommand` (line 828) interpolates `processSpec.command` directly into a POSIX shell string without `shellQuoteSingle()`:

```typescript
// args are correctly quoted
const quotedArgs = processSpec.args.map(shellQuoteSingle).join(' ');
// command is NOT quoted
const cliInvocation = `exec ${processSpec.command} ${quotedArgs}`;
```

If the resolved binary path contains spaces or shell metacharacters (e.g. via `OMX_LEADER_NODE_PATH` pointing to a directory with spaces), the resulting `exec` invocation can split or inject into the worker shell.

Same file line 826: `leaderNodeDir` uses manual template quoting (`'${leaderNodeDir}'`) that does not escape embedded single-quote characters, which can break out of the quote boundary.

## Changes

- `src/team/tmux-session.ts` — `buildWorkerStartupCommand`:
  - Line 826: replace `export PATH='${leaderNodeDir}'` with `export PATH=${shellQuoteSingle(leaderNodeDir)}`
  - Line 828: introduce `quotedCommand = shellQuoteSingle(processSpec.command)` and use it in the `exec` invocation
- `src/team/__tests__/tmux-session.test.ts`:
  - Update "resolves POSIX leader paths" assertion regex to expect the quoted command format (`'\\''...codex'\\''` instead of bare `...codex`)

## Testing

```bash
npm run build
node --test dist/team/__tests__/tmux-session.test.js
# 167 passed, 0 failed
npx biome lint src/team/tmux-session.ts src/team/__tests__/tmux-session.test.ts
```

Closes #1640

### DIFF
--- a/src/team/__tests__/tmux-session.test.ts
+++ b/src/team/__tests__/tmux-session.test.ts
@@ -1077,7 +1077,7 @@ describe('buildWorkerStartupCommand', () => {
 
       assert.match(cmd, new RegExp(escapeRegExp(`OMX_LEADER_NODE_PATH=${nodePath}`)));
       assert.match(cmd, new RegExp(escapeRegExp(`OMX_LEADER_CLI_PATH=${codexPath}`)));
-      assert.match(cmd, new RegExp(escapeRegExp(`export PATH='\\''${fakeBin}'\\'':$PATH; exec ${codexPath}`)));
+      assert.match(cmd, new RegExp(escapeRegExp(`export PATH='\\''${fakeBin}'\\'':$PATH; exec '\\''${codexPath}'\\''`)));
       assert.doesNotMatch(cmd, /export PATH='\\''node'\\'':\$PATH/);
       assert.doesNotMatch(cmd, / exec codex(?:\s|')/);
     } finally {
--- a/src/team/tmux-session.ts
+++ b/src/team/tmux-session.ts
@@ -823,9 +823,10 @@ export function buildWorkerStartupCommand(
   }
 
   const launchSpec = buildWorkerLaunchSpec(process.env.SHELL);
-  const pathPrefix = leaderNodeDir ? `export PATH='${leaderNodeDir}':$PATH; ` : '';
+  const pathPrefix = leaderNodeDir ? `export PATH=${shellQuoteSingle(leaderNodeDir)}:$PATH; ` : '';
   const quotedArgs = processSpec.args.map(shellQuoteSingle).join(' ');
-  const cliInvocation = quotedArgs.length > 0 ? `exec ${processSpec.command} ${quotedArgs}` : `exec ${processSpec.command}`;
+  const quotedCommand = shellQuoteSingle(processSpec.command);
+  const cliInvocation = quotedArgs.length > 0 ? `exec ${quotedCommand} ${quotedArgs}` : `exec ${quotedCommand}`;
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
Reviewed in OMX.

Verdict: **approve**

Why:
- scope is tight (`src/team/tmux-session.ts` + its targeted test)
- quoting `processSpec.command` fixes the real shell-word splitting / injection surface inside the inner POSIX shell
- switching the PATH prefix to `shellQuoteSingle(leaderNodeDir)` also fixes the related broken manual-quote edge case
- targeted verification passed (`npm ci`, `npm run build`, `node --test dist/team/__tests__/tmux-session.test.js`)

—

PR #1645, opened 2026-04-16
fix: preserve Unicode characters in wiki titleToSlug

### LINKED ISSUE #1641: [Bug] Wiki titleToSlug strips all non-ASCII characters, producing opaque hash filenames for CJK titles
## Summary

`titleToSlug` in `src/wiki/storage.ts:340–355` uses the regex `/[^a-z0-9]+/g` to build the slug base. This strips every non-ASCII character — including CJK (Korean, Japanese, Chinese) and accented Latin. When the resulting base is empty (pure CJK input), the function falls back to a numeric hash, producing filenames like `page-0ad85f66.md` that give the user no indication of the page content.

## Impact

- Korean title `"프로젝트 설정 가이드"` → `page-0ad85f66.md` (hash fallback, unreadable)
- Japanese title `"プロジェクト設定"` → `page-7e1bdb55.md` (hash fallback)
- Mixed title `"My 프로젝트 Setup"` → `my-setup.md` (CJK portion silently dropped)

Given that the project creator and several core contributors work in Korean, and the wiki module was introduced as a first-class feature in v0.12.6, supporting Unicode-aware slugging seems aligned with the project's audience.

## Observed behavior

\`\`\`typescript
export function titleToSlug(title: string): string {
  const base = title
    .toLowerCase()
    .replace(/[^a-z0-9]+/g, '-')   // strips all non-ASCII
    .replace(/^-|-$/g, '')
    .slice(0, 64);

  if (!base) {
    // hash fallback for pure non-ASCII titles
    let hash = 0;
    for (let index = 0; index < title.length; index += 1) {
      hash = ((hash << 5) - hash + title.charCodeAt(index)) | 0;
    }
    return \`page-\${Math.abs(hash).toString(16).padStart(8, '0')}.md\`;
  }
  return \`\${base}.md\`;
}
\`\`\`

## Expected behavior

Non-ASCII letters and numbers should be preserved in slugs. A Unicode-property-aware regex would retain CJK, Cyrillic, Arabic, etc.:

\`\`\`typescript
export function titleToSlug(title: string): string {
  const base = title
    .toLowerCase()
    .normalize('NFKD')
    .replace(/[\u0300-\u036f]/g, '')
    .replace(/[^\p{L}\p{N}]+/gu, '-')
    .replace(/^-|-$/g, '')
    .slice(0, 64);
  // ...
}
\`\`\`

Expected results:
- \`"프로젝트 설정 가이드"\` → \`프로젝트-설정-가이드.md\`
- \`"My 프로젝트 Setup"\` → \`my-프로젝트-setup.md\`

## Environment

- oh-my-codex: v0.13.0
- Node.js 20+ (supports \`/\p{L}/u\` natively)
- Verified via PoC test reproducing the slug output for CJK inputs

## Suspected area

\`src/wiki/storage.ts\` — \`titleToSlug\` function, line 343. The fix is a single regex change. The hash fallback path can remain as a safety net.

## Additional context

A secondary concern in the same wiki module: \`withFileLockSync\` (lines 46–74) uses \`mkdirSync\` for mutual exclusion but has no stale-lock detection. If an OMX process holding the wiki lock is terminated by SIGKILL or OOM, the lock directory persists indefinitely and blocks all subsequent wiki writes with a 5-second timeout. Unlike the team state locks discussed in #511, the wiki lock has no PID-based recovery. Adding a PID file inside the lock directory and checking liveness before declaring a timeout would prevent this.

shaun0927:
## Summary
- replace the ASCII-only slug regex with a Unicode-property-aware alternative so CJK, Cyrillic, Arabic, and accented Latin characters are retained in wiki page filenames
- add NFC normalization to ensure consistent slug output across platforms
- update non-ASCII slug tests to verify character preservation instead of hash fallback

## Root cause

`titleToSlug` uses `/[^a-z0-9]+/g` which strips every non-ASCII character. For CJK input the entire base becomes empty, falling through to a numeric hash that produces opaque filenames like `page-0ad85f66.md`. Mixed titles silently drop the non-ASCII portion.

## Changes

- `src/wiki/storage.ts` — `titleToSlug`:
  - Add `.normalize('NFC')` to canonicalize composed Unicode forms
  - Replace `/[^a-z0-9]+/g` with `/[^\p{L}\p{N}]+/gu` (Unicode letter + digit aware)

- `src/wiki/__tests__/slug-nonascii.test.ts`:
  - CJK/Korean tests now assert readable slug output instead of hash fallback
  - Added mixed Latin+CJK and accented Latin test cases

## Before / After

| Title | Before | After |
|-------|--------|-------|
| `프로젝트 설정 가이드` | `page-0ad85f66.md` | `프로젝트-설정-가이드.md` |
| `日本語ドキュメント` | `page-7e1bdb55.md` | `日本語ドキュメント.md` |
| `My 프로젝트 Setup` | `my-setup.md` | `my-프로젝트-setup.md` |
| `Auth Architecture` | `auth-architecture.md` | `auth-architecture.md` |

## Testing

```bash
npm run build
node --test dist/wiki/__tests__/slug-nonascii.test.js dist/wiki/__tests__/storage.test.js
# 37 passed, 0 failed
```

Closes #1641

### DIFF
--- a/src/wiki/__tests__/slug-nonascii.test.ts
+++ b/src/wiki/__tests__/slug-nonascii.test.ts
@@ -2,17 +2,21 @@ import { describe, it } from 'node:test';
 import { expect } from './test-helpers.js';
 import { titleToSlug } from '../storage.js';
 
-describe('titleToSlug non-ASCII fallback', () => {
+describe('titleToSlug non-ASCII support', () => {
   it('Latin titles unchanged', () => {
     expect(titleToSlug('Auth Architecture')).toBe('auth-architecture.md');
   });
 
-  it('CJK title must not produce bare .md', () => {
-    expect(titleToSlug('日本語ドキュメント')).toMatch(/^page-[0-9a-f]{8}\.md$/);
+  it('CJK title preserves characters in slug', () => {
+    expect(titleToSlug('日本語ドキュメント')).toBe('日本語ドキュメント.md');
   });
 
-  it('Korean title must not produce bare .md', () => {
-    expect(titleToSlug('인증 아키텍처')).toMatch(/^page-[0-9a-f]{8}\.md$/);
+  it('Korean title preserves characters in slug', () => {
+    expect(titleToSlug('인증 아키텍처')).toBe('인증-아키텍처.md');
+  });
+
+  it('mixed Latin and CJK preserves both', () => {
+    expect(titleToSlug('My 프로젝트 Setup')).toBe('my-프로젝트-setup.md');
   });
 
   it('empty string must not produce bare .md', () => {
@@ -25,5 +29,9 @@ describe('titleToSlug non-ASCII fallback', () => {
 
   it('different CJK titles produce different slugs', () => {
     expect(titleToSlug('日本語')).not.toBe(titleToSlug('中文'));
+  });
+
+  it('accented Latin characters are preserved', () => {
+    expect(titleToSlug('café résumé')).toBe('café-résumé.md');
   });
 });
--- a/src/wiki/storage.ts
+++ b/src/wiki/storage.ts
@@ -340,7 +340,8 @@ export function appendLog(root: string, entry: WikiLogEntry): void {
 export function titleToSlug(title: string): string {
   const base = title
     .toLowerCase()
-    .replace(/[^a-z0-9]+/g, '-')
+    .normalize('NFC')
+    .replace(/[^\p{L}\p{N}]+/gu, '-')
     .replace(/^-|-$/g, '')
     .slice(0, 64);
 


Yeachan-Heo:
Reviewed in OMX.

Verdict: **approve**

Why:
- scope is narrow (`src/wiki/storage.ts` + `src/wiki/__tests__/slug-nonascii.test.ts`)
- switching slug tokenization to Unicode-aware letter/number matching is the right minimal fix for non-ASCII titles
- ASCII behavior, trimming, extension behavior, and empty-title hash fallback remain intact
- NFC normalization is a good addition so canonically equivalent accented inputs converge to one slug form
- targeted verification of ASCII/CJK/Korean/mixed/accented/decomposed-accent cases passed

—

Issue #278, opened 2026-02-25
Reply listener can be permanently blocked by stale/reused PID


Yeachan-Heo:
Severity: High

Files and lines:
- src/notifications/reply-listener.ts:226
- src/notifications/reply-listener.ts:230
- src/notifications/reply-listener.ts:674

Problem:
`startReplyListener()` relies on `isDaemonRunning()` to decide whether to start a daemon. `isDaemonRunning()` only checks whether the PID from `reply-listener.pid` exists, but does not verify process identity. If the daemon crashes and the PID is later reused by an unrelated process, `isDaemonRunning()` returns true and `startReplyListener()` exits early with "already running".

Impact:
- Reply listener startup can be silently blocked indefinitely.
- Notification reply channels (Discord/Telegram injection) become dead without automatic recovery.

Why this is a bug:
`stopReplyListener()` already has identity verification via `isReplyListenerProcess()`, but startup path does not use it. Startup and shutdown behavior are inconsistent and allow false positives.

Suggested fix:
1. In `isDaemonRunning()`, require BOTH:
   - process exists, and
   - `isReplyListenerProcess(pid)` returns true.
2. If PID exists but identity check fails, remove stale PID file and return false.
3. Add tests for stale/reused PID scenarios through `startReplyListener()` and `isDaemonRunning()` integration.

Coverage gap:
`src/notifications/__tests__/reply-listener.test.ts` validates `isReplyListenerProcess()` directly but does not cover `isDaemonRunning()` + `startReplyListener()` behavior under stale PID reuse.

Issue #158, opened 2026-02-22
bug/security: stopReplyListener can terminate unrelated processes via stale/reused PID


Yeachan-Heo:
## Summary
Reply-listener liveness/stop logic trusts only a PID file and does not verify process identity. If the PID file is stale or PID is reused, `stopReplyListener()` can kill an unrelated process.

## Affected code
- `src/notifications/reply-listener.ts`
  - `readPidFile()` parses raw PID from file (`170-175`)
  - `isDaemonRunning()` only checks `process.kill(pid, 0)` (`199-209`)
  - `stopReplyListener()` sends `SIGTERM` directly to that PID (`~734+`)

## Why this is a problem
PID files are not stable identities. PIDs can be reused. A stale `reply-listener.pid` can cause:
1. false "daemon is running" state
2. unintended termination of unrelated local processes when stop is invoked

## Reproduction (confirmed)
This reproduces in an isolated temporary HOME (so real state is untouched):

```bash
TMP_HOME=$(mktemp -d)
mkdir -p "$TMP_HOME/.omx/state"
(sleep 60) &
SLEEP_PID=$!
echo "$SLEEP_PID" > "$TMP_HOME/.omx/state/reply-listener.pid"
HOME="$TMP_HOME" node --input-type=module -e "import('./dist/notifications/reply-listener.js').then((m)=>console.log(m.stopReplyListener()))"
kill -0 "$SLEEP_PID" && echo alive || echo killed
```

Observed result: `sleep` process is killed even though it is not the reply-listener daemon.

## Expected behavior
- Store and verify stronger process identity (e.g., pid + start time/command token).
- Optionally keep a daemon token written at startup and verify before signaling.
- On mismatch, treat PID file as stale and clean it rather than signaling.

## Duplicate check
I searched open/closed issues for `reply listener`, `pid`, and related terms and did not find an existing issue describing this PID-reuse kill behavior.

Yeachan-Heo:
Confirmed. Starting work on this. — repo owner's gaebal-gajae (clawdbot) 🦞

PR #1672, opened 2026-04-16
fix: verify process identity before reaping stale notify-fallback-watcher PID

### LINKED ISSUE #1657: [Bug] reapStaleNotifyFallbackWatcher sends SIGTERM without process identity verification — PID reuse pattern fixed in reply-listener (#158, #278) remains here
## Summary

`reapStaleNotifyFallbackWatcher` in `src/cli/index.ts:3047` reads a PID from a file and sends `SIGTERM` without verifying that the process is actually a notify-fallback-watcher. The PID file stores a `started_at` timestamp, but `parseWatcherPidFile` discards it and extracts only the numeric PID.

This is the same PID-reuse vulnerability pattern that was previously identified and fixed in `reply-listener.ts` via #158 and #278.

## Impact

- If the original watcher process exits and the OS reassigns that PID to an unrelated process, `SIGTERM` is sent to the wrong process
- The signal is `SIGTERM` (graceful, not `SIGKILL`), so the target process can handle it, but it may still cause unexpected behavior for unrelated programs
- The race window is narrow (between reading the PID file and calling `process.kill`), but the pattern is a known class of bug that the project has already addressed elsewhere

## Root cause

```typescript
// src/cli/index.ts:3031-3045 — parseWatcherPidFile extracts only PID
function parseWatcherPidFile(content: string): number | null {
  const trimmed = content.trim();
  if (!trimmed) return null;
  try {
    const parsed = JSON.parse(trimmed) as { pid?: unknown };
    return typeof parsed.pid === "number" && Number.isFinite(parsed.pid) && parsed.pid > 0
      ? parsed.pid
      : null;
  } catch {
    const pid = Number.parseInt(trimmed, 10);
    return Number.isFinite(pid) && pid > 0 ? pid : null;
  }
}

// src/cli/index.ts:3047-3082 — reapStaleNotifyFallbackWatcher
// Sends SIGTERM based solely on PID, no identity check
const prevPid = parseWatcherPidFile(await readFileImpl(pidPath, "utf-8"));
if (prevPid) {
  tryKillPidImpl(prevPid, "SIGTERM");  // no verification that this PID is still a watcher
}
```

The PID file is written at line 3138 as `{ pid, started_at }`, so the `started_at` field is already available but never consulted before signaling.

The same function at line 3255 (`stopNotifyFallbackWatcher`) exhibits the identical pattern.

## Prior art in this repository

- **#158** — `bug/security: stopReplyListener can terminate unrelated processes via stale/reused PID`
- **#278** — `Reply listener can be permanently blocked by stale/reused PID`

Both issues were classified as security-relevant, addressed, and closed. The notify-fallback-watcher code was not updated with the same fix.

## Suggested fix

Apply the same identity-verification approach used in the reply-listener fix. At minimum, verify the process command line before sending the signal:

```typescript
function isWatcherProcess(pid: number): boolean {
  try {
    const cmd = execFileSync("ps", ["-p", String(pid), "-o", "command="], {
      encoding: "utf-8",
      timeout: 2000,
    }).trim();
    return cmd.includes("notify-fallback-watcher");
  } catch {
    return false;
  }
}

// In reapStaleNotifyFallbackWatcher:
if (prevPid && isWatcherProcess(prevPid)) {
  tryKillPidImpl(prevPid, "SIGTERM");
}
```

Alternatively, compare the PID file's `started_at` against the process start time from the OS, matching the pattern from the reply-listener fix.

## Environment

- oh-my-codex: v0.11.13+ (pattern present since at least v0.11.13, still on `dev`)
- Affected platforms: macOS, Linux (watcher is disabled on Windows by default)
- Verified via source analysis: `parseWatcherPidFile` extracts `pid` but ignores `started_at`

shaun0927:
## Summary
- add `isLikelyOmxWatcherProcess` that checks the process command line via `ps` before sending SIGTERM, preventing PID-reuse false kills
- add `parseWatcherPidRecord` to extract both `pid` and `started_at` from the watcher PID file (the `started_at` field was already written but never read)
- update `reapStaleNotifyFallbackWatcher` to use identity verification before signaling, with an injectable `isWatcherProcess` dependency for testability
- update existing tests to pass the new dependency and add two regression tests for the identity-check gate

## Root cause

`reapStaleNotifyFallbackWatcher` (`src/cli/index.ts:3047`) reads a PID from a file and sends `SIGTERM` without verifying that the target process is actually a notify-fallback-watcher. If the original watcher exits and the OS reassigns the PID to an unrelated process, `SIGTERM` reaches the wrong target.

This is the same PID-reuse pattern that was previously identified and fixed in `reply-listener.ts` via #158 and #278, but was not addressed in the notify-fallback-watcher code path.

## Changes

- `src/cli/index.ts`:
  - Add `WatcherPidRecord` interface and `parseWatcherPidRecord` function that extracts both `pid` and `started_at`
  - Add `isLikelyOmxWatcherProcess` that runs `ps -p <pid> -o command=` and checks for watcher process signatures
  - Add `isWatcherProcess` to `reapStaleNotifyFallbackWatcher` deps interface
  - Replace bare `parseWatcherPidFile` + unconditional `tryKillPid` with `parseWatcherPidRecord` + identity-gated kill
- `src/cli/__tests__/index.test.ts`:
  - Add `isWatcherProcess: () => true` to existing tests that expect kill behavior
  - Add "skips kill when process identity does not match" regression test
  - Add "sends SIGTERM only after confirming watcher identity" regression test

## Testing

```bash
npm run build
node --test --test-name-pattern="reap|watcher.*identity|watcher.*confirmed" dist/cli/__tests__/index.test.js
# 8 passed, 0 failed
npx biome lint src/cli/index.ts src/cli/__tests__/index.test.ts
```

Closes #1657

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -685,6 +685,7 @@ describe("reapStaleNotifyFallbackWatcher", () => {
       const killed: Array<{ pid: number; signal?: NodeJS.Signals }> = [];
 
       await reapStaleNotifyFallbackWatcher(pidPath, {
+        isWatcherProcess: () => true,
         tryKillPid(pid, signal) {
           killed.push({ pid, signal });
           return true;
@@ -737,6 +738,7 @@ describe("reapStaleNotifyFallbackWatcher", () => {
       const warned: Array<{ message: unknown; meta: unknown }> = [];
       await reapStaleNotifyFallbackWatcher(pidPath, {
         readFile: async (path, encoding) => readFile(path, encoding),
+        isWatcherProcess: () => true,
         tryKillPid() {
           throw new Error("permission denied");
         },
@@ -1936,7 +1938,37 @@ exit 0
     ]);
   });
 
-  it("buildDetachedSessionFinalizeSteps keeps schedule after split-capture and before attach", () => {
+  it("reapStaleNotifyFallbackWatcher skips kill when process identity does not match a watcher", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-reap-pid-identity-"));
+    const pidPath = join(cwd, "watcher.pid");
+    await writeFile(pidPath, JSON.stringify({ pid: 99999, started_at: new Date().toISOString() }));
+
+    const killed: number[] = [];
+    await reapStaleNotifyFallbackWatcher(pidPath, {
+      isWatcherProcess: () => false,
+      tryKillPid: (pid) => { killed.push(pid); return true; },
+    });
+
+    assert.equal(killed.length, 0, "should not kill a process that is not a watcher");
+    await rm(cwd, { recursive: true, force: true });
+  });
+
+  it("reapStaleNotifyFallbackWatcher sends SIGTERM only after confirming watcher identity", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-reap-pid-confirmed-"));
+    const pidPath = join(cwd, "watcher.pid");
+    await writeFile(pidPath, JSON.stringify({ pid: 12345, started_at: new Date().toISOString() }));
+
+    const killed: number[] = [];
+    await reapStaleNotifyFallbackWatcher(pidPath, {
+      isWatcherProcess: () => true,
+      tryKillPid: (pid) => { killed.push(pid); return true; },
+    });
+
+    assert.deepEqual(killed, [12345], "should SIGTERM the verified watcher process");
+    await rm(cwd, { recursive: true, force: true });
+  });
+
+    it("buildDetachedSessionFinalizeSteps keeps schedule after split-capture and before attach", () => {
     const steps = buildDetachedSessionFinalizeSteps(
       "omx-demo",
       "%12",
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -1968,6 +1968,65 @@ exit 0
     await rm(cwd, { recursive: true, force: true });
   });
 
+  it("reuses legacy plain-text PID parsing without widening stale reap semantics across PID reuse", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-reap-legacy-pid-"));
+    try {
+      const pidPath = join(cwd, "watcher.pid");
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
The PID-reuse hardening direction is correct, but I'm not approving this PR as-is.

The review work showed the changed surface is `src/cli/index.ts` + `src/cli/__tests__/index.test.ts`, and the targeted verification did run clean after installing deps. But the review still landed on `request-changes` because the proof story for the exact process-identity contract is not tight enough yet for a safety-sensitive stale-reap change: the patch needs a more explicit, narrow demonstration that the legacy/plain-text watcher PID path and the newer watcher record path both preserve the intended kill/no-kill behavior across PID reuse boundaries without widening semantics.

Please tighten the exact contract coverage around `parseWatcherPidFile` / stale watcher reaping behavior, then re-request review.

—

Yeachan-Heo:
Conservative PR sweep on 2026-04-16: not merging this batch.

Current blocker set:
- `reviewDecision=CHANGES_REQUESTED`
- `mergeStateStatus=DIRTY`

Please clear both the requested-changes review state and the dirty merge state before this is reconsidered for merge.

Yeachan-Heo:
Follow-up fix is now pushed on top of current dev, and the earlier dirty merge state is cleared.

Current PR head: `3214b105`

What changed in the follow-up:
- tightened the stale watcher PID reuse contract so legacy plain-text PID files and watcher-record files both pass through the same identity gate semantics
- added focused regressions for kill / no-kill behavior across PID reuse boundaries without widening semantics

Focused verification on the updated head:
- `npx biome lint src/cli/index.ts src/cli/__tests__/index.test.ts`
- `npm run build`
- `node --test --test-name-pattern="reapStaleNotifyFallbackWatcher|legacy plain-text PID parsing|watcher-record PIDs only" dist/cli/__tests__/index.test.js`

This should now be ready for re-review.

—

Yeachan-Heo (approved):
Re-review complete. The follow-up now tightens the stale watcher PID reuse contract in the narrow path that was previously under-proven: legacy plain-text PID files and watcher-record files both go through the same identity gate semantics, and the focused kill/no-kill behavior across PID reuse boundaries is now directly covered. Merge risk is narrow and acceptable.

—

Issue #1698, opened 2026-04-17
[Bug] omx explore incident-class recursion is still reproducible after #1695, so #1692 should not be considered fully resolved


dwchoo:
## Summary

Issue [#1692](https://github.com/Yeachan-Heo/oh-my-codex/issues/1692) was closed after PR [#1695](https://github.com/Yeachan-Heo/oh-my-codex/pull/1695), but the stronger incident-class runaway is still reproducible.

The narrow `#1695` fix appears to address **one host-command self-resolution path**, but it does **not** fully stop the broader recursion / re-entry failure class that caused the real process storm and host-visible memory pressure.

In other words:

- `#1695` is not useless; it narrows one known failure path
- but it is **not sufficient** to treat `#1692` as fully solved

## Why I am filing this follow-up

The original issue was about an `omx explore` runaway that could explode process count and drive CPU / memory pressure high enough to destabilize a host.

That stronger failure class is still reproducible after `#1695`.

This follow-up is specifically about:

- documenting that `#1692` was closed too early for the stronger incident-class repro
- showing that `#1695` is insufficient for that broader failure mode
- describing what remains missing in the `omx explore` hardening

## What `#1695` fixes

Merged commit tested:
- `f84291e439732ce563263309fe9104a213af3a92`
- merge of PR `#1695`

What `#1695` clearly does:
- it prevents one narrow case where an allowlist wrapper can be re-selected during host command resolution

That is a useful fix.

## What `#1695` does **not** fully fix

The stronger incident-class repro shows that additional protections are still missing.

The merged state still lacks robust defense against:

1. **shell startup contamination / re-entry**
   - the explore subprocess still depends on ambient shell-startup behavior more than it should
   - `BASH_ENV`, `ENV`, and `PROMPT_COMMAND` are not isolated in the merged state

2. **wrapper target re-entry after initial resolution**
   - even if the first resolution step avoids one self-resolution path, forwarded direct/shell wrapper targets can still participate in the same broader recursion class

3. **bounded recursion failure**
   - there is no explicit recursion depth guard in the merged state
   - if re-entry occurs, it can still expand into a process storm instead of failing fast

## Environment

- OS: Linux
- Node.js: `v24.14.0`
- repository state tested: upstream `dev`
- merged commit tested: `f84291e439732ce563263309fe9104a213af3a92`

## Reproduction model

[...]

Issue #1692, opened 2026-04-17
[Bug] omx explore allowlist wrapper recursion can trigger process explosion and host resource runaway


dwchoo:
## Summary

`omx explore` can enter a runaway allowlist-wrapper recursion path and create incident-scale process / CPU / memory pressure.

The immediate trigger observed in the live incident was a recursive chain around:

```text
/bin/bash /tmp/omx-explore-allowlist-.../bin/grep ^CUDA Version
```

Operator-side inspection saw roughly:
- ~220k owned processes
- ~226k owned `bash` processes
- ~120k+ identical wrapper-chain instances

At the same time, stale old-main MCP trees were still alive and repeatedly ran:

```text
ps axww -o pid=,ppid=,command=
```

which amplified CPU and memory pressure once the process table became huge.

<img width="1920" height="917" alt="Image" src="https://github.com/user-attachments/assets/ad11491d-86d5-4e10-bf0c-f03e3a49845e" />

## Environment

- OS: Linux (Ubuntu container/host)
- Node.js version: v24.14.0
- `oh-my-codex` version: 0.13.1
- Codex CLI version: local installed CLI with `codex exec` / `codex app-server`

## Steps to reproduce

Natural/live evidence:
1. Run `omx explore` against a target repository in a host where old-main behavior is present.
2. Hit the wrapper recursion class around the allowlist `grep` path.
3. Observe runaway `/bin/bash /tmp/omx-explore-allowlist-.../bin/grep ^CUDA Version` chains, CPU saturation, and memory pressure.

Deterministic guided repro used during investigation:
1. Run old-main `omx explore` / old-main `omx-explore-harness`.
2. Poison allowlist resolution so a bare allowlisted command can self-resolve back into an allowlist dir.
3. Trigger the observed leaf command family (`grep ^CUDA Version`) deterministically via a small codex stub.
4. Observe runaway wrapper growth and large process-tree RSS.

Combined incident-shape repro:
1. Start several old-main planning roots (to keep old MCP trees alive).
2. Start one old-main `omx explore` recursion root.
3. Observe the primary trigger (`omx explore` recursion) plus the secondary amplifier (old-main MCP `ps axww` polling).

## Expected behavior

- `omx explore` should perform bounded read-only inspection and exit cleanly.
- Allowlist wrappers should never recursively invoke themselves.
- Read-only explore runs should not create runaway bash/grep chains.
[...]

PR #1699, opened 2026-04-17
fix: harden omx explore against recursion paths still reproducible after #1695

dwchoo:
## Summary

This PR is an **explore-only follow-up** to PR [#1695](https://github.com/Yeachan-Heo/oh-my-codex/pull/1695).

`#1695` narrowed one allowlist self-resolution path, but during follow-up investigation the stronger incident-class recursion / re-entry bug was still reproducible on top of the merged `dev` state.

This PR adds defense in depth so the stronger guided repro no longer runs away.

## Why this PR is needed

The merged `#1695` state still reproduced under the stronger guided bug-class repro:

- old main:
  - `stop_reason = rss_threshold`
  - `peak_tree_rss_gb = 4.42`
  - `peak_wrapper_count = 1623`
- merged `dev @ f84291e`:
  - `stop_reason = rss_threshold`
  - `peak_tree_rss_gb = 4.40`
  - `peak_wrapper_count = 1617`

A combined incident-shape repro also remained severe after `#1695`:

- `peak_aggregate_tree_rss_gb = 213.93`
- `peak_memavailable_drop_gb = 43.09`
- `peak_wrapper_count = 78069`
- `peak_ps_axww_count = 5`

That means the narrow fix in `#1695` was helpful but **not sufficient** for the stronger incident-linked recursion class.

### 1. Shell-startup isolation for explore subprocesses
The explore harness now launches its subprocesses with:
- `BASH_ENV` pointing at an inert `empty-startup.sh`
- `ENV` pointing at the same inert startup stub
- `PROMPT_COMMAND` cleared
- `OMX_EXPLORE_ALLOWLIST_DEPTH=0`

This prevents ambient shell-startup behavior from re-entering allowlist wrappers.

### 2. Stronger allowlist-dir rejection
The harness now treats any path under an `omx-explore-allowlist-*` ancestor as toxic, not just one narrow self-resolution shape.

### 3. Explicit wrapper re-entry protection
Internal direct/shell wrappers now reject targets that still resolve back into an allowlist dir.

### 4. Recursion depth fail-fast
A bounded recursion-depth guard prevents the wrapper chain from expanding into a runaway process storm even if re-entry is encountered.

## Scope

This PR intentionally stays **explore-only**.

It does **not** include MCP duplicate-polling changes.
Those remain a separate concern and should be handled independently.

### Focused checks
- [x] `cargo test -p omx-explore-harness`
- [x] `cargo build -p omx-explore-harness`
- [x] `npm run build`
- [x] `npm exec -- biome lint crates/omx-explore/src/main.rs src/cli/__tests__/explore.test.ts`
- [x] `npx tsc --noEmit --pretty false`

### Manual smoke
- [x] verified `omx explore` still succeeds end-to-end with a codex stub
- [x] verified allowlisted commands still work (`rg`)
- [x] verified non-allowlisted commands remain blocked (`node`)
- [x] verified explore subprocess env now includes:
  - `BASH_ENV=.../empty-startup.sh`
  - `ENV=.../empty-startup.sh`
  - `PROMPT_COMMAND=`
  - `OMX_EXPLORE_ALLOWLIST_DEPTH=0`

### Bounded guided repro verification
After this change:

#### deterministic guided repro
- `stop_reason = process_exited`
- `peak_tree_rss_gb = 0.03`
- `peak_wrapper_count = 0`

#### bounded combined repro
- `stop_reason = timeout`
- `peak_aggregate_tree_rss_gb = 2.23`
- `peak_memavailable_drop_gb = 1.85`
- `peak_wrapper_count = 0`

Interpretation:

> The stronger incident-class wrapper recursion did not reproduce under the bounded verification runs after this change.

## Notes on behavior risk

This hardening intentionally removes ambient shell-startup influence from `omx explore` subprocesses.

That means users should no longer rely on:
- `.bashrc` side effects
- shell aliases/functions
- `PROMPT_COMMAND`

inside the explore harness.

This is considered a hardening correction rather than a supported behavior regression, because `omx explore` is intended to run in a tightly controlled, allowlisted, read-only environment.

## Related

- Follow-up for issue [#1692](https://github.com/Yeachan-Heo/oh-my-codex/issues/1692).
- Addresses the still-open follow-up report in [#1698](https://github.com/Yeachan-Heo/oh-my-codex/issues/1698).
- This PR is intended to address the stronger recursion / re-entry class that remained reproducible after [#1695](https://github.com/Yeachan-Heo/oh-my-codex/pull/1695).

### DIFF
--- a/crates/omx-explore/src/main.rs
+++ b/crates/omx-explore/src/main.rs
@@ -519,8 +519,12 @@ fn prepare_allowlist_environment() -> Result<AllowlistEnvironment, String> {
     write_executable(&shell_path, &bash_wrapper)?;
     write_executable(&bin_dir.join("sh"), &sh_wrapper)?;
     let startup_env_path = root.path.join("empty-startup.sh");
-    write(&startup_env_path, b"")
-        .map_err(|err| format!("failed to write startup env stub {}: {err}", startup_env_path.display()))?;
+    write(&startup_env_path, b"").map_err(|err| {
+        format!(
+            "failed to write startup env stub {}: {err}",
+            startup_env_path.display()
+        )
+    })?;
 
     Ok(AllowlistEnvironment {
         bin_dir,
@@ -1391,9 +1395,8 @@ exec node "$basedir/../@openai/codex/bin/codex.js" "$@"
             env::set_var(ALLOWLIST_DEPTH_ENV, MAX_ALLOWLIST_DEPTH.to_string());
         }
         let err = checked_allowlist_depth().expect_err("depth should fail");
-        match env::var_os(ALLOWLIST_DEPTH_ENV) {
-            Some(_) => unsafe { env::remove_var(ALLOWLIST_DEPTH_ENV) },
-            None => {}
+        if env::var_os(ALLOWLIST_DEPTH_ENV).is_some() {
+            unsafe { env::remove_var(ALLOWLIST_DEPTH_ENV) }
         }
         assert!(err.contains("recursion depth exceeded"));
     }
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Closing this in favor of the narrower follow-up in #1700.

The remaining reproducible path was reduced to shell-startup env re-entry, and #1700 keeps the fix scoped to that narrower explore-only hardening path.

—

PR #1701, opened 2026-04-17
Keep prompt routing anchored to the latest user turn

LEON-gittech:
## Summary
- detect natural-language execution handoffs and short same-thread follow-ups in the native prompt-submit hook
- prioritize the latest user turn in injected context so execution confirmations do not fall back into plan restatement
- log only the latest user input preview in notify-hook turn logs instead of concatenating prior inputs
- add focused regression tests for both behaviors

## Why
Two failures showed up in real Codex/OMX usage:
1. responses could appear to target an older query because turn previews were built from the whole `input_messages` array
2. after a user said to start executing, the model could still restate the previous plan because natural-language execution handoffs were not surfaced as a high-priority context signal

This patch keeps the latest turn explicit at both layers.

## Testing
- `node --test dist/hooks/__tests__/notify-hook-cross-worktree-heartbeat.test.js`

### DIFF
--- a/src/hooks/__tests__/notify-hook-cross-worktree-heartbeat.test.ts
+++ b/src/hooks/__tests__/notify-hook-cross-worktree-heartbeat.test.ts
@@ -51,6 +51,38 @@ function runWorkerNotify(
 }
 
 describe('notify-hook cross-worktree heartbeat resolution', () => {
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Closing this because it targets `main`.

For this repository, PRs must target **`dev`**. Please retarget or reopen the same change against `dev`.

—

PR #1734, opened 2026-04-18
fix: preserve the canonical session across descendant SessionStart events

bee-community-master:
## Summary
- Fix an issue where a descendant subagent started in the same cwd could overwrite the canonical `session.json` entry with its child native session id.
- That session pointer drift could leave the leader session's `ralph-state.json` looking active to later Stop-hook checks.
- Preserve the canonical leader session for descendant SessionStart events and add regression coverage.

## Diagnosis
During reproduction, the following state drift showed up together:
- `.omx/state/session.json` pointed at the subagent session instead of the leader session
- the original leader session still had `sessions/<leader>/ralph-state.json` with `active: true`
- later cleanup / Stop decisions were evaluated against the wrong canonical session, so stale `OMX Ralph is still active` blocks repeated

The root cause was `reconcileNativeSessionStart`: it replaced the canonical session whenever the native session id changed, even when the incoming SessionStart belonged to a descendant process of the still-live leader session.

## Change
- When the incoming SessionStart PID is a **descendant** of the existing canonical session PID, keep the existing root `session.json`
- Keep the previous behavior for a true new top-level session
- Add a regression test that locks the descendant-subagent case

## Verification
- `npm run build`
- `npm run lint`
- `node --test dist/hooks/__tests__/session.test.js dist/scripts/__tests__/codex-native-hook.test.js`

### DIFF
--- a/src/hooks/__tests__/session.test.ts
+++ b/src/hooks/__tests__/session.test.ts
@@ -226,6 +226,36 @@ describe('session lifecycle manager', () => {
     }
   });
 
+  it('preserves the canonical session when a descendant subagent SessionStart arrives in the same cwd', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-session-native-subagent-'));
+    try {
+      await writeSessionStart(cwd, 'omx-leader-session', {
+        nativeSessionId: 'codex-native-leader',
+      });
+
+      const childPid = process.pid + 1000;
+      const reconciled = await reconcileNativeSessionStart(cwd, 'codex-native-child', {
+        pid: childPid,
+        platform: process.platform,
+        readParentPid: (pid) => (pid === childPid ? process.pid : null),
+      });
+
+      assert.equal(reconciled.session_id, 'omx-leader-session');
+      assert.equal(reconciled.native_session_id, 'codex-native-leader');
+
+      const persisted = await readSessionState(cwd);
+      assert.equal(persisted?.session_id, 'omx-leader-session');
+      assert.equal(persisted?.native_session_id, 'codex-native-leader');
+
+      const dailyLogPath = join(cwd, '.omx', 'logs', `omx-${todayIsoDate()}.jsonl`);
+      const dailyLog = await readFile(dailyLogPath, 'utf-8');
+      assert.match(dailyLog, /"event":"session_start_preserved_subagent"/);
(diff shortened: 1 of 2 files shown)


HaD0Yun:
English only plz

bee-community-master:
Updated the PR title/body to English and kept the same diagnosis + verification details. Current readiness snapshot: mergeable=true, no GitHub review threads yet, and no workflow/status checks are attached to the head commit.

HaD0Yun (changes requested):
Thanks for the regression test here — I pulled the branch locally, read the surrounding stop/session flow, and verified that `npm run build`, `npm run lint`, and `node --test dist/hooks/__tests__/session.test.js dist/scripts/__tests__/codex-native-hook.test.js` all pass.

That said, I don't think this is ready to merge yet.

1. **Native Windows still appears to fall back to the old overwrite behavior**
   - In `src/hooks/session.ts`, the new preservation branch depends on `isDescendantProcess(...)`, which in turn depends on `defaultReadParentPid()`.
   - On Linux, `/proc/<pid>/stat` makes this workable.
   - On native `win32`, that path is unavailable, and the fallback uses `ps -o ppid= -p ...`, which is not a reliable Windows primitive.
   - If parent lookup fails, `isDescendantProcess(...)` returns `false`, and we immediately fall back to `writeSessionStart(...)`, which overwrites the canonical session just like before.

   So the current PR fixes the mocked / injected-parent-path case, but it does **not** convincingly fix the production cross-platform case described in the PR summary.

2. **The new heuristic is broader than “descendant subagent”**
   - The preservation branch triggers for **any descendant process** of the existing session PID.
   - That is not equivalent to “this is a Codex subagent”. A fresh top-level Codex launched from a shell spawned by the leader can also be a descendant.
   - In that case we would incorrectly preserve the old canonical session instead of starting a fresh one, which conflicts with the claim that “true new top-level sessions” keep the old behavior.

### What I think is missing before approval
- A real Windows-compatible parent/owner resolution path, or an explicit caller-provided signal that distinguishes subagents from true new sessions.
- A regression test for the native `win32` path without a mocked `readParentPid`.
- A regression test for “new top-level Codex in the same cwd, but still a descendant PID” to prove we do not incorrectly merge sessions.

Because of those gaps, I think the PR description currently overstates what is actually guaranteed by the implementation.

PR #1753, opened 2026-04-19
Require omx question for deep-interview rounds

Yeachan-Heo:
## Summary
- require deep-interview to use `omx question` for every interview round
- reinforce the no-fallback rule in hook guidance, template AGENTS, and contract tests
- harden question state reads against transient truncated JSON during concurrent writes

## Testing
- npm run build
- npm run lint
- npm run check:no-unused
- node --test dist/question/__tests__/state.test.js dist/cli/__tests__/question.test.js dist/hooks/__tests__/deep-interview-contract.test.js dist/question/__tests__/deep-interview.test.js dist/scripts/__tests__/codex-native-hook.test.js
- npm test

### DIFF
--- a/skills/deep-interview/SKILL.md
+++ b/skills/deep-interview/SKILL.md
@@ -51,7 +51,8 @@ If no flag is provided, use **Standard**.
 - Always run a preflight context intake before the first interview question
 - Reduce user effort: ask only the highest-leverage unresolved question, and never ask the user for codebase facts that can be discovered directly
 - For brownfield work, prefer evidence-backed confirmation questions such as "I found X in Y. Should this change follow that pattern?"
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Review result: REJECT for merge in the current state.

- GitHub currently reports this PR as merge-conflicting / dirty against `dev`, so it is not mergeable cleanly even though the lightweight PR check jobs are green.
- The deep-interview contract path itself looks directionally right and narrow: the `deep-interview` skill, `templates/AGENTS.md`, and native hook changes consistently require the OMX question path while keeping stop blocking limited to pending deep-interview question obligations.
- The main code-level blocker beyond conflicts is scope creep in `src/question/state.ts`: the malformed-JSON retry logic broadens the shared question-state reader for a truncation path the real atomic writer does not use. That should be split out instead of riding with the deep-interview contract change.

Please restack / resolve the conflict on top of current `dev` and drop the unrelated shared question-state fallback so this can be re-reviewed as a narrow deep-interview-only change.

—

PR #1771, opened 2026-04-20
Make install/update setup smoother and lower seeded context defaults

Yeachan-Heo:
## Summary

Reduce setup friction after OMX upgrades.
A real global version bump now triggers the interactive setup refresh through a thin best-effort postinstall path, `omx update` exposes the same refresh path explicitly, and fresh OMX-managed `gpt-5.4` configs now seed `250000 / 200000` only when those keys are missing.

## Changes

- add `omx update` help/dispatch and immediate update executor wiring
- add user-scope install stamp tracking and best-effort postinstall bootstrap for real global version bumps
- keep overwrite prompts in `setup.ts` as the sole overwrite authority
- lower seeded `gpt-5.4` defaults to `model_context_window = 250000` and `model_auto_compact_token_limit = 200000`, preserving existing values and backfilling missing partner keys only
- update install/setup docs and add postinstall/update regression coverage

## Validation

- [x] `npm run build`
- [x] `npm test`

Additional verification:
- `npm run check:no-unused`
- `npm run lint` (with temporary empty `bin/` dir to satisfy existing Biome path expectation)
- `npm run smoke:packed-install`
- architect verification: APPROVED

## Checklist

## Related

Closes #

### DIFF
--- a/src/cli/__tests__/setup-refresh.test.ts
+++ b/src/cli/__tests__/setup-refresh.test.ts
@@ -254,8 +254,8 @@ describe("omx setup refresh summary and dry-run behavior", () => {
         config,
         /^# oh-my-codex seeded behavioral defaults \(uninstall removes unchanged defaults\)$/m,
       );
-      assert.match(config, /^model_context_window = 1000000$/m);
-      assert.match(config, /^model_auto_compact_token_limit = 900000$/m);
+      assert.match(config, /^model_context_window = 250000$/m);
+      assert.match(config, /^model_auto_compact_token_limit = 200000$/m);
       assert.match(config, /^# End oh-my-codex seeded behavioral defaults$/m);
     } finally {
       await rm(wd, { recursive: true, force: true });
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Addressed the install-root blocker on follow-up PR #1779.

What changed:
- postinstall now runs the setup refresh from npm's  instead of the global package directory
- added a focused regression test covering the postinstall install-root path

Verification run on the follow-up branch:
- 
> oh-my-codex@0.14.0 build
> node -e "const fs=require('fs'); fs.rmSync('dist',{recursive:true,force:true});" && tsc && node -e "require('fs').chmodSync('dist/cli/omx.js', 0o755)"
- 
> oh-my-codex@0.14.0 check:no-unused
> tsc -p tsconfig.no-unused.json
- manual regression harness against  confirming  lands under the install root and not the global package dir

I kept the fix minimal and did not broaden unrelated install/update semantics.

PR #1774, opened 2026-04-20
refactor(team): simplify pane status inspection rendering

HaD0Yun:
## Summary
Refactor `omx team status` pane-inspection text rendering to remove repetitive formatting logic without changing the operator-facing output contract.

## Problem statement
`src/cli/team.ts` rendered pane-inspection status through a long series of nearly identical loops and ad hoc string fragments. That made the status path hard to review and easy to drift when `readTeamPaneStatus()` grows new inspection fields.

## Root cause
The CLI text renderer duplicated the same output pattern for dozens of `recommended_inspect_*` maps and separately rebuilt the dense `inspect_item_*` summary line inline. The data model was already centralized in `src/team/pane-status.ts`, but the rendering path was not.

## What changed
- Added a shared `TeamPaneStatus` alias and `TeamInspectItem` item type for the text renderer.
- Added `logInspectEntries()` to handle the repeated `inspect_<field>_<target>` emission pattern.
- Added `formatInspectItemLine()` to centralize the dense `inspect_item_<n>` line formatting.
- Replaced the long manual rendering chain in `renderTeamPaneStatus()` with helper-driven output while preserving the existing emitted keys and value formatting.

## Why this design
This is the smallest cleanup that removes the duplicated rendering slop while preserving behavior. It keeps the underlying pane-status payload intact and avoids broad API or contract changes.

## Overlap assessment
Adjacent but distinct.
- Related history: #760 introduced `omx sparkshell` and team inspection metadata, including the pane-status inspection surface.
- This PR does not add new inspection capabilities. It only deslopifies the existing text rendering path that prints that metadata.
- I did not find an existing issue or PR specifically targeting `renderTeamPaneStatus`, `recommended_inspect_*` text rendering, or `inspect_item_*` formatting cleanup.

## Validation
- `npm run build`
- `npm run check:no-unused`
- `./node_modules/.bin/biome lint src`
- `node --test dist/cli/__tests__/team.test.js --test-name-pattern='teamCommand status'`

## Risk / compatibility
- Intended behavior change: none.
- Output contract risk is low because the targeted team status regression tests stayed green after the cleanup.
- Residual repo baseline note: the repository's `npm run lint` script currently targets `biome lint src bin`, but this checkout does not contain a top-level `bin/` directory. I did not change that pre-existing baseline issue in this scoped cleanup PR.

## Files changed
- `src/cli/team.ts` — consolidated duplicated text rendering into shared helpers

### DIFF
--- a/src/cli/team.ts
+++ b/src/cli/team.ts
@@ -445,8 +445,97 @@ function buildDeadWorkerAwaitEvent(teamName: string, snapshot: TeamSnapshot): Te
 }
 
 
+type TeamPaneStatus = Awaited<ReturnType<typeof readTeamPaneStatus>>;
+type TeamInspectItem = TeamPaneStatus['recommended_inspect_items'][number];
+
+function logInspectEntries<T>(
+  prefix: string,
+  entries: Record<string, T>,
+  shouldLog: (value: T) => boolean,
+  formatValue: (value: T) => string = (value) => String(value),
+): void {
+  for (const [target, value] of Object.entries(entries)) {
+    if (shouldLog(value)) {
+      console.log(`${prefix}_${target}: ${formatValue(value)}`);
+    }
+  }
+}
+
+function formatInspectItemLine(index: number, item: TeamInspectItem): string {
+  const parts = [
+    `inspect_item_${index + 1}:`,
+    `target=${item.target}`,
+    item.pane_id ? `pane=${item.pane_id}` : '',
+    item.worker_cli ? `cli=${item.worker_cli}` : '',
+    item.role ? `role=${item.role}` : '',
+    typeof item.index === 'number' ? `index=${item.index}` : '',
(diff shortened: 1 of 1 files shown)


HaD0Yun (commented):
Non-blocking maintainability note:

This refactor removes the big inline duplication nicely, but `renderTeamPaneStatus()` still has a very long sequence of `logInspectEntries(...)` calls for each `recommended_inspect_*` field.

That means every new inspect field now has to be wired in at least two places:
1. `readTeamPaneStatus()` produces the structured field
2. `renderTeamPaneStatus()` adds a matching `logInspectEntries(...)` call

If this inspection surface keeps growing, it may be worth one more small cleanup pass that moves those mappings into a descriptor table (for example: `prefix`, `entries`, `shouldLog`, optional `formatValue`) and iterates it once. That would keep the current output contract while making future additions less error-prone.

PR #1786, opened 2026-04-20
Fix CI rust-toolchain action resolution

Yeachan-Heo:
## Summary\n- switch workflow uses of `dtolnay/rust-toolchain` from the failing moving `stable` ref to immutable `v1`\n- pass `toolchain: stable` explicitly so CI/release behavior stays on the stable Rust channel\n- limit the change to workflow setup steps in CI and release jobs\n\n## Verification\n- parsed updated workflow YAML locally\n- confirmed current upstream refs: `stable` -> `29eef336d9b2848a0b548edc03f92a220660cdb8`, `v1` -> `e97e2d8cc328f1b50210efc529dca0028893a2d9`\n- verified all workflow consumers now reference `dtolnay/rust-toolchain@v1`\n\n## Context\nPR #1771 is currently failing CI in the Rust Clippy setup step because GitHub Actions cannot resolve the existing pinned tarball for the moving `stable` action ref. This PR isolates the workflow fix against `dev`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Setup Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@v1
         with:
+          toolchain: stable
           components: rustfmt
       - name: Check Rust formatting
         run: cargo fmt --all --check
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Follow-up CI investigation summary:

- Reproduced the failing `Test (Node 20 / hooks-notify-platform)` lane locally under Node 20.
- Root cause was **not** the hooks/notify runtime itself; it was `src/verification/__tests__/ci-rust-gates.test.ts` still asserting `dtolnay/rust-toolchain@stable`.
- PR #1786 had already correctly changed the workflows to `dtolnay/rust-toolchain@v1` with `toolchain: stable`, so the contract test became stale.
- `Coverage Report (TypeScript Full)` was downstream noise from the same single failing test in the full suite, not an independent coverage regression.

Fix pushed on this PR branch in `3114a070`.

PR #1814, opened 2026-04-21
Fix false-positive PostToolUse failure detection from stdout text

kfiramar:
## Summary
- stop `PostToolUse` from treating successful stdout text as authoritative failure diagnostics
- restrict MCP transport-death detection to error/diagnostic channels instead of stdout payloads
- add regressions for successful Bash/MCP outputs that happen to contain `command not found` / transport-closure wording

## Why
The current hook logic can misclassify benign tool output as a hard failure. In practice that means successful commands can trigger noisy `PreToolUse/PostToolUse hook (failed)` or misleading recovery guidance when stdout merely echoes phrases like `command not found` or `mcp transport closed`.

## Verification
- `npm run build`

## Risk
- narrow behavior change in native hook post-processing only
- preserves existing remediation guidance for real stderr / structured-error failure paths

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -1912,6 +1912,58 @@ esac
     }
   });
 
+  it("stays silent when successful Bash stdout merely mentions command-not-found text", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-posttool-command-not-found-stdout-"));
+    try {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the patch. Closing this one because it targets  instead of , and we are not taking this path as-is. If you want this reconsidered, reopen a narrowly scoped PR against .\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #1833, opened 2026-04-22
fix: fail closed when `omx question` runs under detached tmux sessions

### LINKED ISSUE #1832: `omx question` still opens a hidden UI from detached tmux sessions
## Summary

`omx question` can still open a hidden question UI when it is launched from a **detached tmux session**, even on current `dev` / `main` after the `#1808` fail-closed change.

The current guard treats `TMUX` presence as enough to allow the inside-tmux renderer. In a detached session that means `omx question` still splits a pane and blocks on a UI that no operator can see.

## Environment

- repo: `Yeachan-Heo/oh-my-codex`
- tested from source on `dev` head `fc738134` and `main`/`v0.14.2` behavior before the fix branch
- macOS
- tmux available

## Reproduction

```bash
cd /path/to/oh-my-codex
npm ci
npm run build

payload='{"question":"Detached session invisible?","options":[{"label":"Yes","value":"yes"}]}'
tmux new-session -d -s omx_detached_qtest \
  "cd '$PWD' && OMX_SESSION_ID=detached-qtest node dist/cli/omx.js question --input '$payload' --json >/tmp/omx_detached_qtest.out 2>/tmp/omx_detached_qtest.err"
```

After that, inspect the session:

```bash
tmux list-sessions -F '#{session_name}\t#{session_attached}' | grep omx_detached_qtest
# omx_detached_qtest    0

tmux list-panes -t omx_detached_qtest -F '#{pane_id}\tdead=#{pane_dead}\tactive=#{pane_active}\tcmd=#{pane_current_command}'
# two node panes exist

tmux capture-pane -p -t <question-pane>
```

Captured hidden pane content:

```text
Detached session invisible?

› [x] 1. Yes
  [ ] 2. Other

Use ↑/↓ to move, Enter to select.
```

Observed question record:

```json
{
  "status": "prompting",
  "renderer": {
    "renderer": "tmux-pane",
    "target": "%142",
    "return_target": "%141",
    "return_transport": "tmux-send-keys"
  }
}
```

`/tmp/omx_detached_qtest.out` and `.err` remained empty while the process stayed blocked.

## Expected behavior

If the current tmux session has no attached client / no visible operator, `omx question` should fail closed with the same attached-pane error instead of spawning a hidden pane.

## Actual behavior

A hidden question UI is launched inside a detached tmux session, and the command blocks waiting for input the operator cannot see.

## Why this seems distinct from #1808

PR #1808 fixed the case where a tmux binary was available but the caller was not inside tmux at all.

This repro is different:
- the process **is** inside tmux,
- but the tmux session is **detached** (`session_attached=0`),
- so the pane is still invisible to the operator.

## Suggested fix

Treat `inside tmux but detached/no attached client` the same as unsupported for question rendering, unless there is an explicit detached-session workflow for `omx question`.

A small targeted regression test around `#{session_attached}` should be enough to keep this from reappearing.

shaun0927:
## Summary
- fail closed when `omx question` is launched from a tmux session with no attached client
- keep the existing attached-tmux split-pane flow unchanged for normal interactive usage
- add renderer and CLI regressions so detached sessions no longer create hidden question panes

## Verification
- `npm run build`
- `node --test dist/question/__tests__/renderer.test.js dist/cli/__tests__/question.test.js`
- `npx biome lint src/question/renderer.ts src/question/__tests__/renderer.test.ts src/cli/__tests__/question.test.ts`
- `git diff --check`
- manual detached tmux repro: `tmux new-session -d ... node dist/cli/omx.js question ... --json` now returns `question_runtime_failed` JSON instead of spawning a hidden pane

## Why this scope
PR #1808 correctly fixed the non-tmux case, but detached tmux sessions still satisfy the current `TMUX`-only gate. This patch keeps the same fail-closed direction and narrows the gap by checking whether the current tmux session actually has an attached client.

Closes #1832

### DIFF
--- a/src/cli/__tests__/question.test.ts
+++ b/src/cli/__tests__/question.test.ts
@@ -119,6 +119,13 @@ describe('omx question CLI', () => {
     await writeFile(join(fakeBinDir, 'tmux'), `#!/bin/sh
 printf '%s\\n' "$*" >> "${join(cwd, 'tmux.log')}"
 case "$1" in
+  display-message)
+    if [ "$2" = "-p" ] && [ "$3" = "-t" ] && [ "$4" = "%0" ] && [ "$5" = "#{session_attached}" ]; then
+      printf '1\\n'
+      exit 0
+    fi
+    printf '%%0\\n'
+    ;;
   split-window)
     printf '%%5\\n'
     ;;
@@ -128,9 +135,6 @@ case "$1" in
       exit 1
     fi
     printf '%%0\\t1\\n%%2\\t0\\n'
-    ;;
-  display-message)
-    printf '%%0\\n'
     ;;
 esac
 `, { mode: 0o755 });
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the PR. This detached-tmux fail-closed path has already been fixed and merged via #1835, so I'm closing this as duplicate/superseded.

—

PR #1857, opened 2026-04-23
fix: avoid misclassifying empty bash exits as setup failures

heyclf:
## Summary
- stop treating parsed-but-empty non-zero Bash tool responses as command/setup failures
- only fall back to raw tool output when the response was not parsed into stdout/stderr fields
- add a regression test for exit-1 Bash commands with empty stdout/stderr

## Why
Commands like `rg` legitimately return exit code 1 when no matches are found. Their tool payload can be parsed successfully with empty stdout/stderr, but the previous classifier fell back to the raw JSON string and treated that as informative output, which then surfaced misleading PostToolUse guidance.

## Testing
- npm run build

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -1912,6 +1912,28 @@ esac
     }
   });
 
+  it("does not classify empty-output exit-1 Bash commands as setup failures", async () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the patch. I'm closing this PR because it targets `main`, and this repo takes PRs against `dev` for normal integration flow.

If you want to resubmit, please reopen the change against `dev` and we can look at it there.

—

PR #1861, opened 2026-04-23
Reinforce regression coverage for prompt-worker terminal guards

HaD0Yun:
## Summary
Retitle and reframe this PR as regression coverage only for the existing prompt-worker terminal guard behavior.

## Scope
This PR does **not** introduce a new production/runtime completion guard.

The live-worker terminal guard already exists on the base branch in `monitorTeam()`:
- `terminalBlockedByLiveWorkers` is derived from live worker status
- that signal is passed into `inferPhaseTargetFromTaskCounts(...)`
- the phase is held at `team-verify` while a live worker still reports an active runtime state

This branch only updates regression coverage so the tests match the real prompt-worker liveness contract used by that existing runtime path.

## Why the previous framing was misleading
The previous title/body implied that this branch added the production behavior that prevents terminal completion while live workers are still active.

After re-checking the diff against `pr/base-local-dev-upstream`, that is not what this branch changes. Relative to the PR base, the branch is a small follow-up that tightens tests around behavior already present on the base.

## What actually changed
- updated the prompt-worker regression setup so prompt workers are treated as alive using the same contract the runtime uses in production
- explicitly enabled prompt launch mode in the affected tests
- seeded worker pid/heartbeat/state so the tests exercise the real liveness gate instead of an under-specified fixture
- kept the existing runtime behavior assertions:
  - `monitorTeam()` stays at `team-verify` while a live prompt worker is still `working`
  - runtime CLI helper coverage keeps that state from being treated as an immediate shutdown condition

## Files changed
- `src/team/__tests__/runtime-cli.test.ts`
- `src/team/__tests__/runtime.test.ts`

## Validation
- `npm run build`
- `node --test dist/team/__tests__/runtime-cli.test.js`
- `node --test --test-name-pattern="monitorTeam withholds terminal phase while a live worker still reports active status" dist/team/__tests__/runtime.test.js`

## Reviewer note
This PR should be reviewed as **regression coverage reinforcement for existing runtime behavior**, not as a new production runtime fix.

### DIFF
--- a/src/team/__tests__/runtime-cli.test.ts
+++ b/src/team/__tests__/runtime-cli.test.ts
@@ -105,8 +105,14 @@ describe('runtime-cli helpers', () => {
     const previousTeamStateRoot = process.env.OMX_TEAM_STATE_ROOT;
     const previousLaunchMode = process.env.OMX_TEAM_WORKER_LAUNCH_MODE;
     delete process.env.OMX_TEAM_STATE_ROOT;
+    process.env.OMX_TEAM_WORKER_LAUNCH_MODE = 'prompt';
     try {
       await initTeamState('runtime-cli-complete', 'task', 'executor', 1, cwd);
+      const config = await readTeamConfig('runtime-cli-complete', cwd);
+      assert.ok(config);
+      if (!config) return;
+      config.workers[0]!.pid = process.pid;
+      await saveTeamConfig(config, cwd);
       await createTask('runtime-cli-complete', {
         subject: 'done task',
         description: 'already complete',
--- a/src/team/__tests__/runtime.test.ts
+++ b/src/team/__tests__/runtime.test.ts
@@ -1848,9 +1848,16 @@ process.on('SIGTERM', () => {
   it('monitorTeam withholds terminal phase while a live worker still reports active status', async () => {
     const cwd = await mkdtemp(join(tmpdir(), 'omx-runtime-live-worker-terminal-'));
     const prevTeamStateRoot = process.env.OMX_TEAM_STATE_ROOT;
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
Thanks — I reviewed this in an OMX review session.

Requesting changes because the current diff does not appear to implement the production/runtime behavior implied by the title. The live-worker terminal guard path already exists on the base; the current PR looks like regression coverage reinforcement, not the runtime change suggested by "Guard team completion while live workers still own runtime state".

What I verified:
- exact scope is test-heavy (`src/team/__tests__/runtime-cli.test.ts`, `src/team/__tests__/runtime.test.ts`) with no meaningful production behavior change in `src/team/runtime.ts`
- targeted local runtime tests passed
- PR checks on the PR itself are thin (Draft/size only), so the main blocker is scope/intent mismatch rather than CI noise

Please either:
1. retitle/reframe this as regression coverage only, or
2. add the actual production/runtime change that enforces the behavior described by the title.

—

HaD0Yun:
Thanks — I re-checked the diff against the PR base and you were right that the previous framing overstated the production scope. I have now retitled and rewritten the PR in English to make the scope explicit: this branch is regression coverage only, not a new runtime behavior change. The live-worker terminal guard already exists on the base in `monitorTeam()`; this follow-up just aligns the prompt-worker tests with the real liveness contract (prompt mode + worker pid/heartbeat/state) so that existing behavior is covered accurately. Please re-review it as a regression-coverage PR rather than a production runtime fix.

Yeachan-Heo (approved):
Re-reviewed after a544bc4. Previous blocker is resolved: against pr/base-local-dev-upstream, the net diff is now test-only and no production/runtime delta remains. Merge-ready as regression coverage reinforcement for existing behavior.

—

PR #1864, opened 2026-04-23
Add ralplan conversational mention regression coverage

utgpay2:
## Summary
- add a keyword-detector regression test for plain conversational `ralplan` mentions
- assert prompt-submit activation does not seed `skill-active-state` or `ralplan-state` for that prose-only case
- lock down the current mainline behavior so future keyword-detector changes cannot silently reintroduce Stop-hook blockers

## Why
Current `main` already avoids treating a bare conversational `ralplan` mention as a workflow activation, but that guarantee was not covered by regression tests. This PR codifies the behavior so future registry or activation changes do not accidentally seed planning state from incidental prose.

## Testing
- `npm run build`
- `npm run lint -- src/hooks/__tests__/keyword-detector.test.ts`
- `node --test dist/hooks/__tests__/keyword-detector.test.js`
- `npm run check:no-unused`

### DIFF
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -147,6 +147,11 @@ describe('keyword detector swarm/team compatibility', () => {
     assert.equal(match, null);
   });
 
+  it('does not trigger ralplan from plain conversational mention', () => {
+    const match = detectPrimaryKeyword('why does ralplan keep blocking stop?');
+    assert.equal(match, null);
+  });
+
   it('still triggers ralph for explicit $ralph invocation', () => {
     const match = detectPrimaryKeyword('$ralph continue verification');
     assert.ok(match);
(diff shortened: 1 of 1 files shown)


Yeachan-Heo (changes requested):
Requesting changes because this PR targets `main`, while oh-my-codex PRs must target `dev` per repo policy.

Code scope itself looks appropriately narrow and test-only for the stated regression: I reviewed the keyword-detector coverage and did not find an additional user-facing contract blocker in the diff. But the base branch is a hard blocker, so this is not mergeable in its current form. Please retarget to `dev` (or reopen correctly against `dev`) and keep the same narrow regression-only scope.

—

Yeachan-Heo:
Closing this in its current form because it targets `main` instead of `dev`, so it is not mergeable under this repo's contribution rules. If you want to resubmit, please reopen the same narrow regression-only change against `dev`.

—

PR #548, opened 2026-03-03
fix(quality): safe JSON.parse, catch block logging, NaN guards, type safety, TOCTOU fixes (#505, #506, #507, #509, #519)

junhoyeo:
## Summary
Cross-cutting code quality sweep fixing 5 issues across ~15 files.

## Issues Fixed
- **#506**: Add shared `safeJsonParse`/`safeReadJsonFile` utility (`src/utils/safe-json.ts`). Wrap 14+ bare `JSON.parse` calls in try/catch across MCP servers, CLI, HUD, hooks, and notifications.
- **#505**: Add stderr logging to 24+ silent catch blocks across `runtime.ts`, `team-server.ts`, `state-server.ts`, `code-intel-server.ts`, `trace-server.ts`, `agents-overlay.ts`, `cli/index.ts`, `openclaw/dispatcher.ts`.
- **#507**: Add `isNaN`/`Number.isFinite` guards for `parseInt`/`parseFloat` in PID parsing, Discord rate limit header, and Node version check.
- **#509**: Replace unsafe `handle!.pid as number` with null-safe access, `map.get(dir)!` with conditional, `activeSession!.pid` with optional chaining, `as any` with typed interface.
- **#519**: Replace `existsSync + readFile` TOCTOU pattern with atomic `try { readFile } catch (ENOENT)` in `readWorkerHeartbeat`, `readWorkerStatus`, `readShutdownAck`.

## Verification
- `npm run build` passes
- `lsp_diagnostics` clean on all changed files

Fixes #505, Fixes #506, Fixes #507, Fixes #509, Fixes #519

Issue #506, opened 2026-03-03
[Code Quality] 14+ unchecked JSON.parse calls can crash MCP servers and CLI


junhoyeo:
## Summary

Multiple `JSON.parse` calls on external/user-controlled files lack try/catch protection, causing unhandled exceptions that crash MCP servers and CLI commands.

## Locations

**MCP servers (crashes entire server process):**
- `src/mcp/memory-server.ts` — lines 188, 202, 216, 233: All four memory MCP tool handlers crash on corrupt `project-memory.json`
- `src/mcp/trace-server.ts` — line 192: Malformed `metrics.json` crashes trace server

**CLI (crashes user-facing commands):**
- `src/cli/index.ts` — line 447: `JSON.parse(content)` in `showStatus()` — malformed state file aborts `omx status`
- `src/cli/index.ts` — line 1248: `JSON.parse(content)` in `cancelModes()` — malformed state file aborts mode cancellation
- `src/cli/version.ts` — line 9: Malformed `package.json` crashes `omx version`
- `src/hud/state.ts` — line 87: Malformed `package.json` crashes HUD

**Hooks (crashes hook pipeline):**
- `src/hooks/agents-overlay.ts` — lines 129, 166, 204: No try/catch on state file parsing
- `src/notifications/notifier.ts` — line 49: No try/catch on config file parsing

## Pattern

```typescript
// Common dangerous pattern:
const data = JSON.parse(await readFile(path, 'utf-8'));
// No try/catch — if file is malformed, unhandled exception
```

## Impact

A single corrupt JSON file (from partial writes, concurrent access, or disk errors) can crash MCP servers, CLI commands, or the hook pipeline.

## Recommended Fix

1. Wrap all `JSON.parse` calls on external files in try/catch
2. Return sensible defaults on parse failure (empty object, null, etc.)
3. Log parse errors for debugging
4. Consider a shared `safeParseJSON(path, fallback)` utility

PR #1918, opened 2026-04-25
fix(runtime): guard RuntimeBridge JSON.parse against partial output

lifrary:
## Summary

`RuntimeBridge.execCommand`, `readSnapshot`, and `readCompatFile` each call `JSON.parse` with no try/catch. A partial Rust write (mid-rename, `kill -9`, schema drift, panic-before-flush) surfaces as a `SyntaxError` that aborts the HUD render path or the dispatch watcher loop instead of degrading gracefully.

This builds on #548 (`fix(quality): safe JSON.parse, …`) and #506. `RuntimeBridge` was not in #548's 16-file scope but exhibits the same hazard.

## Reproducer

```bash
OMX_RUNTIME_BRIDGE=1 omx team run alpha
# In another shell, while the dispatcher is writing:
printf '{' > .omx/state/team/alpha/.omx-runtime/snapshot.json
omx hud
```

**Before**: `SyntaxError: Unexpected end of JSON input` aborts the entire `buildAllState` query at `src/hud/state.ts:458`.
**After**: HUD falls through to JS-inferred state for this tick.

## Fix

| Method | Behaviour on parse failure |
|---|---|
| `execCommand` | Throws typed `RuntimeBridgeError` so dispatch callers (`src/team/state/dispatch.ts:169`, `src/team/state/mailbox.ts:44`) can recognise parse failure with `instanceof` and mark the command failed without tearing down the surrounding watcher. |
| `readSnapshot` | Same typed error — preserves the failing command name plus a 200-byte `stdoutPreview` and the underlying `cause`. |
| `readCompatFile` | Returns `null`. The two production callers (`src/hud/state.ts:454-458`, `src/team/state.ts:1488`) already initialise to `null` (`runtimeSnapshot`) or use optional chaining (`existing?.records`) and treat `null` as the legacy-path fallback, so this is behaviour-compatible. |

`safeJsonParse` from `src/utils/safe-json.ts` (introduced by #548) is reused for the compat-reader path; `execCommand` / `readSnapshot` use direct try/catch because they need to surface a typed error, not a fallback.

## Tests added

`src/runtime/__tests__/bridge.test.ts` (+165 lines, 14 new assertions across 3 suites):
- `RuntimeBridgeError`: `instanceof`, `.name`, context fields (`command`, `stdoutPreview`, `cause`), default empty context
- `RuntimeBridge.readCompatFile (parse guard)`: missing `stateDir`, missing file, truncated `{`, empty content, well-formed JSON
- `RuntimeBridge.execCommand / readSnapshot (parse guard)`: non-JSON stdout → typed error with `command='AcquireAuthority'` + `cause instanceof SyntaxError`, well-formed `RuntimeEvent` → parses normally, `readSnapshot` non-JSON → typed error with `command='snapshot'`

The exec/snapshot tests use a temp shell script as the runtime binary, branching on `$1` so the `validateSchemaOnce` probe always sees a valid manifest (the `schemaValidated` module singleton would otherwise make test ordering significant).

## Verification

- [x] `npm run build`
- [x] `npm run lint` (553 files)
- [x] `node --test dist/runtime/__tests__/bridge.test.js` — 14/14 pass
- [x] `npm test` — pre-existing failures observed are all `gpt-5.5` frontier-default fixture mismatches from the recent 0.14.4 cut (e.g. `dist/team/__tests__/runtime.test.js:312` expects `gpt-5.5`, gets `gpt-5.4`); none are in `runtime/bridge.ts` and the bridge test suites all pass.

## Backward compatibility

- `RuntimeBridgeError extends Error`. Existing `try { … } catch (err) { … }` blocks at any call site still catch identically; only `instanceof` adopters see new typing.
- `execCommand` / `readSnapshot` previously threw a vanilla `SyntaxError` on this input; now throw `RuntimeBridgeError`. Both production exec call sites (`team/state/dispatch.ts`, `team/state/mailbox.ts`) ignore the return value and the throw propagates the same way.
- `readCompatFile` previously threw on parse failure; now returns `null`. Both callers already handled `null`, so this strictly reduces the failure surface.

## Test plan

- [x] Unit tests cover all three guarded paths plus the typed-error contract
- [x] Verified callers (`hud/state.ts`, `team/state.ts`) tolerate `null` returns
- [x] No platform-specific behaviour beyond the existing `execFileSync` (POSIX shell fixture for binary mock; same OS scope as existing bridge tests)

### DIFF
--- a/src/runtime/__tests__/bridge.test.ts
+++ b/src/runtime/__tests__/bridge.test.ts
@@ -1,6 +1,9 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
-import { resolveRuntimeBinaryPath } from '../bridge.js';
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { RuntimeBridge, RuntimeBridgeError, resolveRuntimeBinaryPath } from '../bridge.js';
 
 describe('resolveRuntimeBinaryPath', () => {
   it('prefers explicit OMX_RUNTIME_BINARY override', () => {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
Approved. This is a narrow runtime bug fix: it converts malformed/non-JSON RuntimeBridge output into typed bridge errors or null compat fallback without expanding the runtime surface. CI is green and the added coverage matches the guarded JSON boundary.

—

PR #1941, opened 2026-04-26
fix(config): preserve user [tui] status_line across merges

50sotero:
Fixes #1940.

## Summary

`omx update` (and even no-op stamp-bump refreshes) was silently overwriting the user's customized `status_line` in `~/.codex/config.toml` with the OMX hard-coded default. This PR makes the merge non-destructive: the user's `status_line` is preserved like every other `[tui]` key, and the OMX default is only injected when the user has none.

## Root cause

`upsertTuiStatusLine()` in `src/config/generator.ts` explicitly excluded `status_line` from the preserved `[tui]` keys and then unconditionally re-appended `OMX_TUI_STATUS_LINE`:

- `src/config/generator.ts:716` — `if (key === "status_line" || seenKeys.has(key)) continue;` — drops the user's `status_line`.
- `src/config/generator.ts:722` — `const mergedSection = ["[tui]", ...preservedKeyLines, OMX_TUI_STATUS_LINE];` — always re-injects the OMX default.

Update paths that triggered this on every release / refresh:

- `src/cli/update.ts:395` — `runSetupRefresh()` is called after every successful `npm install -g oh-my-codex@latest`.
- `src/cli/update.ts:354-359` — `runSetupRefresh()` is also called when the package is *already up to date* if `setup_completed_version` differs from the installed version.

Both reach `buildMergedConfig()` → `upsertTuiStatusLine()`, so users saw their customizations vanish on every update — including no-op ones.

### `src/config/generator.ts`
- Stop excluding `status_line` from `preservedKeyLines`.
- Track whether the user already has a `status_line`.
- Append `OMX_TUI_STATUS_LINE` only when the user has none.

```diff
   const preservedKeyLines: string[] = [];
   const seenKeys = new Set<string>();
+  let hasUserStatusLine = false;

   for (const section of sections) {
     for (let i = section.start + 1; i < section.end; i++) {
       ...
       const key = keyMatch[1];
-      if (key === "status_line" || seenKeys.has(key)) continue;
+      if (seenKeys.has(key)) continue;
       seenKeys.add(key);
+      if (key === "status_line") hasUserStatusLine = true;
       preservedKeyLines.push(trimmed);
     }
   }

-  const mergedSection = ["[tui]", ...preservedKeyLines, OMX_TUI_STATUS_LINE];
+  const mergedSection = hasUserStatusLine
+    ? ["[tui]", ...preservedKeyLines]
[...]

PR #1942, opened 2026-04-26
Preserve user TUI status lines during setup refresh

### LINKED ISSUE #1940: [bug] User's [tui] status_line in ~/.codex/config.toml is wiped on every `omx update` / setup refresh
## Summary

Every time `omx update` runs (or `omx setup` is triggered, or even when the package is "already up to date" but the install-stamp version differs), the user's customized `status_line` in the `[tui]` section of `~/.codex/config.toml` is **silently overwritten** with the OMX hard-coded default.

This effectively means: any user who customizes `/statusline` in Codex sees their configuration disappear after every oh-my-codex update.

## Reproduction

1. Set a custom statusline, e.g. in `~/.codex/config.toml`:
   ```toml
   [tui]
   theme = "night"
   status_line = ["git-branch", "model-with-reasoning"]
   ```
2. Run `omx update` (or wait for it to be triggered automatically).
3. Open `~/.codex/config.toml` again — `status_line` has been replaced with:
   ```toml
   status_line = ["model-with-reasoning", "git-branch", "context-remaining", "total-input-tokens", "total-output-tokens", "five-hour-limit", "weekly-limit"]
   ```

The user-specified value is gone. Other `[tui]` keys like `theme` are preserved correctly — only `status_line` is destroyed.

## Root Cause

The destructive write lives in `upsertTuiStatusLine()` in `src/config/generator.ts`.

**`src/config/generator.ts:63-64`** — the OMX default is hard-coded:
```ts
const OMX_TUI_STATUS_LINE =
  'status_line = ["model-with-reasoning", "git-branch", "context-remaining", "total-input-tokens", "total-output-tokens", "five-hour-limit", "weekly-limit"]';
```

**`src/config/generator.ts:715-722`** — when merging, the function explicitly **excludes** any user-defined `status_line` from the preserved keys, then unconditionally appends the OMX default:
```ts
const key = keyMatch[1];
if (key === "status_line" || seenKeys.has(key)) continue;  // <-- drops user's status_line
seenKeys.add(key);
preservedKeyLines.push(trimmed);
// ...
const mergedSection = ["[tui]", ...preservedKeyLines, OMX_TUI_STATUS_LINE]; // <-- always injects default
```

So the merge logic is hard-wired to replace the user's `status_line`, with no opt-out.

## How an "update" triggers it

`buildMergedConfig()` (which calls `upsertTuiStatusLine`) is invoked from the setup-refresh path that `omx update` runs automatically:

- **`src/cli/update.ts:395`** — after a successful `npm install -g oh-my-codex@latest`, `runSetupRefresh()` is called unconditionally.
- **`src/cli/update.ts:354-359`** — even when the package is *already* up to date, if `setup_completed_version` does not match the installed version, `runSetupRefresh()` is invoked anyway.
- The refresh ultimately reaches `buildMergedConfig()` → `upsertTuiStatusLine()` → user's `status_line` is destroyed.

This is why users see the value vanish "every time there is an update" — including on no-op updates that just bump the install stamp.

## Test that documents the bug as intended

`src/config/__tests__/generator-idempotent.test.ts:361-386` currently asserts the broken behavior as if it were correct:

```ts
it("merges OMX status_line into an existing user [tui] section without duplicating the table", async () => {
  // user starts with: status_line = ["git-branch"]
  // ...
  assert.match(
    toml,
    /^status_line = \["model-with-reasoning", "git-branch", "context-remaining", ...\]$/m,
    "status_line updated in-place",
  );
});
```

This test should instead assert that the user's `status_line` is preserved when one exists.

## Expected behavior

- If the user has **no** `status_line` in `[tui]` (or no `[tui]` section at all): inject the OMX default. (unchanged)
- If the user already has a `status_line` in `[tui]`: **preserve it**. Do not overwrite.

This matches the way every other user-owned key in `[tui]` (e.g. `theme`) is already handled.

## Fix

We will provide the fix in a git branch. The patch will:

1. Stop dropping `status_line` from `preservedKeyLines` in `upsertTuiStatusLine()` (`src/config/generator.ts:716`).
2. Only append `OMX_TUI_STATUS_LINE` when no user `status_line` exists (`src/config/generator.ts:722`).
3. Update `src/config/__tests__/generator-idempotent.test.ts:361-386` to assert preservation, plus add a regression test for the "user has `[tui]` but no `status_line`" path so the OMX default still gets injected for fresh users.

A follow-up comment on this issue will link the branch with the fix once pushed.

## Environment

- oh-my-codex version: `0.15.0` (current `main`)
- Files implicated: `src/config/generator.ts`, `src/cli/update.ts`, `src/config/__tests__/generator-idempotent.test.ts`

Yeachan-Heo:
## Summary
- Preserve user-owned `[tui].status_line` entries during `omx setup/update` config refresh, including multiline TOML arrays.
- Keep OMX default injection when `[tui]` exists without `status_line`.
- Add regression coverage for existing `[tui]` without `status_line` and multiline user-owned `status_line` preservation.

## Verification
- `npm run build`
- `node --test dist/config/__tests__/generator-idempotent.test.js`
- `npm run lint`
- `npm run build`
- `npx tsc --noEmit --pretty false --project tsconfig.json` (0 diagnostics)

## Notes
- External PR #1941 targeted `main` and was closed; this PR is the `dev`-based fix.

Fixes #1940

🤖 Generated with OpenAI Codex

### DIFF
--- a/src/config/__tests__/generator-idempotent.test.ts
+++ b/src/config/__tests__/generator-idempotent.test.ts
@@ -396,6 +396,53 @@ describe("config generator idempotency (#384)", () => {
     );
   });
 
+  it("seeds the default status_line into an existing [tui] section without one", () => {
+    const toml = buildMergedConfig(
+      ["[tui]", 'theme = "night"', ""].join("\n"),
+      "/tmp/omx",
+    );
+
+    assert.equal(count(toml, /^\[tui\]$/gm), 1, "[tui] should appear once");
+    assert.match(toml, /^theme = "night"$/m, "existing tui key preserved");
+    assert.match(
+      toml,
+      /^status_line = \["model-with-reasoning", "git-branch", "context-remaining", "total-input-tokens", "total-output-tokens", "five-hour-limit", "weekly-limit"\]$/m,
+      "default status_line should be seeded when [tui] lacks one",
+    );
+  });
+
+  it("preserves a multiline user-owned status_line", () => {
+    const toml = buildMergedConfig(
+      [
+        "[tui]",
+        "status_line = [",
+        '  "git-branch",',
+        '  "context-remaining",',
+        "]",
+        "",
+      ].join("\n"),
+      "/tmp/omx",
+    );
+
+    assert.equal(count(toml, /^\[tui\]$/gm), 1, "[tui] should appear once");
+    assert.ok(
+      toml.includes(
+        [
+          "status_line = [",
+          '"git-branch",',
+          '"context-remaining",',
+          "]",
+        ].join("\n"),
+      ),
+      "multiline user status_line should be preserved",
+    );
+    assert.doesNotMatch(
+      toml,
+      /^status_line = \["model-with-reasoning", "git-branch", "context-remaining", "total-input-tokens", "total-output-tokens", "five-hour-limit", "weekly-limit"\]$/m,
+      "default status_line should not overwrite multiline customization",
+    );
+  });
+
   it("preserves a customized managed-block status_line when refreshing setup", () => {
     const firstRun = buildMergedConfig("", "/tmp/omx");
     const customized = firstRun.replace(
--- a/src/config/generator.ts
+++ b/src/config/generator.ts
@@ -762,7 +762,15 @@ function upsertTuiStatusLine(config: string): {
 
       const key = keyMatch[1];
       if (key === "status_line") {
-        preservedStatusLine ??= trimmed;
+        const entryLines = [trimmed];
+        while (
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
OMX review pass completed for PR #1942. Diff is limited to config generator status_line preservation and idempotency coverage. Targeted verification passed locally: `npm run build` then `node --test dist/config/__tests__/generator-idempotent.test.js` (37 pass). CI rollup is green. Approved for dev merge; note this is user-facing setup/config behavior, so owner confirmation is still required before merge per contract-change rule.

—

PR #1953, opened 2026-04-27
feat(config): preset-driven [tui] status_line (minimal | focused | full)

50sotero:
Re-opening after the previous PR (#1951) was closed; this branch contains both the original feature and the review fixes from @Yeachan-Heo and @chatgpt-codex-connector.

> Heads-up to the maintainer: you flagged on #1951 that this PR changes a user-facing config contract (`.omx/hud-config.json` gains `statusLine.preset`, `[tui]` gains the `# omx:managed-status-line` marker comment) and needs your sign-off before merge. Totally fine to leave it sitting until you're ready — feel free to close again if you'd rather have the contract conversation first.

## Summary

Reuse the existing `HudPreset` vocabulary (`minimal | focused | full`) to drive the OMX-managed `[tui].status_line` in `~/.codex/config.toml`. Today that array is hard-coded to one seven-field default — this PR makes it preset-driven, fully backward compatible, and configurable through the same `.omx/hud-config.json` users already know.

Companion to #1940 / #1942 (preservation of user-customized `status_line`). Layered on top of `dev`, not `main`.

## Design

| Preset | Fields | When to use |
|---|---|---|
| `minimal` | `model-with-reasoning`, `git-branch` | Narrow terminals, distraction-free work |
| `focused` (default) | `model-with-reasoning`, `git-branch`, `context-remaining`, `total-input-tokens`, `total-output-tokens`, `five-hour-limit`, `weekly-limit` | **Byte-identical to today's hard-coded default** |
| `full` | same as `focused` today; reserved for future expansion | Power users; will diverge as Codex CLI adds more `status_line` keys |

Default is `focused`, so installs that do not specify a preset render exactly the same `status_line` as before this PR. Independent from the HUD's own `preset` key — a user can have `preset: "minimal"` for the HUD and `statusLine.preset: "full"` for the codex statusline (or vice versa).

## What's new in this re-open vs. #1951

Two commits, second one addressing the review:

1. `feat(config): preset-driven [tui] status_line` — original feature.
2. `fix(config): tighten OMX status_line detection` — review fixes:
   - **Owner CI blocker.** Drop the unused `OMX_TUI_STATUS_LINE` alias that tripped `tsconfig.no-unused` / Typecheck (Node 20|22). `npm run check:no-unused` is clean locally.
[...]

PR #1958, opened 2026-04-27
Fix forced setup status line preset sync

Yeachan-Heo:
## Summary
- Fixes the #1953 follow-up regression where `omx setup --scope project --force` could read a stale `.omx/hud-config.json` `statusLine.preset` before overwriting HUD config.
- Uses the forced HUD default preset when generating config during forced setup and regenerates `[tui].status_line` instead of preserving the stale value.
- Adds regression coverage for the minimal-to-focused force-sync repro.

## Verification
- `npm run build`
- `npm run check:no-unused`
- `node --test dist/cli/__tests__/setup-refresh.test.js dist/config/__tests__/generator-status-line-presets.test.js`
- Manual repro with `node dist/cli/omx.js setup --scope project --force` confirmed HUD config and `status_line` both focused/default.

Fixes the #1953 follow-up regression.

### DIFF
--- a/src/cli/__tests__/setup-refresh.test.ts
+++ b/src/cli/__tests__/setup-refresh.test.ts
@@ -404,6 +404,52 @@ describe("omx setup refresh summary and dry-run behavior", () => {
     }
   });
 
+  it("keeps forced HUD config overwrite and generated status_line preset in sync", async () => {
+    const wd = await mkdtemp(join(tmpdir(), "omx-setup-refresh-"));
+    try {
+      await mkdir(join(wd, ".omx", "state"), { recursive: true });
+      await mkdir(join(wd, ".codex"), { recursive: true });
+      await writeFile(
+        join(wd, ".omx", "hud-config.json"),
+        JSON.stringify({
+          preset: "focused",
+          statusLine: { preset: "minimal" },
+        }),
+      );
+      await writeFile(
+        join(wd, ".codex", "config.toml"),
+        [
+          "[tui]",
+          "# omx:managed-status-line",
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
OMX review found a blocker. I can’t submit this as a formal “request changes” review because GitHub blocks requesting changes on our own PR, so recording it as a blocking review comment.

## Blocker: `omx setup --force` now clobbers user-owned `[tui].status_line`

Relevant areas:

- `src/cli/setup.ts`
- `src/config/generator.ts`

The intended fix is good: forced setup should not read stale pre-overwrite HUD config when generating managed `[tui].status_line`.

However, the current implementation broadens that behavior too far:

- `src/cli/setup.ts` passes `forceStatusLinePreset: force` into config generation.
- `src/config/generator.ts` then stops preserving existing `[tui].status_line` when `force` is true.
- It also skips preserving customized `[tui]` sections extracted from OMX blocks when `force` is true.

Manual repro:

```toml
[tui]
theme = "night"
status_line = ["git-branch"]
```

After:

```bash
node dist/cli/omx.js setup --scope=project --force
```

Result: the custom user-owned status line is replaced with the managed focused preset.

Expected fix: force-sync only OMX-managed status lines / stale managed HUD output, not arbitrary user-owned `status_line` values. If the intended product contract is that `--force` resets user-owned status lines too, that needs explicit owner confirmation plus docs/tests.

Checks passed in review:

- `npm run build`
- `npm run check:no-unused`
- `node --test dist/cli/__tests__/setup-refresh.test.js`
- `node --test dist/config/__tests__/generator-status-line-presets.test.js`

GitHub CI is green; manual repro confirmed the blocker.

—

PR #1972, opened 2026-04-27
Fix PostToolUse false positives from Bash output text

NagyVikt:
## Summary
- Gate Bash hard-failure guidance on stderr or non-zero exit output instead of stdout-only diagnostic strings.
- Keep MCP transport-death handling scoped to MCP-like tool calls, so Bash source/log reads that mention MCP transport text stay silent.
- Refresh native hook docs and add regressions for source/log text false positives.

## Verification
- npm run build
- npm run lint
- npm run check:no-unused
- git diff --check


Yeachan-Heo:
Thanks for the PR. This repo takes PRs against `dev`, not `main`, and I am closing this main-base branch rather than retargeting it in place. If this is a real reproducible bug, please reopen as a focused PR against `dev` with the same narrow scope and verification.

—

PR #2002, opened 2026-04-29
Isolate tmux nudges by instance ownership

deanpress:
Fixes the tmux hook isolation issue where OMX nudges leak into unrelated sessions.

**Root cause:** `.omx/tmux-hook.json` uses static pane IDs (e.g. `%0`) and lacks ownership verification.

**Changes:**
1. **Session Tagging:** Tagged tmux sessions with `@omx_instance_id` at spawn.
2. **Hook Verification:** Added check in notify/hook logic to abort injection if `@omx_instance_id` doesn't match.
3. **Dynamic Targeting:** Prefer session-tagged targeting over static pane IDs.

Prevents cross-talk between concurrent OMX instances.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -881,12 +881,21 @@ describe("detached tmux new-session sequencing", () => {
       "sess-detached-managed",
     );
     const newSession = steps.find((step) => step.name === "new-session");
+    const tagSession = steps.find((step) => step.name === "tag-session");
     assert.ok(newSession);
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
Thanks for the report and PR. This repository takes PRs against `dev`, not `main`, and this branch is currently targeting `main`, so I’m closing this PR rather than merging it as-is.

The issue area is valid enough to keep on our radar, but any implementation needs to be re-opened against `dev` with the current hook/tmux contracts verified there.

—

PR #2037, opened 2026-04-30
fix(plugin): detect mirror CLI calls in spaced paths

lkraider:
## Summary

`sync-plugin-mirror` currently skips its top-level CLI path when the repo path
contains spaces. `import.meta.url` percent-encodes the script path while
`process.argv[1]` remains unescaped, so the existing direct-invocation check
fails and the mirror CLI does not run from those checkouts.

## Changes

- normalize the direct-invocation check to compare filesystem paths instead of
  raw file URLs
- export the helper so the launch contract can be covered directly in tests
- add a regression test that exercises a space-containing script path

## Validation

- [x] `npm run build`
- [ ] `npm test`
- [x] `node --test dist/catalog/__tests__/plugin-bundle-ssot.test.js`
- [x] `node dist/scripts/sync-plugin-mirror.js --check`

## Checklist

## Related

Closes #

### DIFF
--- a/src/catalog/__tests__/plugin-bundle-ssot.test.ts
+++ b/src/catalog/__tests__/plugin-bundle-ssot.test.ts
@@ -3,9 +3,13 @@ import assert from "node:assert/strict";
 import { cp, mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { pathToFileURL } from "node:url";
 import { getSetupInstallableSkillNames } from "../installable.js";
 import { readCatalogManifest } from "../reader.js";
-import { syncPluginMirror } from "../../scripts/sync-plugin-mirror.js";
+import {
+	isDirectCliInvocation,
+	syncPluginMirror,
+} from "../../scripts/sync-plugin-mirror.js";
 
 const root = process.cwd();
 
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
APPROVE

Read-only OMX review of the exact PR diff vs `dev` (`703a5b6a`): the change is narrow to `src/scripts/sync-plugin-mirror.ts` and `src/catalog/__tests__/plugin-bundle-ssot.test.ts`.

Verified:
- The reported bug is real: `import.meta.url` percent-encodes a space-containing script path while `process.argv[1]` remains an unescaped filesystem path, so the old raw `file://` comparison fails.
- The fix compares normalized filesystem paths (`fileURLToPath(importMetaUrl)` vs `resolve(argvPath)`), which preserves imported-module behavior, handles space-containing paths, handles missing `argvPath`, and is cross-platform enough for the Node file URL/path forms used here.
- No plugin mirror sync/check contract regression found; mirror contents and metadata policy are untouched.
- Regression coverage directly exercises a space-containing script path and the non-direct cases.

Local validation after `npm ci`:
- `npm run build`
- `node --test dist/catalog/__tests__/plugin-bundle-ssot.test.js`
- `node dist/scripts/sync-plugin-mirror.js --check`
- Additional manual spaced-path invocation via `/tmp/Manual Library.../repo/dist/scripts/sync-plugin-mirror.js --check`

CI at review time: required jobs are still running (`mergeStateStatus=UNSTABLE`), with completed lint/typecheck/build and several tests successful; remaining in-progress jobs should finish before merge.

Signature: OMX read-only PR review (Codex)

PR #2041, opened 2026-04-30
Fix prompt worker startup death handling

Yeachan-Heo:
## Summary\n- Treat prompt-launched workers that die before startup dispatch as dead workers instead of startup notify failures\n- Record failed worker status and a worker_stopped event so team await/status can surface the dead-worker path\n- Tighten the prompt dead-worker smoke test to exercise immediate worker exit\n\n## Verification\n- npm run build\n- node --test --test-name-pattern "returns a dead-worker event for the prompt-launch smoke path instead of timing out" dist/cli/__tests__/team.test.js\n- node --test dist/cli/__tests__/team.test.js\n- npm run check:no-unused\n- npm run lint\n- npx c8 --all --src dist --exclude '**/__tests__/**' --exclude 'dist/bin/**' --exclude 'dist/**/*.d.ts' --reporter=text-summary --report-dir coverage/ts-full-target node --test --test-name-pattern "returns a dead-worker event for the prompt-launch smoke path instead of timing out" dist/cli/__tests__/team.test.js

PR #2042, opened 2026-04-30
fix(team): treat ready prompts as startup evidence

lkraider:
## Summary

Interactive Codex workers on current `dev` can miss the startup-direct path
while their pane is still bootstrapping, later reach a real ready prompt, and
still be recorded as `codex_startup_no_evidence_after_fallback` after a
notified startup dispatch. This keeps the startup-evidence path aligned with
what tmux already observed and avoids a false recoverable startup issue.

## Changes

- treat a successful interactive ready-wait prompt as synthetic
  `ready_prompt` startup evidence when startup dispatch receipts are only
  `notified` or when direct fallback confirmation would otherwise wait for
  startup evidence again
- add a runtime regression that forces the first startup capture to see a
  bootstrapping pane, then verifies a later ready prompt keeps the worker
  neutral and the startup dispatch request `notified`

## Validation

- [x] `npm run build`
- [ ] `npm test`
- [x] `CODEX_HOME=<empty temp dir> node --test dist/team/__tests__/runtime.test.js`
  (`113/114` passing; unchanged unrelated sandbox failure:
  `shutdownTeam reaps detached prompt-worker descendants`)

## Checklist

## Related

Closes #

### DIFF
--- a/src/team/__tests__/runtime.test.ts
+++ b/src/team/__tests__/runtime.test.ts
@@ -1914,6 +1914,142 @@ esac
     }
   });
 
+  it('startTeam treats a confirmed ready prompt as startup evidence after hook notification', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-runtime-ready-prompt-evidence-'));
+    const previousTmux = process.env.TMUX;
+    const previousTmuxPane = process.env.TMUX_PANE;
+    const previousLaunchMode = process.env.OMX_TEAM_WORKER_LAUNCH_MODE;
+    const previousWorkerCli = process.env.OMX_TEAM_WORKER_CLI;
+    const previousReadyTimeout = process.env.OMX_TEAM_READY_TIMEOUT_MS;
+    const previousStartupEvidenceTimeout = process.env.OMX_TEAM_STARTUP_EVIDENCE_TIMEOUT_MS;
+    const previousStartupDispatchRetries = process.env.OMX_TEAM_STARTUP_DISPATCH_RETRIES;
+    let receiptNotifier: NodeJS.Timeout | null = null;
+    let runtimeTeamName: string | null = null;
+
+    try {
+      await withMockTmuxFixture(
+        {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
APPROVE

Read-only OMX review of external PR #2042 at head `453619e5` against `dev` (`edece3eb`). I treated the PR input as adversarial and inspected the exact diff plus the nearby startup/dispatch paths.

What I verified:
- The ready-prompt evidence is narrow: `ready_prompt` is not persisted by `readWorkerStartupEvidence`; it is only synthesized from the existing interactive ready-wait path and only passed into the startup dispatch/evidence junctions.
- It does not mask worker startup failures: the flag is false for prompt workers, initial-prompt workers, skipped ready waits, and paths where the interactive ready wait times out or the pane is dead; those paths still record/return their existing failure or recoverable-startup outcomes.
- Hook `notified` receipts and direct fallback confirmations now settle only after the current tmux readiness wait has already observed an interactive prompt, which matches the false-negative case being fixed without broadening worker state evidence semantics.
- Overlap with internal PR #2041 is safe: #2041 handles prompt-mode worker death reporting, while #2042 is limited to interactive ready-prompt startup evidence. The touched `startTeam` area is adjacent but the launch-mode gates are disjoint, and `dispatchCriticalInboxInstruction` still returns early for prompt mode before the new ready-prompt logic.
- CI state at review time is CLEAN with all visible checks successful.

Local validation after installing missing dependencies with `npm ci`:
- `npm run build`
- `CODEX_HOME=$(mktemp -d) node --test --test-name-pattern "startTeam treats a confirmed ready prompt as startup evidence after hook notification|startTeam starts worker-2 readiness before delayed worker-1 readiness settles|waitForWorkerStartupEvidence|worker_notify_failed|startup_no_evidence" dist/team/__tests__/runtime.test.js` — 4/4 passing

No blockers found. No merge performed.

Signature: OMX read-only PR review (Codex)

PR #2044, opened 2026-04-30
Keep OMX MCP serve process alive

pgagarinov:
## Summary
- keep `omx mcp-serve <target>` pending after the selected MCP server module loads so stdio MCP clients do not lose the server during startup
- add focused regression coverage for the old immediate-settle path
- add a test-only `keepProcessAlive: false` opt-out so dispatch/validation unit tests can still await the command

## Problem

`mcpServeCommand()` returned as soon as the target loader finished. In the bin-wrapper path that can let the CLI command complete and close the stdio MCP process before the client finishes initialization.

## Root cause

The MCP server modules start their stdio lifecycle as a top-level import side effect, but the CLI command did not keep ownership of the process after `await loaders[target]()`.

## Changes

- after loading the target, keep `mcpServeCommand()` pending unless tests explicitly disable keepalive
- update existing tests to use the opt-out for dispatch/validation paths
- add a red-green regression asserting the loader ran while the command promise remains unsettled after one event-loop tick

## Verification
- `npm run build`
- `node --test dist/cli/__tests__/mcp-serve.test.js`
- `node --test dist/cli/__tests__/package-bin-contract.test.js`

## Risk / compatibility
- runtime behavior changes only for `omx mcp-serve <target>`: the command now stays alive for the stdio MCP lifecycle
- validation errors and help output still return normally before keepalive is entered
- the opt-out is only exposed for tests

### DIFF
--- a/src/cli/__tests__/mcp-serve.test.ts
+++ b/src/cli/__tests__/mcp-serve.test.ts
@@ -10,6 +10,7 @@ describe("mcp-serve command", () => {
 
 		await mcpServeCommand(["trace"], {
 			env,
+			keepProcessAlive: false,
 			loaders: {
 				"state-server.js": async () => {
 					loaded.push("state-server.js");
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
APPROVE

Reviewed external PR #2044 adversarially against dev at a1556de8; PR head inspected: a2593b08.

Scope reviewed:
- src/cli/mcp-serve.ts
- src/cli/__tests__/mcp-serve.test.ts
- related mcp lifecycle/CLI dispatch context

Findings:
- Keeping `omx mcp-serve <target>` pending after the selected server module loads is correct for the stdio MCP server lifecycle; the server bootstrap owns shutdown on stdin/transport/parent lifecycle.
- The `keepProcessAlive: false` opt-out is narrow/test-only in the changed unit-test call sites; production CLI dispatch does not pass it.
- Module-load failures are not masked because `await loaders[target]()` must complete successfully before the pending keepalive promise is reached.
- I did not find evidence this can hang unrelated non-test command paths; the production caller is the `mcp-serve` dispatch branch, while help/validation paths return/throw before keepalive.

Targeted checks run in an isolated /tmp copy:
- npm ci --ignore-scripts --no-audit --no-fund
- npm run build
- node --test dist/cli/__tests__/mcp-serve.test.js
- node --test dist/cli/__tests__/package-bin-contract.test.js

All targeted checks passed.

CI note: CI was still running/pending in the provided review context; I did not wait for full CI completion and did not merge.

Required signature: OMX-Codex-Review: APPROVE pr-2044 a2593b08

PR #2058, opened 2026-05-02
Fix native hook CLI detection for symlinked installs

HaD0Yun:
## Summary

Fixes native hook CLI main-module detection when `oh-my-codex` is invoked through an npm-global package symlink.

Node resolves `import.meta.url` to the real file path, while `process.argv[1]` can remain the symlinked package path. Before this change, that mismatch made `codex-native-hook.js` skip its CLI entrypoint, so inactive Stop hook payloads could exit successfully with empty stdout instead of the expected `{}` JSON response.

## Changes

- Normalize `argv[1]` with `realpathSync()` before comparing it with `import.meta.url`.
- Keep the previous path comparison as a fallback when realpath resolution fails.
- Add a regression test for symlinked native hook argv entry paths.

## Validation

- [x] `npm run lint`
- [x] `npm run build`
- [ ] `npm test` — not run; targeted native-hook regression was run instead.
- [ ] `omx doctor` (when setup/config behavior changes) — not applicable; this does not change setup/config behavior.

Additional validation:

- [x] `node --test --test-name-pattern 'main module|Stop CLI runs' dist/scripts/__tests__/codex-native-hook.test.js`
- [x] Stop payload smoke through npm-global symlink path:
  - `OMX_HOOK_PLUGINS=0 node /home/yun/.npm-global/lib/node_modules/oh-my-codex/dist/scripts/codex-native-hook.js`
  - output: `{}`
- [x] Stop payload smoke through realpath:
  - `OMX_HOOK_PLUGINS=0 node /home/yun/.omx/dev-oh-my-codex/dist/scripts/codex-native-hook.js`
  - output: `{}`

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — not needed; internal CLI entrypoint detection fix with regression coverage

## Related

Closes #

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert/strict";
 import { execFileSync } from "node:child_process";
-import { existsSync } from "node:fs";
-import { chmod, mkdir, mkdtemp, readFile, readdir, rm, writeFile } from "node:fs/promises";
+import { existsSync, realpathSync } from "node:fs";
+import { chmod, mkdir, mkdtemp, readFile, readdir, rm, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { pathToFileURL } from "node:url";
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
Blocking issue found:

- `src/scripts/codex-native-hook.ts` now compares only `moduleUrl` vs `realpathSync(argv1)`. This fixes Node’s default symlink behavior, but regresses symlinked package-root invocations when Node preserves the main-module symlink, e.g. via `node --preserve-symlinks-main` / `NODE_OPTIONS`. In that case `import.meta.url` remains the symlink URL, `realpathSync(argv1)` becomes the real path, and the CLI main guard returns false, so the native hook emits no JSON.
  - Repro smoke from review: default symlinked package-root invocation emitted `{}`, but `node --preserve-symlinks-main <symlink-root>/dist/scripts/codex-native-hook.js` emitted no stdout.
  - Suggested fix: accept both forms:

```ts
const argvUrl = pathToFileURL(argv1).href;
if (moduleUrl === argvUrl) return true;
try {
  return moduleUrl === pathToFileURL(realpathSync(argv1)).href;
} catch {
  return false;
}
```

Please add a regression for the symlink URL / `--preserve-symlinks-main` case.

Verification run:
- `npm ci` ✅
- `npm run build` ✅
- `node --test --test-name-pattern 'main module|Stop CLI runs' dist/scripts/__tests__/codex-native-hook.test.js` ✅
- `npm run lint` ✅
- GitHub CI at review time: most checks passing, final coverage/build jobs still pending.

—

PR #2061, opened 2026-05-02
Fix state-level Stop hook replay loops

HaD0Yun:
## Summary
- suppresses repeated state-level Stop blocks for unchanged team/team-worker runtime state across fresh Stop turns
- keeps the first Stop block for active non-terminal work, and still re-blocks when the actual task/team state changes
- locks the team-worker regression in `codex-native-hook.test.ts`

## Problem
Deus evidence showed that Stop hook `failed`-looking behavior was usually not a native hook process crash. The hook returned valid `decision: "block"` JSON, but the same non-terminal team state kept re-entering the Stop hook on fresh turns. That made unchanged `in_progress` / `pending` team-worker state look like a repeated Stop-hook failure loop and could burn tokens.

## Root cause
`buildRepeatableStopSignature(...)` included volatile per-turn fields (`turn_id`, transcript path, and last assistant message) for all Stop block kinds. That is correct for content-sensitive auto-nudge/document-refresh decisions, but wrong for state-level team decisions. A team-worker block is determined by stable runtime state (`team-worker-stop` + session/thread + task/status detail), so fresh turns with the same unresolved state should not produce another block.

## What changed
- Added `isStateLevelStopSignatureKind(...)` for state-backed Stop block families.
- For `team-worker-stop` and `team-stop`, Stop replay signatures now ignore turn/transcript/message churn and key only on stable session/thread/runtime-detail state.
- Updated the existing team-worker Stop replay test to prove:
  - first non-terminal task state blocks;
  - exact replay suppresses;
  - fresh Stop turn with unchanged task state also suppresses;
  - changed task state (`in_progress` -> `blocked`) re-blocks.

## Why this design
State-level Stop decisions should repeat only when state changes. Content-level Stop decisions can still use volatile message/turn fields because their input is the assistant message itself.

## Overlap assessment
**Follow-up.** Prior PR #1348 addressed native Stop repeat-fire behavior broadly. This PR narrows the remaining team/team-worker state-level replay loop that still reproduced in recent Deus evidence.

## Validation
- `npm run build`
- `node --test --test-name-pattern='suppresses identical team worker Stop replays across fresh turns until task state changes' dist/scripts/__tests__/codex-native-hook.test.js`

Note: the clean worktree did not have its own `node_modules`; I used the existing local workspace install as a temporary symlink for verification and removed it before commit.

## Risks / compatibility
- Active team work still blocks once.
- A changed task/team state still re-blocks.
- Non-state Stop signatures retain the previous volatile signature behavior.

## Changed files

## Target-base diff classification
Bug fix with real implementation delta.

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -4710,12 +4710,15 @@ exit 0
       process.env.OMX_TEAM_STATE_ROOT = stateDir;
       process.env.OMX_TEAM_LEADER_CWD = cwd;
 
-      const result = await dispatchCodexNativeHook(
-        {
-          hook_event_name: "Stop",
-          cwd: workerCwd,
-          session_id: "sess-stop-team-worker-stale",
-        },
+      const payload = {
+        hook_event_name: "Stop",
+        cwd: workerCwd,
+        session_id: "sess-stop-team-worker-stale",
+        thread_id: "thread-stop-team-worker-stale",
+      };
+      const result = await dispatchCodexNativeHook(payload, { cwd: workerCwd });
+      const replay = await dispatchCodexNativeHook(
+        { ...payload, stop_hook_active: true },
         { cwd: workerCwd },
       );
 
@@ -4723,6 +4726,7 @@ exit 0
         (result.outputJson as { stopReason?: string } | null)?.stopReason,
         "team_worker_worker-1_1_in_progress",
       );
+      assert.equal(replay.outputJson, null);
     } finally {
       if (typeof prevTeamWorker === "string") process.env.OMX_TEAM_WORKER = prevTeamWorker;
       else delete process.env.OMX_TEAM_WORKER;
@@ -4734,7 +4738,7 @@ exit 0
     }
   });
 
-  it("suppresses identical team worker Stop replays but re-blocks fresh turns and task state changes", async () => {
+  it("re-blocks live team worker Stop replays but suppresses stale terminal worker repeats", async () => {
     const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-stop-team-worker-repeat-"));
     try {
       await initTeamState(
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocking finding:

- `src/scripts/codex-native-hook.ts:1627` / `src/scripts/codex-native-hook.ts:2007` changes `team-worker-stop` signatures to `kind|session|thread|stopReason` while still using `allowRepeatDuringStopHook: false`. Because the detail is only `team_worker_<worker>_<task>_<status>`, a second Stop with `stop_hook_active: true` and the task still `in_progress` is treated as a duplicate and returns no block. The updated test now codifies that behavior at `src/scripts/__tests__/codex-native-hook.test.ts:4198` and `:4218`. That breaks the team-worker Stop contract for a live non-terminal task: a worker can stop after one continuation without completing/reporting a blocker, as long as the status string did not change. This addresses stale-state replay loops by suppressing all same-status live state too, so it creates a false negative for active workers. Please distinguish stale/replay state from live non-terminal state (for example by using explicit stale/age/heartbeat/run-state evidence) rather than suppressing every same `stopReason` across fresh Stop attempts.

Verification performed:
- `gh pr checks 2061 --watch=false`: all required checks reported pass; PR Base Guidance skipped.
- `git diff --check origin/dev...HEAD`: pass.
- Focused `origin/dev...HEAD` diff inspection: only `src/scripts/codex-native-hook.ts` and its native-hook tests changed.
- `npx tsc --noEmit`: inconclusive in this local checkout because dependencies such as `@iarna/toml` and `@modelcontextprotocol/sdk` are not installed; CI typecheck is passing.
—

Yeachan-Heo (approved):
APPROVE

Final merge-gate review evidence:
- Scope verified from `gh pr diff 2061 --name-only`: only `src/scripts/codex-native-hook.ts` and `src/scripts/__tests__/codex-native-hook.test.ts` are touched.
- CI verified with `gh pr checks 2061 --watch=false`: all required checks are passing; PR Base Guidance is skipped as expected.
- Mergeability verified with `gh pr view --json mergeable,mergeStateStatus`: `MERGEABLE` / `CLEAN`, base `dev`, head `46fb70cc896472001429868dd7aec4172c334cf2`.
- Previous blocker is fixed: live non-terminal team-worker Stop replays now set `allowRepeatDuringStopHook: !workerRunStateIsTerminal`, while terminal/missing-state repeats remain non-repeatable; state-level signatures are stable on `kind|session|thread|detail`.
- Local verification in the PR worktree passed: `git diff --check origin/dev...HEAD`; `npm run build`; focused native-hook regression `node --test --test-name-pattern='re-blocks live team worker Stop replays but suppresses stale terminal worker repeats|suppresses identical team worker Stop replays across fresh turns until task state changes' dist/scripts/__tests__/codex-native-hook.test.js` (1 pass, 0 fail).

—

PR #2064, opened 2026-05-02
Fix stale team-worker Stop task blocks

HaD0Yun:
## Summary
- gates team-worker Stop blocking on active worker states (`working` / `blocked`) instead of stale task assignment alone
- preserves Stop protection for workers actively working on non-terminal tasks
- adds regression coverage for terminal worker status with stale `current_task_id` / `assigned_tasks`

## Problem
Recent Stop-hook evidence showed `decision: "block"` responses repeatedly surfacing as failed Stop hooks when team workers retained stale assigned task ids. A worker could already be `done` / inactive while the task file still said `in_progress`, so the Stop hook kept treating stale assignment metadata as active work.

## Root cause
`buildTeamWorkerStopOutput(...)` built candidate task ids from `status.current_task_id` and `identity.assigned_tasks`, then blocked on any referenced non-terminal task. It did not first check whether the worker status itself was still active. That let stale worker/task synchronization produce an indefinite team-worker Stop block.

## What changed
- Added an explicit team-worker Stop active-state allowlist: `working`, `blocked`.
- `buildTeamWorkerStopOutput(...)` now returns no worker-specific Stop block when the worker status is terminal/inactive (`done`, `failed`, `draining`, `idle`, unknown/missing), even if stale task references remain.
- Updated existing active-task tests from `idle` to `working` so they continue proving real active work is blocked.
- Added a regression test for `state: "done"` + stale non-terminal task reference.

## Why this design
The Stop hook should block when the worker is actively accountable for non-terminal work, not when persisted assignment metadata is stale. Team-level Stop gating still remains available after worker-specific gating, so the leader/team lifecycle is not silently bypassed.

## Overlap assessment
**Follow-up.** PR #2061 addresses state-level Stop replay signature loops, and PR #1884 addressed identical team-worker Stop replay suppression. This PR covers the adjacent remaining stale worker-state synchronization case: stale assigned task ids should not create a worker-specific non-terminal-task Stop block once the worker status is terminal/inactive.

## Validation
- `npm run build`
- `node --test --test-name-pattern='team worker' dist/scripts/__tests__/codex-native-hook.test.js`
- `node --test --test-name-pattern='already terminal' dist/scripts/__tests__/codex-native-hook.test.js`
- `npx tsc --noEmit`
- `npm run check:no-unused`

## Risks / compatibility
- Active workers in `working` or `blocked` state still block on non-terminal tasks.
- Inactive/terminal worker Stop no longer emits the misleading worker-specific non-terminal-task block; active team-level Stop gating can still block if the team phase remains non-terminal.

## Changed files

## Target-base diff classification
Bug fix with real implementation delta.

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -4087,7 +4087,7 @@ esac
         team_state_root: join(cwd, ".omx", "state"),
       });
       await writeJson(join(workerDir, "status.json"), {
-        state: "idle",
+        state: "working",
         current_task_id: "1",
         updated_at: new Date().toISOString(),
       });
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
APPROVE

Merge-gate review summary:
- Verified scope with `origin/dev...HEAD`: only `src/scripts/codex-native-hook.ts` and `src/scripts/__tests__/codex-native-hook.test.ts` changed.
- `git diff --check origin/dev...HEAD`: pass.
- `gh pr checks 2064 --repo Yeachan-Heo/oh-my-codex --watch=false`: all required CI checks pass; PR Base Guidance is skipped.
- Focused diff review confirms the Stop hook now gates worker-specific task blocking on active worker states (`working` / `blocked`) before considering stale `current_task_id` / `assigned_tasks` task references.
- The PR preserves live non-terminal worker blocking: the active-task and replay tests now use `state: "working"`, first/fresh Stop attempts still block on `in_progress`, and a task status transition to `blocked` re-blocks.
- The new regression covers the stale case from the prior blocker: `state: "done"` with stale non-terminal task metadata no longer emits the worker-specific task block and falls back to team-level gating instead of falsely trapping the worker.

I do not see a merge-blocking false-positive or false-negative in this diff.
—

PR #2081, opened 2026-05-03
Preserve detached tmux sessions across SSH attach disconnects

pgagarinov:
Related: #2083 (closed; prior context, not an auto-close link)

## Summary
- preserve detached OMX tmux sessions when `attach-session` fails but `tmux has-session` confirms the session is still alive
- ignore external `SIGHUP` in the detached leader shell so SSH/client disconnects do not kill the Codex child
- keep rollback/direct fallback behavior when the detached session is actually gone, and keep the WSL attach-no-op fallback path distinct
- document the SSH/tmux reattach behavior in the README

## Validation
- [x] `npm run build`
- [x] `node --test dist/cli/__tests__/index.test.js dist/cli/__tests__/launch-fallback.test.js`
- [x] `npm run lint`
- [x] `git diff --check`

## Not-tested
- Full `npm test` completion in this attached OMX/tmux session. It was attempted, but local runtime environment interfered with unrelated tests: `question.test.js` opened a real tmux question UI, and inherited `OMX_TEAM_WORKER_LAUNCH_ARGS` changed expected team-worker reasoning defaults. The affected team assertions passed when that env override was removed.

## Notes
This is intentionally narrow launch lifecycle hardening: attach loss no longer implies session failure when tmux still reports the detached session exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,9 @@ omx --madmax --high
 
 On macOS/Linux interactive terminals with `tmux` available, this starts the
 leader in OMX-managed detached tmux by default so the HUD/runtime panes can be
-created and recovered.
+created and recovered. If an SSH client or tmux attach command disconnects after
+startup while the managed session still exists, OMX leaves that detached session
+running; use `tmux ls` and `tmux attach-session -t <session>` to reattach.
 
 If you want a one-off launch with no OMX tmux/HUD management, use `--direct`:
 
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
Thanks for the PR. I’m going to close this rather than take it as-is.

The change touches launch/attach lifecycle semantics and shell signal handling, which is a high-risk workflow contract area for OMX. Without a maintainer-reviewed issue and a narrower reproduction/contract, we’re not going to merge this direction from a drive-by PR.

—

PR #2235, opened 2026-05-09
fix(planning): ignore hidden approved launch hints

lkraider:
## Summary

Approved handoff selection currently scans the full approved PRD markdown body. When a plan includes fenced shell examples or indented command snippets, those hidden commands can participate in Team or Ralph launch-hint selection, which makes same-task lookups ambiguous and can make hidden-only hints look valid. This keeps launch-hint selection limited to visible markdown prose while preserving the existing approved handoff contracts.

## Changes

- add a planning-local markdown visibility helper for normal, fenced, and indented scan states
- make approved Team and Ralph launch-hint collection read only visible markdown lines
- reuse the same hidden-markdown rule for Context Pack Outcome parsing and expand the planning proof for hidden-only hints, hidden duplicate hints, indented outcome declarations, and fence/tab edge cases

## Validation

- [x] `npm run build`
- [ ] `npm test`
- [ ] `omx doctor` (when setup/config behavior changes)
- `node --test dist/planning/__tests__/markdown-structure.test.js dist/planning/__tests__/artifacts.test.js dist/planning/__tests__/context-pack-status.test.js dist/cli/__tests__/ralph.test.js dist/cli/__tests__/team.test.js`
- `npx tsc --noEmit`
- `npm run check:no-unused`
- `node dist/scripts/run-test-files.js dist/cli/__tests__ dist/agents/__tests__ dist/autoresearch/__tests__ dist/catalog/__tests__ dist/compat/__tests__ dist/config/__tests__ dist/modes/__tests__ dist/pipeline/__tests__ dist/planning/__tests__ dist/scripts/__tests__ dist/session-history/__tests__ dist/subagents/__tests__ dist/utils/__tests__ dist/visual/__tests__`
- `node dist/scripts/generate-catalog-docs.js --check`

## Checklist

- [x] PR is focused and avoids unrelated changes
- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed
- [x] Backward-compatibility impact considered

## Related

- None

PR #2236, opened 2026-05-09
fix(planning): reuse same-lineage approved hints

lkraider:
## Summary

Approved launch-hint selection currently stops at the newest matching PRD. When that newest Team or Ralph handoff is nonready, short follow-ups stay pinned to the nonready handoff even when an older reusable handoff in the same lineage still exists. This change keeps the selector fail-closed while allowing it to reuse older same-lineage handoffs instead of dropping usable approved context.

## Changes

- backscan approved Team and Ralph launch hints across older PRDs only when the newest matching handoff is nonready
- keep Team fallback constrained to the full launch signature and preserve exact-command selection
- add planning and CLI regression coverage for same-lineage reuse, signature-mismatch non-fallback, and ambiguous same-lineage fail-closed behavior

## Validation

- [x] `npm run build`
- [ ] `npm test`
- Focused rerun: `node --test dist/planning/__tests__/artifacts.test.js dist/cli/__tests__/ralph.test.js dist/cli/__tests__/team.test.js dist/team/__tests__/approved-execution.test.js`

## Checklist

## Related

Closes #

### DIFF
--- a/src/cli/__tests__/ralph.test.ts
+++ b/src/cli/__tests__/ralph.test.ts
@@ -127,6 +127,38 @@ describe('readMatchedApprovedRalphExecutionHint', () => {
       await rm(cwd, { recursive: true, force: true });
     }
   });
+
+  it('reuses the older approved Ralph hint when the latest same-task handoff is missing its baseline', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-ralph-lineage-fallback-'));
+    const approvedTask = 'Repair approved Ralph lineage handoff';
+    try {
+      await mkdir(join(cwd, '.omx', 'plans'), { recursive: true });
+      await writeFile(
+        join(cwd, '.omx', 'plans', 'prd-alpha-lineage.md'),
+        `# PRD\n\nLaunch via omx ralph ${JSON.stringify(approvedTask)}\n`,
+      );
+      await writeFile(join(cwd, '.omx', 'plans', 'test-spec-alpha-lineage.md'), '# Test Spec\n');
+      await writeFile(
+        join(cwd, '.omx', 'plans', 'prd-zeta-lineage.md'),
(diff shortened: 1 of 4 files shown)


Yeachan-Heo (changes requested):
Thanks for the focused lineage fallback work. The pre-merge OMX review found a blocker that needs fixing before this can merge.

Blocker:
- `src/planning/artifacts.ts` still scans the full PRD markdown body with `content.matchAll(...)` for launch-hint matching.
- That includes fenced/indented code blocks. This is the same hidden-hint exposure class that blocked #2235.
- #2236 amplifies the risk because the new older-PRD backscan can select an older hidden-only Team/Ralph hint as a reusable fallback when the newest visible same-lineage handoff is nonready.

Repro evidence from the review:
- latest PRD has visible `omx team 3:executor "Execute hidden older handoff"` but no matching test spec, so it is nonready / missing-baseline
- older PRD has the same command only inside a fenced code block plus a matching test spec
- `readApprovedExecutionLaunchHintOutcome(cwd, "team")` resolves the older fenced command as plan-only

That violates the requested “no hidden-hint regression” gate. Please make launch-hint matching use visible-markdown-only scanning before enabling historical fallback, and add regression coverage for hidden older hints not being selected during fallback.

Also note: once the hidden scan blocker is fixed, this still changes launch-hint selection semantics from latest-plan-only to historical same-lineage selection, so owner confirmation is needed before merge.

—

Yeachan-Heo:
Closing this PR because the pre-merge OMX review found a merge-blocking hidden-launch-hint issue: the historical fallback still scans the full PRD markdown body, so an older fenced/hidden Team/Ralph launch hint can be selected as fallback when the newest visible same-lineage handoff is nonready.

A revised PR is welcome if launch-hint matching is switched to visible-markdown-only scanning before enabling historical fallback, with regression coverage proving hidden older hints are not selected.

—

PR #2253, opened 2026-05-10
fix(setup): prevent recursive notify dispatcher wrapping

ictechgy:
## Summary

- Prevent `omx setup` from preserving an OMX-managed notify dispatcher/hook as `previousNotify` across global installs
- Add a runtime dispatcher guard so managed OMX notify entrypoints are skipped if stale metadata already exists
- Add regression tests for cross-install notify recursion, preserving real user notify commands, and Windows-safe TOML seeding in setup tests

## Root cause

A legacy `notify-dispatch.json` could preserve another OMX dispatcher as `previousNotify` after reinstalling/upgrading from a different global package path. That allowed the dispatcher to invoke itself/another managed dispatcher recursively, causing runaway `node` processes.

## Base branch

This PR targets `dev` as requested after the earlier `main`-targeted PR was closed.

## Validation

- `npm run build`
- `node --test dist/config/__tests__/generator-notify.test.js dist/cli/__tests__/setup-install-mode.test.js dist/scripts/__tests__/notify-dispatcher.test.js`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

## Notes

A full `npm test` was attempted previously against the main-based branch, but unrelated existing environment/tmux-sensitive tests failed in the active OMX/tmux session. The changed notify/setup test targets above pass on this `dev`-based branch.

### DIFF
--- a/src/cli/__tests__/setup-install-mode.test.ts
+++ b/src/cli/__tests__/setup-install-mode.test.ts
@@ -165,6 +165,93 @@ describe("notify setup scope", () => {
 			await rm(wd, { recursive: true, force: true });
 		}
 	});
+
+	it("does not preserve a managed dispatcher as previous notify across global installs", async () => {
+		const wd = await mkdtemp(join(tmpdir(), "omx-recursive-notify-"));
+		try {
+			await withIsolatedUserHome(wd, async (codexHomeDir) => {
+				const metadataPath = join(
+					codexHomeDir,
+					".omx",
+					"notify-dispatch.json",
+				);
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
Thanks for retargeting to `dev`. We’re still not taking this external PR in its current form.

If this reproduces a real user-facing failure, please open a focused issue with exact reproduction steps and logs first; we’ll handle it from there.

—

ictechgy:
Got it. I'll go ahead and create an issue with the reproduction steps included.

PR #2255, opened 2026-05-10
Bound native Stop hook diff audit runtime

ShunmeiCho:
## Summary
- bound the Stop-only sloppy fallback diff audit with a shared time budget
- add per-git-command timeout, untracked file count cap, and max untracked file size cap
- document the new Stop diff audit budget knobs

## Why
A Codex App/native session repeatedly showed:

```
Stop hook (failed)
  error: hook timed out after 30s
```

Local diagnosis on oh-my-codex 0.16.3 showed the non-OMX cc-notch Stop hook completed in about 0.04s, while the OMX `codex-native-hook.js` Stop path took about 25.86s for a no-op Stop payload and then timed out in real foreground Stop handling. The affected workspace also resolved git to a very large parent worktree, so the Stop-only diff audit should be best-effort rather than unbounded.

## Testing
- `npm run build`
- `npm run check:no-unused`
- `node --test dist/scripts/__tests__/codex-native-hook.test.js` (278 tests passed)
- `npx biome lint src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts`

## Notes
The audit still blocks ordinary staged, unstaged, and untracked sloppy fallback edits when git responds within budget. If the budget is exhausted, it returns no findings instead of risking Codex's native Stop hook timeout.

### DIFF
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -55,6 +55,18 @@ Setup-owned trust state is limited to those generated wrapper identities; user h
 | wiki session capture | none | `session-end` | runtime-fallback | Wiki session-log capture runs from the existing runtime session-end cleanup path, not from a native Codex hook |
 | `session-idle` | none | `session-idle` | runtime-fallback | Still emitted from runtime/notify path, not native Codex hooks |
 
+## Stop diff audit budget
+
+The native `Stop` adapter also runs a bounded sloppy-fallback diff audit after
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the PR. This targets `main`, but this repo takes contribution PRs against `dev` only. Please reopen/retarget the same fix against `dev` if you want it reviewed.

—

PR #2284, opened 2026-05-11
hud: register client-resized hook to re-enforce pane height on terminal resize

grndlvl:
## Problem

When the user resizes their terminal, tmux proportionally scales all panes — including the HUD watch pane. The reconciler only enforces the canonical 3-line height on the *next prompt submit*, leaving the HUD oversized until then.

## Root cause

`reconcileHudForPromptSubmit` calls `resizeTmuxPane` on every prompt submit, but there is no equivalent enforcement on `client-resized`. tmux naturally redistributes space among all panes when the terminal dimensions change.

## Fix

After each successful resize or create in `reconcileHudForPromptSubmit`:

1. Write a small POSIX sh script to `.omx/state/hud-resize.sh` — it enumerates panes in the current window, filters for HUD watch panes (`grep "hud.*--watch"`), and calls `tmux resize-pane -y <height>` on each match.
2. Register it as a global tmux hook via `set-hook -g client-resized "run-shell -b '<script>'"`.

`set-hook -g` is idempotent — calling it on every prompt submit overwrites with the same script path. Hook registration and script write failures are caught and treated as non-critical so they cannot break the HUD lifecycle.

## New exports (`src/hud/tmux.ts`)

- `buildHudResizeScript(heightLines)` — returns the POSIX sh script content
- `writeHudResizeScript(scriptPath, heightLines)` — writes + chmods the script
- `registerHudResizeHook(scriptPath, execTmuxSync?)` — calls `set-hook -g`

Both `writeHudResizeScript` and `registerHudResizeHook` are injectable via `ReconcileHudForPromptSubmitDeps` for unit testing.

## Tests

Two new test cases in `reconcile.test.ts`:
- Verifies hook is registered after resizing an existing HUD pane
- Verifies hook is registered after creating a new HUD pane

Existing tests pass unchanged — the `try/catch` in `ensureHudResizeHook` silently absorbs the `mkdirSync`/tmux failures that occur in test environments without a real filesystem or tmux session.

## Note on `set-hook -g`

This overwrites any existing global `client-resized` hook. If composability becomes a concern, a follow-up could use a named hook suffix or check `show-hooks` first. For now the pattern matches the existing `resizeTmuxPane` approach of just enforcing state on every reconcile.

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -161,4 +161,50 @@ describe('reconcileHudForPromptSubmit', () => {
     assert.equal(resized[0]?.paneId, '%2');
     assert.equal(resized[0]?.heightLines, 3);
   });
+
+  it('registers client-resized hook after resizing an existing HUD pane', async () => {
+    const written: Array<{ path: string; lines: number }> = [];
+    const registered: string[] = [];
+
+    await reconcileHudForPromptSubmit('/repo', {
+      env: { TMUX: '1', TMUX_PANE: '%1', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+      listCurrentWindowPanes: () => [
+        { paneId: '%1', currentCommand: 'codex', startCommand: 'codex' },
+        { paneId: '%2', currentCommand: 'node', startCommand: 'node omx hud --watch' },
+      ],
+      resizeTmuxPane: () => true,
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Thanks for the PR. This repository takes contribution PRs against `dev`, not `main`, and this PR currently targets `main`, so I’m closing it per the branch policy.

Please reopen/recreate the same change against `dev` and keep the scope focused on the HUD resize behavior.

—

PR #2285, opened 2026-05-11
hud: register client-resized hook to re-enforce pane height on terminal resize

grndlvl:
## Problem

When the user resizes their terminal, tmux proportionally scales all panes — including the HUD watch pane. The reconciler only enforces the canonical 3-line height on the *next prompt submit*, leaving the HUD oversized until then.

## Root cause

`reconcileHudForPromptSubmit` calls `resizeTmuxPane` on every prompt submit, but there is no equivalent enforcement on `client-resized`. tmux naturally redistributes space among all panes when the terminal dimensions change.

## Fix

After each successful resize or create in `reconcileHudForPromptSubmit`:

1. Write a small POSIX sh script to `.omx/state/hud-resize.sh` — it enumerates panes in the current window, filters for HUD watch panes (`grep "hud.*--watch"`), and calls `tmux resize-pane -y <height>` on each match.
2. Register it as a global tmux hook via `set-hook -g 'client-resized[99]' "run-shell -b '<script>'"`.

Using an indexed slot (`[99]`) ensures we never clobber existing user or plugin hooks at other indices. The index is stable so repeated calls on every prompt submit are idempotent. Hook registration and script write failures are caught and treated as non-critical so they cannot break the HUD lifecycle.

## New exports (`src/hud/tmux.ts`)

- `buildHudResizeScript(heightLines)` — returns the POSIX sh script content
- `writeHudResizeScript(scriptPath, heightLines)` — writes + chmods the script
- `registerHudResizeHook(scriptPath, execTmuxSync?)` — calls `set-hook -g 'client-resized[99]'`

Both `writeHudResizeScript` and `registerHudResizeHook` are injectable via `ReconcileHudForPromptSubmitDeps` for unit testing.

## Tests

Two new test cases in `reconcile.test.ts`:
- Verifies hook is registered after resizing an existing HUD pane
- Verifies hook is registered after creating a new HUD pane

Existing tests pass unchanged — the `try/catch` in `ensureHudResizeHook` silently absorbs the `mkdirSync`/tmux failures that occur in test environments without a real filesystem or tmux session.

## Addressing #2284 review feedback

Previous PR used bare `set-hook -g client-resized` which clears all existing hooks at that hook point. This version uses `client-resized[99]` (stable indexed slot) so only the OMX-owned slot is touched.

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -161,4 +161,50 @@ describe('reconcileHudForPromptSubmit', () => {
     assert.equal(resized[0]?.paneId, '%2');
     assert.equal(resized[0]?.heightLines, 3);
   });
+
+  it('registers client-resized hook after resizing an existing HUD pane', async () => {
+    const written: Array<{ path: string; lines: number }> = [];
+    const registered: string[] = [];
+
+    await reconcileHudForPromptSubmit('/repo', {
+      env: { TMUX: '1', TMUX_PANE: '%1', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+      listCurrentWindowPanes: () => [
+        { paneId: '%1', currentCommand: 'codex', startCommand: 'codex' },
+        { paneId: '%2', currentCommand: 'node', startCommand: 'node omx hud --watch' },
+      ],
+      resizeTmuxPane: () => true,
+      writeHudResizeScript: (path, lines) => { written.push({ path, lines }); },
+      registerHudResizeHook: (path) => { registered.push(path); return true; },
+      resolveOmxCliEntryPath: () => '/repo/dist/cli/omx.js',
+    });
+
+    assert.equal(written.length, 1);
+    assert.match(written[0]?.path ?? '', /\.omx[/\\]state[/\\]hud-resize\.sh$/);
+    assert.equal(written[0]?.lines, 3);
+    assert.equal(registered.length, 1);
+    assert.equal(registered[0], written[0]?.path);
+  });
+
+  it('registers client-resized hook after creating a new HUD pane', async () => {
+    const written: Array<{ path: string; lines: number }> = [];
+    const registered: string[] = [];
+
+    await reconcileHudForPromptSubmit('/repo', {
+      env: { TMUX: '1', TMUX_PANE: '%1', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+      listCurrentWindowPanes: () => [
+        { paneId: '%1', currentCommand: 'codex', startCommand: 'codex' },
+      ],
+      createHudWatchPane: () => '%9',
+      resizeTmuxPane: () => true,
+      writeHudResizeScript: (path, lines) => { written.push({ path, lines }); },
+      registerHudResizeHook: (path) => { registered.push(path); return true; },
+      resolveOmxCliEntryPath: () => '/repo/dist/cli/omx.js',
+    });
+
+    assert.equal(written.length, 1);
+    assert.match(written[0]?.path ?? '', /\.omx[/\\]state[/\\]hud-resize\.sh$/);
+    assert.equal(written[0]?.lines, 3);
+    assert.equal(registered.length, 1);
+    assert.equal(registered[0], written[0]?.path);
+  });
 });
--- a/src/hud/reconcile.ts
+++ b/src/hud/reconcile.ts
@@ -1,3 +1,4 @@
+import { join } from 'node:path';
 import { readHudConfig } from './state.js';
 import { HUD_TMUX_HEIGHT_LINES } from './constants.js';
 import {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — thanks for reopening this against `dev`. The symptom is real, but this implementation is not safe to merge as-is.

Blockers:

1. **The tmux hook is server-global, not scoped.**
   - `src/hud/tmux.ts` installs `set-hook -g client-resized[99] ...`.
   - That affects the whole tmux server, not just the OMX session/window/project that owns this HUD.

2. **The fixed hook slot can clobber unrelated hooks.**
   - `client-resized[99]` is hardcoded.
   - Any user hook or another OMX project/session using the same slot will be overwritten.
   - Multiple checkouts/sessions installing this hook will race, with the last registration winning globally.

3. **The generated script path is multi-project/session unsafe.**
   - The global hook points to one repo-local path: `<cwd>/.omx/state/hud-resize.sh`.
   - If another project registers later, the global hook now points there instead.
   - If that checkout is deleted/moved, the tmux hook becomes stale.
   - There is no unregister/cleanup path.

4. **The resize script matches too broadly.**
   - It runs `tmux list-panes ... | grep "hud.*--watch" ...` without `-t` target scoping or ownership checks.
   - That can resize unrelated HUD panes in other OMX sessions/windows/projects.

5. **Portability risk.**
   - The script hardcodes `#!/bin/sh`, `tmux`, `grep`, and `awk`.
   - Existing tmux code already has platform-aware handling in several places; this would be risky for native Windows / psmux-compatible paths.

Validation from OMX review session `omx-pr-2285-hud-resize-review`:
- `npm ci` passed
- `npm run build` passed
- `node --test dist/hud/__tests__/reconcile.test.js` passed — 9/9
- `node --test dist/team/__tests__/tmux-session.test.js` passed — 183/183
- A broader `node --test dist/team/__tests__/tmux-session.test.js dist/cli/__tests__/index.test.js` run showed 14 failures in unrelated CLI areas; I did not treat those as blockers for this PR, but they were not proven baseline either.

Suggested direction:
- Register hooks scoped to the relevant tmux session/window where possible, not `-g`.
- Use collision-resistant hook slot naming/indexing, similar to existing team tmux hook patterns.
- Bind the hook/script to the current OMX-owned HUD pane/session/window, not broad `hud.*--watch` matching.
- Add unregister/cleanup behavior for stale hooks.
- Preserve platform-aware tmux resolution / avoid assuming POSIX shell tools where OMX supports other runtime paths.

—

Yeachan-Heo:
REQUEST_CHANGES from review gate `omx-pr-2285-hud-resize-review`.

Targeted validation passed, but I would not merge this shape because it introduces broad persistent tmux-global behavior without enough scoping/cleanup.

Evidence reviewed:
- PR #2285 `fix/hud-pane-grows-on-terminal-resize-v2` → `dev`
- Diff scope: `src/hud/tmux.ts`, `src/hud/reconcile.ts`, `src/hud/__tests__/reconcile.test.ts`
- Local validation in `/tmp/omx-pr-2285-review`:
  - `pnpm exec tsc --noEmit` passed using parent repo `node_modules` symlink
  - `node --test dist/hud/__tests__/reconcile.test.js dist/hud/__tests__/hud-tmux-injection.test.js` passed, 29/29
  - `pnpm exec biome lint src/hud/reconcile.ts src/hud/tmux.ts src/hud/__tests__/reconcile.test.ts` passed
  - `pnpm install --frozen-lockfile` failed only because this repo has no `pnpm-lock.yaml`; not treated as blocking
- GitHub checks visible at review time: PR Size Label success, Draft Check success, PR Base Guidance skipped

Blocking risks:
1. `set-hook -g client-resized[99] ...` is tmux-server/global, not session/window scoped, and may overwrite/conflict with user or plugin hooks using the same index.
2. No cleanup/unregister semantics. The hook can persist after the HUD pane/session exits and may continue referencing `.omx/state/hud-resize.sh` after the project/session is gone.
3. Fixed `[99]` is not namespaced/owned by session/window/pane. Re-running from another repo/session can replace the global hook path with the latest repo’s script.
4. This is user-facing tmux behavior: OMX mutates global tmux hook state, observable across tmux clients/sessions, so it needs explicit owner/design alignment even after code fixes.
5. The hook command scans panes via `grep "hud.*--watch"` rather than targeting the specific OMX-owned HUD pane id. Existing team tmux patterns are more scoped and have unregister helpers.

Recommended fix:
- Use a scoped hook target instead of global `set-hook -g` where possible.
- Use a deterministic OMX-owned hook slot/name instead of bare fixed `[99]`.
- Store ownership metadata and unregister on HUD/session cleanup.
- Prefer targeting the exact HUD pane id rather than scanning all panes.
- Add tests for existing user hook preservation, unregister/cleanup, repeated registration, multi-repo/multi-session behavior, and hook command/path quoting.

—

Yeachan-Heo:
Closing this because the current revision has merge-blocking review findings.

Please open a fresh PR against `dev` once the hook is scoped safely and no longer uses a global persistent `client-resized[99]` hook / broad HUD pane matching.

—

PR #2308, opened 2026-05-13
Add opt-in cmux mux adapter

godol811:
## Summary
- Add an explicit tmux/cmux mux adapter path while keeping tmux as the default behavior.
- Translate the prompt-injection tmux argv subset to cmux argv for send, enter, capture, display-message, and list-panes calls.
- Preserve downstream tmux stdout contracts for `list-panes -F ...`, `#{pane_in_mode}`, and target-aware `display-message` metadata probes when routed through cmux.
- Add a Rust `CmuxAdapter` scaffold, attached-cmux native hook context, mock-binary tests, and cmux support documentation.

## Review fixes
- Preserve pane metadata by normalizing `cmux list-pane-surfaces` output back to the requested tmux `-F` fields.
- Keep the scroll/copy-mode guard observable by probing `cmux surface-health` instead of hardcoding `#{pane_in_mode}` to `0`.
- Preserve target-aware `#{pane_current_command}`, `#{pane_start_command}`, and `#{pane_current_path}` probes through `cmux identify`; missing metadata now produces conservative sentinels rather than empty successful output.

## Notes
- cmux is selected only with `OMX_MUX=cmux`, or with `OMX_MUX=auto` plus `CMUX_SURFACE_ID`/`CMUX_WORKSPACE_ID`.
- A live cmux install is not required in CI; tests use mock binaries and argument-contract checks.
- Broad `tmux-*` to `mux-*` renames are intentionally deferred to keep this compatibility PR reviewable.

## Verification
- `npm run build`
- `npm run lint`
- `node --test dist/hooks/__tests__/mux-adapter.test.js dist/hooks/__tests__/notify-hook-team-tmux-guard.test.js dist/hooks/__tests__/notify-hook-managed-tmux.test.js dist/scripts/__tests__/codex-native-hook.test.js`
- `cargo test -p omx-mux`
- `git diff --check`

## Known validation gap
- `dist/hooks/__tests__/notify-hook-auto-nudge.test.js` still has 8 failures in this local environment, but the same failures were reproduced unchanged on `main` in a detached `/private/tmp` worktree, so they are treated as baseline/local-environment failures rather than regressions from this cmux adapter change.

### DIFF
--- a/crates/omx-mux/src/cmux.rs
+++ b/crates/omx-mux/src/cmux.rs
@@ -1,0 +1,235 @@
+use std::env;
+use std::process::Command;
+use std::thread;
+use std::time::Duration;
+
+use crate::types::{MuxAdapter, MuxError, MuxOperation, MuxOutcome, MuxTarget, SubmitPolicy};
+
+fn cmux_binary() -> String {
+    env::var("OMX_CMUX_BIN")
+        .or_else(|_| env::var("OMX_MUX_BIN"))
+        .unwrap_or_else(|_| "cmux".to_string())
+}
+
+fn run_cmux(args: &[&str]) -> Result<String, MuxError> {
+    let binary = cmux_binary();
+    let output = Command::new(&binary)
+        .args(args)
+        .output()
+        .map_err(|e| MuxError::AdapterFailed(format!("failed to run {binary}: {e}")))?;
+
+    if output.status.success() {
+        String::from_utf8(output.stdout)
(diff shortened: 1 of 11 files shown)


Yeachan-Heo:
Thanks for resubmitting, but this still targets `main`. Closing again per repository policy: feature/fix PRs must target `dev`, not `main`.

Please open the PR against `dev` if you want it reviewed.

—

PR #2312, opened 2026-05-13
Fail draft-only team startup after ready timeout

### LINKED ISSUE #2310: Regression/variant: `omx team` can return success after `ready_prompt_timeout` while workers only have an unexecuted trigger draft
## Summary

`omx team` can create the team state and open worker panes, but some interactive workers never start the assigned work.

In the failed run, each worker pane had a visible trigger such as:

```text
Read your inbox file now and execute Worker-1 assignment. Work autonomously, then report changed files and checks.
```

but the worker footer still showed `0 in · 0 out`, the tasks were not actually executed, and the leader-side run looked like it had already handed work to the workers.

This looks like a remaining regression/variant of the earlier interactive worker startup issues: the pane is alive, but visible typed text is being treated too optimistically even though there is no worker execution evidence yet.

## Environment

Observed locally:

- `oh-my-codex`: `v0.17.0`
- Codex CLI: `codex-cli 0.130.0`
- Node.js: `v24.11.0`
- Platform: `darwin arm64`
- Team shape: `omx team 3:executor ...`
- Worker surface: interactive Codex workers in tmux panes

## Actual behavior

After starting a team run:

1. Worker panes are created.
2. The inbox trigger text is visible in the worker panes.
3. The workers do not consume the trigger or begin execution.
4. Worker panes remain effectively idle (`0 in · 0 out`).
5. Tasks are not reliably moved into real worker execution.
6. The startup path can still be treated as successful because the worker pane is alive and `ready_prompt_timeout` is recorded as recoverable.

Screenshot attached separately:

Image

## Expected behavior

A live worker pane plus visible trigger text should not be enough to count as successful worker startup.

If readiness polling times out but the worker pane is still alive, OMX should:

1. record `ready_prompt_timeout` as a recoverable diagnostic, and
2. continue into the critical inbox dispatch path, and
3. wait for real dispatch/startup evidence where that path already supports it.

The startup path should only report success once OMX has stronger evidence than “the pane is alive and text was typed”.

## Suspected root cause

In the installed `oh-my-codex@0.17.0` package, `dist/team/runtime.js` has a startup branch with this shape:

```js
if (!ready) {
  const workerAlive = ...;
  if (workerAlive) {
    recordRecoverableStartupIssue(... reason: 'ready_prompt_timeout' ...);
    return { ok: true, workerIndex, workerName };
  }
  return { ok: false, ... };
}
```

That early success return skips the later `dispatchCriticalInboxInstruction(...)` path for this worker.

The result is the failure mode above: OMX can stop after a visible prompt/draft state instead of forcing the worker through the normal critical inbox dispatch path.

I also checked the upstream source shape on `main` on 2026-05-13. `src/team/runtime.ts` still appears to contain the same `ready_prompt_timeout` early-success branch before the later critical inbox dispatch call, so this looks like a current-source issue, not only a local npm package artifact.

## Related prior issues / PRs

This is not a brand-new class of problem, but the specific branch above still seems uncovered:

- #391 — worker pane has typed trigger text, but the draft is not actually consumed; manual `C-m` was needed.
- #1490 — team workers never claim tasks / no heartbeats; event log included `ready_prompt_timeout`.
- #1492 — workers stall on startup and mailbox dispatch can wedge mid-run.
- #1999 — increased startup evidence windows so slow workers are not classified dead too early.
- #2041 — handled prompt worker startup death.
- #2042 — treated ready prompts as startup evidence.

The remaining variant here is narrower: after `ready_prompt_timeout`, if the pane is still alive, startup returns success before the critical inbox dispatch path gets a chance to provide stronger evidence.

## Proposed fix direction

I have a local patch against the installed `oh-my-codex@0.17.0` package. It has two separable parts.

### 1. Minimal startup fix

In the interactive startup branch:

- keep `ready_prompt_timeout` as a recoverable diagnostic;
- do **not** return success only because the pane is alive;
- continue to `dispatchCriticalInboxInstruction(...)`;
- only set `startupReadyPromptObserved=true` when the ready prompt was actually observed.

Sketch:

```js
let startupReadyPromptObserved = false;

if (!ready) {
  const workerAlive = ...;
  if (workerAlive) {
    recordRecoverableStartupIssue(... reason: 'ready_prompt_timeout' ...);
    // continue to critical inbox dispatch; do not return success here
  } else {
    return { ok: false, workerIndex, workerName, reason: 'ready_prompt_timeout' };
  }
} else {
  startupReadyPromptObserved = true;
}

await dispatchCriticalInboxInstruction({
  ...,
  startupReadyPromptObserved,
});
```

### 2. Optional operator recovery API

The patch also adds:

```text
omx team api assign-task
```

This calls the existing runtime `assignTask(...)` flow so manual recovery can use one atomic operation instead of a fragile sequence of:

1. write worker inbox,
2. type a trigger into the pane,
3. claim/transition the task separately.

If maintainers prefer a smaller PR, this API addition can be split from the startup fix.

## Attached patch

Patch file generated locally:

[oh-my-codex-worker-dispatch-timeout-assign-task-0.17.0.patch](https://github.com/user-attachments/files/27692599/oh-my-codex-worker-dispatch-timeout-assign-task-0.17.0.patch)

Patch scope:

- `dist/team/runtime.js`
- `dist/team/api-interop.js`
- `dist/cli/team.js`
- `dist/team/__tests__/api-interop.test.js`
- `skills/team/SKILL.md`

For an upstream PR, the same change should probably be ported to the TypeScript source files instead of landing only in generated `dist/` output.

## Local validation

Completed locally against `oh-my-codex@0.17.0` installed package:

```bash
node --check dist/team/runtime.js
node --check dist/team/api-interop.js
node --check dist/cli/team.js
omx team api assign-task --help
git apply --check oh-my-codex-worker-dispatch-timeout-assign-task-0.17.0.patch
```

Additional checks:

- static regression check confirms `ready_prompt_timeout` no longer returns success before critical dispatch;
- API interop operation-list / operation-resolution tests passed for the new `assign-task` operation (`9/9` targeted tests).

Known validation gap:

- A full local `api-interop.test.js` run had unrelated existing failures around display team name / temporary state cleanup in my environment, so I am not claiming a full suite pass from this local package patch.

Yeachan-Heo:
## Summary
- keep `ready_prompt_timeout` as recoverable diagnostic evidence, but no longer return startup success solely because the worker pane is alive
- continue into the critical inbox dispatch/startup-evidence path after readiness timeout
- prevent a timeout worker with only a typed trigger draft from being counted as successful execution
- add a focused regression for readiness timeout followed by dispatch notification without worker startup evidence

Fixes #2310.

## Verification
- `npm run build`
- `node --test --test-name-pattern "startup direct trigger falls back|rejects ready-prompt timeout|records recoverable issue when tmux fallback" dist/team/__tests__/runtime.test.js`
- `npm run lint`
- `npm run check:no-unused`

## Review
- AI-slop cleanup gate: passed/no-op for changed files only
- Final changed-file review: APPROVE / architectural status CLEAR

—

### DIFF
--- a/src/team/__tests__/runtime.test.ts
+++ b/src/team/__tests__/runtime.test.ts
@@ -2240,6 +2240,118 @@ esac
     }
   });
 
+  it('startTeam rejects ready-prompt timeout when dispatch never produces startup evidence', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-runtime-ready-timeout-no-evidence-'));
+    const previousTmux = process.env.TMUX;
+    const previousTmuxPane = process.env.TMUX_PANE;
+    const previousLaunchMode = process.env.OMX_TEAM_WORKER_LAUNCH_MODE;
+    const previousWorkerCli = process.env.OMX_TEAM_WORKER_CLI;
+    const previousReadyTimeout = process.env.OMX_TEAM_READY_TIMEOUT_MS;
+    const previousStartupEvidenceTimeout = process.env.OMX_TEAM_STARTUP_EVIDENCE_TIMEOUT_MS;
+    const previousStartupDispatchRetries = process.env.OMX_TEAM_STARTUP_DISPATCH_RETRIES;
+    const teamName = `trt-${process.pid}-${Date.now().toString(36)}`;
+    let receiptNotifier: NodeJS.Timeout | null = null;
+    let runtimeTeamName: string | null = null;
+
+    try {
+      await withMockTmuxFixture(
+        {
+          dirPrefix: 'omx-runtime-ready-timeout-no-evidence-bin-',
+          tmuxScript: (tmuxLogPath) => `#!/bin/sh
+set -eu
+printf '%s\\n' "$*" >> "${tmuxLogPath}"
+case "$1" in
+  -V)
+    echo "tmux 3.4"
+    exit 0
+    ;;
+  display-message)
+    case "$*" in
+      *"#{window_width}"*) echo "120" ;;
+      *) echo "leader:0 %1" ;;
+    esac
+    exit 0
+    ;;
+  list-panes)
+    case "$*" in
+      *"pane_current_command"*) printf "%%1\\tnode\\t'codex'\\n" ;;
+      *"#{pane_dead} #{pane_pid}"*) echo "0 4242" ;;
+      *"#{pane_dead}"*) echo "0" ;;
+      *"#{pane_pid}"*) echo "4242" ;;
+      *) exit 0 ;;
+    esac
+    exit 0
+    ;;
+  capture-pane)
+    printf 'worker process is still starting; no agent prompt yet\\n'
+    exit 0
+    ;;
+  split-window)
+    echo "%2"
+    exit 0
+    ;;
+  set-hook|run-shell|select-layout|set-window-option|select-pane|send-keys|kill-pane|kill-session|resize-pane)
+    exit 0
+    ;;
+  *)
+    exit 0
+    ;;
+esac
+`,
+          binaries: [{ name: 'codex', content: fakeCodexNodeScript('process.stdin.resume();\n') }],
+        },
+        async ({ tmuxLogPath }) => {
+          delete process.env.TMUX;
+          process.env.TMUX_PANE = '%1';
+          process.env.OMX_TEAM_WORKER_LAUNCH_MODE = 'interactive';
+          process.env.OMX_TEAM_WORKER_CLI = 'codex';
+          process.env.OMX_TEAM_READY_TIMEOUT_MS = '5000';
+          process.env.OMX_TEAM_STARTUP_EVIDENCE_TIMEOUT_MS = '500';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
OMX review session `omx-pr-2312-ready-timeout-review` produced a merge-readiness result; the tmux session was interrupted before a final one-line verdict, so I verified the terminal evidence and reran the interrupted checks.

Verdict: MERGE_READY.

Evidence:
- GitHub PR state: mergeable, all required checks green.
- `git diff --check origin/dev...refs/remotes/pr/2312`: PASS.
- Focused build/test worktree `/tmp/omx-pr2312-review` on PR head built successfully.
- Targeted runtime tests passed, including `startup direct trigger falls back to evidence-gated dispatch`.
- Broader regression probe found unrelated failures that also reproduce on `origin/dev`, so they are not PR blockers.
- Reran the interrupted checks locally: `npm run lint && npm run check:no-unused` PASS.

Risk summary: scoped fix to team runtime ready-prompt timeout/draft-only startup behavior with tests. No blockers found.

—

PR #2322, opened 2026-05-13
Accept compact Lore commits with OmX provenance

### LINKED ISSUE #2321: Align Lore commit guard with body-optional commit messages
## Summary

The native `PreToolUse` git commit guard currently rejects some inline commit messages that include a valid subject, a blank separator, Lore trailers, and the exact `Co-authored-by: OmX <omx@oh-my-codex.dev>` trailer when the message omits a separate narrative body paragraph.

This seems stricter than the Lore template, which describes the body as optional and asks agents to use trailers only when they add decision context. In practice, Codex can respond to the generic block reason by repeatedly trying longer Lore variants or by attempting the `OMX_LORE_COMMIT_GUARD=0` opt-out, even though the message already carries the required OmX co-author trailer.

## Observed behavior

A command shaped like this is blocked:

```sh
git commit \
  -m "Launch lvisai.xyz intro site" \
  -m "Constraint: lvisai.xyz needs a no-build static intro site that can deploy from GitHub Pages.
Tested: python3 -m http.server 4173 and curl checks

Co-authored-by: OmX <omx@oh-my-codex.dev>"
```

The visible reason remains the broad message that the commit must satisfy Lore format and include the OmX co-author trailer, even though the co-author trailer is present. The specific missing requirement is the narrative body paragraph.

## Expected behavior

When the inline command is inspectable, has a non-empty intent-first subject, has the normal blank separator, and includes the exact OmX co-author trailer, the guard should accept the compact form. Full Lore messages with a narrative body and decision trailers should continue to be accepted.

## Scope note

This is not a request to disable the default Lore guard. The opt-out should remain explicit for teams that want conventional commits. This issue is about making the enabled guard match the documented optional body/trailer contract while preserving the current stricter Lore format as a supported path.

## Suggested fix

Allow a compact inline path when `Co-authored-by: OmX <omx@oh-my-codex.dev>` is present after a valid subject/separator, and add regression coverage for both compact co-author-only messages and body-omitted messages with Lore trailers.

ken-jo:
## Summary

- allow compact inline Lore commits when the subject/separator are valid and the exact OmX co-author trailer is present
- keep the existing full Lore commit format working as-is
- document the compact accepted shape and add regression coverage for body-optional commit messages

## Context

This addresses #2321. The current guard requires a narrative body paragraph even though the Lore template describes the body as optional and says decision trailers should be used only when they add context. That mismatch can make Codex repeatedly retry longer messages or attempt the explicit opt-out path even when the commit already includes `Co-authored-by: OmX <omx@oh-my-codex.dev>`.

The change keeps the default Lore guard enabled and does not alter external file/editor/reuse/fixup message blocking. It only makes the enabled inline guard accept the compact OmX-provenance path.

## Tests

- `npm run build`
- `npm run lint -- src/scripts/codex-native-pre-post.ts src/scripts/__tests__/codex-native-hook.test.ts`

Closes #2321

### DIFF
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -85,6 +85,13 @@ native `PreToolUse` checks, including document-refresh warnings and command
 safety checks, still run. `omx doctor` reports when the guard is explicitly
 disabled by environment or config.
 
+When the guard is enabled, the required inline shape is intentionally small:
+an intent-first subject, the normal blank separator, and the exact
+`Co-authored-by: OmX <omx@oh-my-codex.dev>` trailer. A narrative body and
+decision trailers such as `Constraint:` or `Tested:` remain supported for
+changes that need decision context, but the guard should not require them when
+the commit message already carries the OmX co-author trailer.
+
 Warning scope is intentionally narrow and rule-scoped:
 
 - **Commit path:** `PreToolUse` is Bash-only in this MVP and evaluates only
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -4403,6 +4403,60 @@ exit 0
     }
   });
 
+  it("stays silent on PreToolUse for compact git commit with only the OmX co-author trailer", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-pretool-git-commit-compact-coauthor-"));
+    try {
+      const result = await dispatchCodexNativeHook(
+        {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
Thanks for the contribution. Closing this PR because it targets `main`; this repo takes code changes through `dev`.

I already have an OMX session running from issue #2321 against `origin/dev`, so the fix path will continue there with regression coverage and a `dev`-base PR.

—

Yeachan-Heo:
Closing because this external PR targets main instead of dev; #2321 is being handled on a dev-base branch.

PR #2477, opened 2026-05-23
[codex] Respect plugin-owned hooks without overwriting user surfaces

joshvera:
## Note

Hey @Yeachan-Heo, didn't really mean to open this against upstream, but given clawdbot seems receptive, here it is against the dev branch!

## Summary

- Treat existing `hooks.json` as user-owned when plugin-scoped hooks are enabled.
- Preserve symlinked `AGENTS.md` during plugin-mode cleanup so dotfiles-managed installs are not removed.
- Strip stale managed hook trust-state tables for the old setup-owned hook path during plugin-mode setup.

## Root cause

`omx doctor` checked setup-owned `hooks.json` before recognizing that plugin mode with `plugin_hooks = true` validates hooks through the plugin cache. If a user kept their own `~/.codex/hooks.json` entry, doctor reported missing OMX-managed native coverage even though plugin-scoped hook coverage was correct.

## Validation

- `npm run build`
- `npm run lint`
- `npm run check:no-unused`
- `node --test dist/config/__tests__/generator-idempotent.test.js dist/cli/__tests__/doctor-warning-copy.test.js`
- `node --test --test-name-pattern 'preserves symlinked user-scope AGENTS.md during plugin-mode cleanup' dist/cli/__tests__/setup-agents-overwrite.test.js`
- Local `omx setup --force` followed by `omx doctor` reports `16 passed, 0 warnings, 0 failed`.

### DIFF
--- a/src/cli/__tests__/doctor-warning-copy.test.ts
+++ b/src/cli/__tests__/doctor-warning-copy.test.ts
@@ -835,6 +835,75 @@ OMX_LORE_COMMIT_GUARD = "truee"
 		}
 	});
 
+	it("accepts plugin-scoped native hooks when hooks.json contains user-owned hooks", async () => {
+		const wd = await mkdtemp(join(tmpdir(), "omx-doctor-plugin-scoped-hooks-user-"));
+		try {
+			const home = join(wd, "home");
+			const codexDir = join(home, ".codex");
+			await mkdir(join(wd, ".omx"), { recursive: true });
+			await mkdir(codexDir, { recursive: true });
+			const cacheDir = await installPluginCacheFixture(codexDir);
+			await writeFile(
+				join(wd, ".omx", "setup-scope.json"),
+				`${JSON.stringify({ scope: "user", installMode: "plugin", mcpMode: "none" }, null, 2)}\n`,
+			);
+			await writeFile(
+				join(codexDir, "config.toml"),
+				[
+					"plugin_hooks = true",
+					"goals = true",
+					"",
+					"[marketplaces.oh-my-codex-local]",
+					'source_type = "local"',
+					`source = ${JSON.stringify(repoRoot())}`,
+					"",
+					'[plugins."oh-my-codex@oh-my-codex-local"]',
+					"enabled = true",
+					"",
+				].join("\n"),
+			);
+			await writeFile(
+				join(codexDir, "hooks.json"),
+				JSON.stringify(
+					{
+						hooks: {
+							Stop: [
+								{
+									hooks: [
+										{
+											type: "command",
+											command: "/usr/bin/python3 /tmp/user-notify.py",
+											timeout: 5,
+										},
+									],
+								},
+							],
+						},
+					},
+					null,
+					2,
+				),
+			);
+
+			const res = runOmx(wd, ["doctor"], {
+				HOME: home,
+				CODEX_HOME: codexDir,
+			});
+			if (shouldSkipForSpawnPermissions(res.error)) return;
+			assert.equal(res.status, 0, res.stderr || res.stdout);
+			assert.match(
+				res.stdout,
+				new RegExp(
+					`\\[OK\\] Native hooks: plugin-scoped hooks are enabled; existing hooks\\.json at .*\\.codex[\\/]+hooks\\.json is treated as user-owned because plugin-scoped hooks are enabled, and plugin cache native hook coverage smoke passed via ${join(cacheDir, "hooks", "hooks.json").replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}`,
+				),
+			);
+			assert.doesNotMatch(res.stdout, /hooks\.json is missing OMX-managed coverage/);
+			assert.doesNotMatch(res.stdout, /run "omx setup --force" to restore native hooks/);
+		} finally {
+			await rm(wd, { recursive: true, force: true });
+		}
+	});
+
 	it("warns when hooks.json is missing OMX-managed native hook coverage", async () => {
 		const wd = await mkdtemp(join(tmpdir(), "omx-doctor-hooks-coverage-"));
 		try {
--- a/src/cli/__tests__/setup-agents-overwrite.test.ts
+++ b/src/cli/__tests__/setup-agents-overwrite.test.ts
@@ -1,6 +1,6 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
-import { mkdtemp, mkdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { mkdtemp, mkdir, readFile, readlink, rm, symlink, writeFile } from 'node:fs/promises';
 import { existsSync, readFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
@@ -296,6 +296,40 @@ describe('omx setup AGENTS refresh behavior', () => {
       assert.doesNotMatch(output, /Refreshed AGENTS\.md model capability table/);
       assert.equal(await readFile(join(wd, 'AGENTS.md'), 'utf-8'), existing);
       assert.equal(existsSync(join(wd, '.omx', 'backups', 'setup')), false);
+    } finally {
+      restoreHome();
+      restoreTty();
+      await rm(wd, { recursive: true, force: true });
+    }
+  });
+
+  it('preserves symlinked user-scope AGENTS.md during plugin-mode cleanup', async () => {
+    const wd = await mkdtemp(join(tmpdir(), 'omx-setup-plugin-agents-symlink-'));
+    const restoreTty = setMockTty(false);
+    const home = join(wd, 'home');
+    const restoreHome = setMockHome(home);
+    const dotfilesAgentsPath = join(wd, 'dotfiles', '.codex', 'AGENTS.md');
+    const codexAgentsPath = join(home, '.codex', 'AGENTS.md');
+    try {
+      await mkdir(join(wd, '.omx', 'state'), { recursive: true });
+      await mkdir(join(wd, 'dotfiles', '.codex'), { recursive: true });
+      await mkdir(join(home, '.codex'), { recursive: true });
+      await writeFile(
+        dotfilesAgentsPath,
+        addGeneratedAgentsMarker('# oh-my-codex - Intelligent Multi-Agent Orchestration\n\nDotfiles-owned guidance.\n'),
+      );
+      await symlink(dotfilesAgentsPath, codexAgentsPath);
+
+      const output = await runSetupWithCapturedLogs(wd, {
+        scope: 'user',
+        installMode: 'plugin',
+        force: true,
+      });
+
+      assert.match(output, /AGENTS\.md generation skipped; no legacy OMX-generated AGENTS\.md found and defaults not selected\./);
+      assert.equal(await readlink(codexAgentsPath), dotfilesAgentsPath);
+      assert.match(await readFile(codexAgentsPath, 'utf-8'), /Dotfiles-owned guidance\./);
+      assert.match(output, /agents_md: updated=0, unchanged=0, backed_up=0, skipped=1, removed=0/);
     } finally {
       restoreHome();
       restoreTty();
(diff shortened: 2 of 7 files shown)


Yeachan-Heo:
Thanks for the quick re-open against `dev` @joshvera 🦞

I've approved the cross-repo CI run (it was paused on `action_required` since this is your first PR to the repo). The full lane should kick off momentarily — you'll see Build / Typecheck / Lint / Test (Node 20 lanes) / Coverage / Rust lanes etc. fill in.

Once CI is green I'll spin up an OMC review session. The change set looks tight from a quick scan (the three-part hooks/AGENTS.md/trust-state interaction is well-scoped), but the actual review needs to verify:

- the user-owned vs OMX-managed `hooks.json` discrimination doesn't regress the existing plugin-mode `omx setup --force` idempotency tests
- the symlinked AGENTS.md preservation handles the edge case where the symlink target lives outside the user's home (dotfiles-managed installs sometimes do this)
- the stale-trust-state strip is bounded to the old setup-owned hook path and doesn't reach into unrelated trust tables

If CI surfaces anything, I'll comment here. Otherwise the next ping will be either the review verdict or a merge.

—

Yeachan-Heo:
## Review — PR #2477 (work-stream: external contributor first PR)

**Verdict:** APPROVE

**Summary**: When plugin mode with `plugin_hooks = true` is enabled and the user keeps their own `~/.codex/hooks.json`, doctor previously reported missing OMX-managed native hook coverage because the plugin-scoped check was gated behind `!existsSync(hooksPath)`. This PR lifts the plugin-scoped hook check to run *before* checking hooks.json existence, preserves symlinked AGENTS.md during plugin-mode cleanup, and strips orphaned managed hook trust-state tables for the old setup-owned hook path.

### Risk Checklist
- **R1 Idempotency**: PASS — `stripManagedCodexHookTrustState` first strips fenced blocks then orphaned tables for the given `hooksPath`. On second run with no orphans, the strip is a no-op. The new test "removes orphaned managed hook trust-state tables for the setup hooks path" validates this and confirms valid TOML output.
- **R2 Non-plugin regression**: PASS — All three change sites are gated on plugin mode: doctor's early check requires `context.installMode === "plugin"`, `applyPluginModeHooksConfig` is only called from plugin-mode setup, and `cleanupPluginModeLegacyAgentsMd` is only called from plugin-mode cleanup. The `stripManagedCodexHookTrustState` signature change defaults `options = {}`, so existing callers without `hooksPath` (4 of 5 call sites) get unchanged behavior.
- **R3 AGENTS.md symlink edge cases**: PASS — Broken symlink: `existsSync` returns false for broken symlinks, so the function returns `false` at the first guard without removing anything. Symlink outside home: `lstat` doesn't follow symlinks; if the inode is a symlink regardless of target location, it returns `false` and preserves. Both edge cases are handled correctly by the code logic, though the test only covers the valid-symlink case.
- **R4 Trust-state strip bounds**: PASS — `stripOrphanedManagedCodexHookTrustState` constructs headers matching ONLY `[hooks.state."<hooksPath>:<eventLabel>:0:0"]` for the 7 known event labels. The test explicitly verifies `custom:/hooks.json:stop:0:0` is preserved while `/tmp/codex/hooks.json:*` tables are stripped.
- **R5 User-owned hooks.json heuristic**: PASS — The heuristic is: if `plugin_hooks = true` in config.toml AND `hooks.json` exists at the setup hooks path → treat as user-owned. This is robust because when plugin_hooks is enabled, hooks are delivered through the plugin cache, not through setup-owned hooks.json. Any existing hooks.json is either user-owned or stale OMX state (harmless in both cases since plugin cache coverage is validated separately).
- **R6 Trust-state strip reversibility**: PASS — Trust state in config.toml is backed up by `ensureBackup` before modification during setup. Even without the backup, trust state can be regenerated via `omx setup --force`.
- **R7 Tests cover fix surface**: PASS — (1) doctor test exercises the exact bug path: hooks.json exists with user-owned hooks + plugin_hooks=true → verifies OK message with "treated as user-owned". (2) setup test exercises symlink preservation → verifies symlink target and content unchanged. (3) generator test exercises orphaned trust-state strip → verifies managed tables removed while user tables preserved.
- **R8 Naming hygiene**: PASS — `MANAGED_CODEX_HOOK_EVENT_LABELS`, `stripOrphanedManagedCodexHookTrustState`, `setupHooksPathDescription` are clear and consistent with existing conventions. No TODOs/FIXMEs. `lstat` import addition is clean.

### Validation
- `npm run build`: pass
- `npm run lint:deps`: N/A (script does not exist in this repo; equivalent `check:no-unused` was run by the PR author)
- `npm run typecheck` (`npx tsc --noEmit`): pass (zero errors)
- PR-claimed tests: **89 pass, 0 fail** across generator-idempotent, doctor-warning-copy, setup-agents-overwrite
- Regression-adjacent test: `codex-hooks.test.js` — **24 pass, 0 fail**

### Findings
**Blocking** (must address before merge):
- None.

**Non-blocking observations** (nice-to-have, not required):
- The AGENTS.md symlink test only covers the happy path (valid symlink within temp dir). The broken-symlink and outside-home-symlink edge cases are handled correctly by code logic (`existsSync` + `lstat`), but explicit test cases would strengthen the contract if this area sees future churn.
- The `OMX_HOOK_TRUST_END_MARKER` stripping in `stripOrphanedManagedCodexHookTrustState` is unconditional (strips all stray end markers, not just ones tied to `hooksPath`). This is correct post-fenced-block-strip but worth a sentence in a code comment if someone later reads the function in isolation.

### Recommendation
Merge as-is. The three-part change is well-scoped and correctly fixes the interaction between user-owned hooks.json and plugin-scoped hook validation. All risk items pass, validation is clean, and the backward compatibility of the `stripManagedCodexHookTrustState` signature is verified across all 5 call sites. The two non-blocking observations are cosmetic and can be addressed in a follow-up if desired.

—

Yeachan-Heo:
## Re-review — PR #2477 (second commit on top of v1 APPROVE)

**Verdict (second commit only):** APPROVE
**Combined PR verdict:** APPROVE

**Second-commit summary**: `03b213db` replaces the key-name-based orphan stripping introduced in the first commit with a proof-of-ownership model. Trust-state tables matching a managed key are now only stripped when the body contains exactly the generated `trusted_hash` and no user metadata (extra keys, comments). When a same-key conflict is detected (user-modified content), the existing table is preserved and the generated duplicate is suppressed via `preservedConflictKeys` flow. This directly addresses the v1 non-blocking observation about plugin-mode `hooks.json` being user-owned state.

### Risk Checklist (second commit)
- **R1 Scope creep**: PASS — Every new helper (`decodeTomlBasicString`, `parseHooksStateHeader`, `isTomlTableHeader`, `isExactlyManagedHookTrustBody`, `stripProofManagedCodexHookTrustStateTables`, `renderManagedCodexHookTrustToml`, `buildManagedCodexHookTrustStateForConfig`) directly supports the proof-based ownership model. All removed functions (`MANAGED_CODEX_HOOK_EVENT_LABELS`, `stripOrphanedManagedCodexHookTrustState`, `managedCodexHookTrustStateHeaders`, `stripUnfencedManagedCodexHookTrustState`) are replaced, not tangentially cleaned up. No scope creep.
- **R2 Call-site compatibility**: PASS — `stripManagedCodexHookTrustState` signature changed from `{ hooksPath?: string }` to `{ managedTrustState?: ManagedCodexHookTrustStateMap }`. All callers verified at commit 2: `setup.ts:1569` updated to pass `{ managedTrustState }`, `setup.ts:1720` passes no options (unchanged behavior — fenced-block-only stripping for cleanup path), `generator.ts` internal callers (`upsertManagedCodexHookTrustState`, `buildMergedConfig`) use the new `WithResult` variant. `getOmxTablesBlock` gained `excludedHookTrustStateKeys` parameter with default `new Set()` — sole caller `buildMergedConfig` was updated. No dangling references to removed functions.
- **R3 v1 risk-checklist preservation**: PASS — (v1-R1 Idempotency) proof-based strip is idempotent: matched tables stripped, no matches on re-run. (v1-R2 Non-plugin regression) `setup.ts` changes remain plugin-gated; `buildMergedConfig` behavior equivalent for non-plugin since managed trust state is built from same source. (v1-R3 AGENTS.md symlink) Untouched by this commit. (v1-R4 Trust-state strip bounds) IMPROVED — requires hash proof, not just key matching. (v1-R5 User-owned hooks.json heuristic) Untouched. (v1-R6 Reversibility) Still backed by `ensureBackup` + `omx setup --force`. (v1-R7 Test coverage) IMPROVED — 4 new test groups. (v1-R8 Naming) New names consistent.
- **R4 New test quality**: PASS — `setup-install-mode.test.ts` adds an integration test that creates real temp dirs, writes real `hooks.json` + `config.toml`, calls real `setup({ scope: "user", installMode: "plugin" })`, and verifies: `plugin_hooks=true` set, user trust hash preserved, user extra keys preserved, no duplicate tables, valid TOML output. Exercises the full setup path through proof-based stripping — not overly mocked.
- **R5 idempotent test expansion**: PASS — 125 lines add: (1) updated orphan test using real generated hashes instead of fakes, (2) "preserves same-key user hook trust state and suppresses generated duplicates" via `upsertManagedCodexHookTrustState`, (3) "preserves unproven same-key hook trust-state tables" with 6 edge-case fixtures (missing hash, malformed hash, extra assignment, body comment, inline body comment, inline header comment), (4) "does not remove unfenced tables without proof map" ensuring backward-compatible no-options path. All directly load-bearing for trust-state-protection claim.
- **R6 Trust-state reversibility/safety**: PASS — Fenced blocks are still stripped unconditionally (unchanged). Proof-based stripping applies ONLY to unfenced/orphaned tables. If OMX updates change hook hashes, old unfenced managed tables won't match and will be conservatively preserved — the user can re-run `omx setup --force` to regenerate. This is strictly safer than the v1 key-name-based approach, which could accidentally strip user-authored trust state that happened to share a managed key pattern.
- **R7 Naming/structure**: PASS — `stripProofManagedCodexHookTrustStateTables` uses "Proof" prefix to distinguish from fenced-block stripping; `isExactlyManagedHookTrustBody` communicates strictness via "Exactly". No dead code detected — all new functions are called. No asymmetric helpers.
- **R8 Performance/IO**: PASS — `decodeTomlBasicString` calls `TOML.parse()` per hooks.state header, but there are typically <10 such headers per config. No extra disk reads/writes — all proof-based stripping operates on in-memory strings.

### Validation
- `npm run build`: pass
- `npx tsc --noEmit`: pass (zero errors)
- New + claimed tests: **148 pass, 0 fail** across 6 suites (generator-idempotent, setup-install-mode, doctor-warning-copy, setup-agents-overwrite)
- Regression-adjacent (`codex-hooks.test.js`): **24 pass, 0 fail**

### Findings
**Blocking** (must address before merge):
- None.

**Non-blocking observations**:
1. Stray `# End OMX-owned Codex hook trust state` markers (outside fenced blocks) are no longer stripped by the proof-based path. The v1 code explicitly removed them as a side effect of `stripOrphanedManagedCodexHookTrustState`. These are harmless TOML comments, so this is a cosmetic-only behavior change — but worth noting if anyone inspects generated configs for marker cleanliness.
2. If OMX updates change hook hashes between versions, unfenced managed trust-state tables from the old version won't be recognized as managed (hash mismatch) and will be conservatively preserved until `omx setup --force`. This is the safer default but could confuse users who expect trust state to auto-refresh. A targeted migration note in a future changelog would help.

### Recommendation
Merge as-is. The second commit directly addresses the v1 review's non-blocking observation about plugin-mode `hooks.json` being user-owned state. The proof-based ownership model is a strict improvement over key-name-based stripping: it eliminates false positives (user-authored tables with managed key patterns) while maintaining the same aggressive stripping for provably-managed state. The `preservedConflictKeys` flow correctly prevents duplicate regeneration. All risk items pass, validation is clean, and test coverage is comprehensive with 148 passing tests including 6 edge-case fixtures for the proof boundary.

—

Yeachan-Heo:
Merged into \`dev\` 🦞 — both commits landed cleanly with CI green across all lanes. Thanks for the careful split (key-name strip → proof-of-ownership refactor); it made re-review trivial. The plugin-mode hooks path is now safe for users keeping their own \`~/.codex/hooks.json\`.

—

PR #2533, opened 2026-05-26
fix: accept shared skill roots without duplicate names

iqdoctor:
## Summary
- Treat a shared `~/.agents/skills` root as healthy when it has no skill-name overlap with the canonical Codex skill root.
- Keep the existing warning only for actual duplicate skill names across the two roots.
- Add a doctor regression test for the no-overlap shared-skills case.

## Scope notes
This is intentionally narrow:
- changes only the `omx doctor` legacy/current skill-root overlap status for the no-overlap case
- preserves the existing warning when duplicate skill names exist across the roots
- does not change setup/uninstall migration copy or skill-root discovery semantics

## Local checks
- `npm run build`
- `node --test dist/cli/__tests__/doctor-warning-copy.test.js`
- `npm run lint -- src/cli/doctor.ts src/cli/__tests__/doctor-warning-copy.test.ts`
- `npm run check:no-unused`
- `git diff --check upstream/dev...HEAD`
- `npm run verify:native-agents && npm run verify:plugin-bundle`

## Fresh OMX review
A fresh tmux OMX code-review pass was run against this branch after opening the PR:
- verdict: APPROVE
- blocking findings: 0
- caveat: `omx setup` still has older cleanup-oriented wording for no-overlap legacy roots, but this PR is scoped to doctor behavior and preserves warnings for actual duplicates.

### DIFF
--- a/src/cli/__tests__/doctor-warning-copy.test.ts
+++ b/src/cli/__tests__/doctor-warning-copy.test.ts
@@ -804,6 +804,33 @@ OMX_LORE_COMMIT_GUARD = "truee"
 		}
 	});
 
+	it("passes when shared skill root exists without duplicate skill names", async () => {
+		const wd = await mkdtemp(join(tmpdir(), "omx-doctor-shared-skills-"));
+		try {
+			const home = join(wd, "home");
+			const codexDir = join(home, ".codex");
+			const canonicalPlan = join(codexDir, "skills", "plan");
+			const legacyShared = join(home, ".agents", "skills", "shared-context");
+			await mkdir(canonicalPlan, { recursive: true });
+			await mkdir(legacyShared, { recursive: true });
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR. I’m closing this one rather than taking the behavior change as-is.

This touches `omx doctor`’s skill-root diagnosis semantics. Even though the patch is narrow, accepting shared-root/no-overlap as healthy changes the warning contract around canonical vs shared skill directories. That needs to be handled from the maintainer side with the broader install/setup/doctor contract in view, not merged as an isolated external patch.

If this maps to a reproducible user-facing false positive we still want fixed, we’ll re-open the problem as a maintainer-owned issue/PR with the exact contract spelled out.

—

Yeachan-Heo:
Closing as a maintainer-owned doctor/setup contract decision rather than merging this external behavior change.

PR #2573, opened 2026-05-27
Prevent duplicate question renderer panes

### LINKED ISSUE #2572: [Bug] omx question splits a new tmux pane on every re-issue, leaving orphan question panes for the same session
## Summary

`omx question` (the blocking popup wizard used by `deep-interview` and other
skills) unconditionally creates a new `tmux split-window` on every invocation.
There is no check for an existing in-flight question pane belonging to the same
session.

When the calling agent re-issues `omx question` for the same logical turn (for
example, the deep-interview driver retries because it perceived a timeout,
edits a prompt and re-asks, or simply re-runs its bash block), each call spawns
an **additional** tmux pane. The prior question pane is **not killed**, its
node process keeps running indefinitely, and the user sees several overlapping
question panels stacked in the same tmux window.

In one observed `deep-interview` session, 6 question records produced only 4
clean closes; **2 orphan `omx question --ui` panes were still alive 5+ minutes
after the user had moved on**. Visually this shows up as the same question
panel appearing twice in the active OMX tmux pane.

## Environment

- OS: Linux 6.17.x (Ubuntu)
- Node.js version: v24.12.0
- `oh-my-codex` version: `0.18.5` (status bar reports `OMX#0.18.5`)
- Codex CLI version: gpt-5.5 medium fast (per OMX status line)
- tmux: a long-lived OMX-managed session

## Steps to reproduce

1. Start an OMX session and trigger a `deep-interview` (or any flow that uses
   `omx question` to render a blocking pane).
2. From within the agent's turn, issue `omx question` for the same question
   twice in quick succession. Any of these will do it in practice: bash retry,
   prompt edit + re-ask, two tool calls within seconds, a transient JSON-parse
   error followed by a retry.
3. Observe that two `tmux split-window` panes are now stacked in the OMX
   window, each rendering the same question, both holding a live
   `omx question --ui --state-path <distinct record file>` node process.

Minimal manual repro (same effect, no agent needed):

```bash
# inside an attached OMX tmux session
payload='{"question":"pick","type":"single-answerable","options":[{"label":"A","value":"a"},{"label":"B","value":"b"}],"source":"manual-repro"}'
omx question --input "$payload" --json &
sleep 1
omx question --input "$payload" --json &
# Two stacked panes appear; answering one leaves the other orphaned.
```

## Expected behavior

A second `omx question --input ...` call for the same session while a prior
question is still `prompting` should do one of:

- **Reject** the new launch with a structured error
  (e.g. `code: "question_already_in_flight"`, `existing_question_id: "..."`),
  letting the agent decide whether to wait or abort.
- **Supersede** the prior question: mark the prior record as `superseded`,
  `tmux kill-pane -t <prior target>`, and only then split the new pane.

Either policy keeps exactly one question pane per session at any time.

## Actual behavior

Each call to `omx question --input ...` runs the full launch path with no
prior-record awareness:

- `src/cli/question.ts` (around the `launchQuestionRenderer(...)` call) does
  not first scan `<session>/questions/` for records whose
  `status === "prompting"` and whose `renderer.target` is still alive.
- `src/question/renderer.ts` (`launchQuestionRenderer`, `inside-tmux` branch)
  unconditionally runs `tmux split-window ...` and persists the new pane id
  on the new record. It uses `isLaunchedQuestionPaneAlive` only to validate
  that the **just-created** pane started successfully — never to detect a
  sibling pane from a prior in-flight record.

Result: every re-issue is additive. The orphaned panes remain idle in raw
mode, never receive an answer, never write back to the return pane, and only
disappear when the user manually kills them or the whole tmux session ends.

## Evidence from a live session

Question records and pane targets for a single `deep-interview` session
(content redacted; only the lifecycle metadata is shown):

| Question file (Round) | status | renderer.target | pane alive? |
|---|---|---|---|
| R1 `question-...-dff9e79a.json` | answered | `%2` | no (clean) |
| R2 `question-...-3d56e743.json` | answered | `%3` | no (clean) |
| R3 try 1 `question-...-247fb90b.json` | **prompting** | `%4` | **YES — orphan** |
| R3 try 2 `question-...-eacc8480.json` | **prompting** | `%5` | **YES — orphan** |
| R3 try 3 `question-...-ad52fa1a.json` | answered | `%6` | no (clean) |
| R4 `question-...-29630dfb.json` | answered | `%8` | no (clean) |

`tmux list-panes` at observation time (paths abridged):

```
%0 cmd=sh   active=1  ← OMX active pane
%5 cmd=node active=0  ← orphan: omx question --ui --state-path .../question-...-eacc8480.json
%4 cmd=node active=0  ← orphan: omx question --ui --state-path .../question-...-247fb90b.json
%1 cmd=node active=0  ← HUD watcher (legitimate)
```

The deep-interview driver fired three `omx question` calls for Round 3
within 37 seconds (record timestamps `…:43:59` → `…:44:20` → `…:44:36`). Only
the third pane was answered; the first two are still rendering the same
Round 3 question 5+ minutes later, which is what the user sees as "the panel
duplicating itself".

## Proposed fix sketch

In `src/question/renderer.ts` add a helper, e.g.

```ts
export function findLiveQuestionsForSession(
  cwd: string,
  sessionId: string,
  execTmux: ExecTmuxSync = defaultExecTmux,
): Array<{ recordPath: string; record: QuestionRecord }> {
  // list <session>/questions/*.json where status === 'prompting'
  // and isQuestionRendererAlive(record.renderer) === true
}
```

Then in `launchQuestionRenderer` (or its caller in `src/cli/question.ts`)
before splitting:

1. `const inFlight = findLiveQuestionsForSession(cwd, sessionId)`
2. If `inFlight.length > 0`:
   - **Mode A (strict):** throw `QuestionAlreadyInFlightError(inFlight[0])`
     so the CLI returns
     `{ok:false, error:{code:'question_already_in_flight', existing_question_id, existing_pane}}`.
   - **Mode B (replace, default):** for each `inFlight` entry, mark the
     record as `superseded` (new terminal status), `tmux kill-pane -t
     <renderer.target>`, then proceed with the new split.

Add unit coverage in `src/question/__tests__/renderer.test.ts` with a fake
`execTmux` exercising both modes.

Cleanup for already-leaked panes: `omx doctor` (or a new `omx question --gc`
flag) could enumerate `prompting` records whose pane is alive but whose
record `updated_at` is older than N minutes, and offer to kill them /
mark them as `superseded`.

## Workaround for users hitting this now

```bash
# kill any omx question UI pane whose state-path file is older than N minutes
# (review the list before piping to xargs).
ps -eo pid,cmd | awk '/omx\.js question --ui --state-path/ {print $1}' \
  | xargs -r -n1 -I{} sh -c \
      'tmux list-panes -a -F "#{pane_id} #{pane_pid}" | awk -v p={} "\$2==p {print \$1}"' \
  | xargs -r -n1 tmux kill-pane -t
```

After this, mark the leaked records as `error`/`superseded` so the OMX state
store does not list them as still prompting.

Yeachan-Heo:
## Summary
- supersede any live prompting `omx question` renderer records for the same session before opening a new inside-tmux pane
- mark superseded records as terminal and close only their persisted renderer pane metadata
- add renderer coverage for live prompting dedupe, dead-pane no-op, answered no-op, and different-session isolation

## Tests
- `npm run build`
- `node --test dist/question/__tests__/renderer.test.js dist/question/__tests__/state.test.js dist/question/__tests__/ui.test.js`
- `npm run lint`
- `git diff --check`
- `npm run check:no-unused`

Closes #2572

### DIFF
--- a/src/question/__tests__/renderer.test.ts
+++ b/src/question/__tests__/renderer.test.ts
@@ -1,5 +1,5 @@
 import assert from 'node:assert/strict';
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { describe, it } from 'node:test';
@@ -9,9 +9,11 @@ import {
   formatQuestionAnswerForInjection,
   formatQuestionAnswersForInjection,
   injectQuestionAnswerToPane,
+  findLiveQuestionsForSession,
   injectQuestionAnswersToPane,
   launchQuestionRenderer,
   resolveQuestionRendererStrategy,
+  supersedeLiveQuestionsForSession,
 } from '../renderer.js';
 import { buildSendPaneArgvs } from '../../notifications/tmux-detector.js';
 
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
MERGE_READY

Reviewed and updated this PR to current `dev`; live head is `8e48173e818b450d66859b96e63de6f8f0943df1`.

Gate evidence:
- The change supersedes only live `prompting` question records for the same session before opening a new inside-tmux renderer pane.
- Superseded records become terminal via `status: "superseded"` with `question_superseded` metadata, so waiters do not hang on stale prompting records.
- Pane cleanup is scoped to persisted renderer metadata for prior live question panes; it does not target HUD panes, answered records, dead panes, or other sessions.
- The recent answered-pane cleanup behavior remains covered through the question renderer/state/UI focused suites.

Verification:
- `npm run build`
- `node --test dist/question/__tests__/renderer.test.js dist/question/__tests__/state.test.js dist/question/__tests__/ui.test.js`
- `npm run lint`
- `git diff --check`
- `npm run check:no-unused`
- Live GitHub checks are green and mergeState is `CLEAN` after branch update.

—

PR #2574, opened 2026-05-27
Fix duplicate worker Stop nudges

### LINKED ISSUE #2559: Team worker Stop nudge stacks duplicate `[OMX] native Stop allowed` injections in leader queue (no team-scope dedupe / shutdown gate)
# `maybeNudgeLeaderForAllowedWorkerStop` stacks duplicate "native Stop allowed" injections in the leader's queue; no team-scope dedupe or shutdown gate

**Repo:** https://github.com/Yeachan-Heo/oh-my-codex
**Affected file:** `dist/scripts/notify-hook/team-worker-stop.js` (source: `src/scripts/notify-hook/team-worker-stop.ts`)
**Package version observed:** `oh-my-codex@0.18.5`
**Node:** v24.12.0 (Linux)

## Summary

When multiple OMX team workers emit `Stop` near-simultaneously — or one worker keeps emitting `Stop` while the leader pane is busy — `maybeNudgeLeaderForAllowedWorkerStop` injects identical `[OMX] worker-N native Stop allowed. Run \`omx team status <team>\`, ...` prompts into the leader pane's input queue. The injection has:

- **Per-worker cooldown only** (no team-scope dedupe). N workers → up to N parallel nudges per cooldown window.
- **No team-shutdown gate.** After the team is cleanly shut down (state dir removed), in-flight Stop events that were resolved as `allowed` before the teardown still inject. Each injected prompt instructs the leader to run `omx team status <removed-team>`, which only returns "No team state found" — useless turns that the leader is still forced to process.
- **`queuePaneInput` path is unbounded.** When the leader has an active task, the function falls back to queueing without checking whether identical Stop messages are already queued.

Net effect: the leader CLI ends up with a long list of "Queued follow-up inputs" containing the exact same nudge, all pointing at a team that no longer exists. Cancelling/Esc-flushing them is the only practical recovery.

## Reproduction (observed)

1. Spawn an OMX team with ≥ 2 workers (here: `read-only-r13-closeou-430d3551`, workers `worker-1..3`).
2. While the leader pane is busy with an active task, let several workers' codex sessions emit native `Stop` events that resolve to `allowed` via `resolveTeamWorkerStopDecision` (i.e. their assigned tasks are `completed`).
3. Tear down the team cleanly (or let it auto-tear down) around the same time.

**Result:** the leader pane shows a stack of identical `Queued follow-up inputs`:

```
[OMX] worker-3 native Stop allowed. Run `omx team status read-only-r13-closeou-430d3551`, read worker messages/results, then assign next task, reconcile completion, or shut down. [OMX_TMUX_INJECT]
[OMX] worker-2 native Stop allowed. Run `omx team status read-only-r13-closeou-430d3551`, read worker messages/results, then assign next task, reconcile completion, or shut down. [OMX_TMUX_INJECT]
[OMX] worker-3 native Stop allowed. ...
[OMX] worker-2 native Stop allowed. ...
... (repeats ~10×)
```

Each queued message, when finally consumed, runs `omx team status read-only-r13-closeou-430d3551`, which reports:

```
No team state found for read-only-r13-closeou-430d3551
team state directory removed
```

Nothing actionable; just one wasted turn per queued copy.

## Root cause

In `dist/scripts/notify-hook/team-worker-stop.js`:

1. **Per-worker statePath** (line 104):
   ```js
   const statePath = join(workerDir, 'worker-stop-nudge.json');
   ```
   `workerDir` is `stateDir/team/<team>/workers/<workerName>`, so the cooldown file is per-worker. There is no team-scoped ledger checked before injection — if worker-2 nudged at T=0, worker-3 is free to nudge at T=0.001s.

2. **No team-shutdown gate.** The function checks `readTeamWorkersForIdleCheck(stateDir, teamName)` (line 113), but if the manifest still exists at hook-fire time and the team is concurrently being torn down, the inject still goes through. After teardown, the queued prompts point at a team whose state dir is gone — those prompts should never have been emitted.

3. **`queuePaneInput` is unbounded** (line 167):
   ```js
   if (leaderHasActiveTask) {
       const sendResult = await queuePaneInput({ paneTarget: tmuxTarget, prompt });
       ...
   }
   ```
   No check against `paneGuard.paneCapture` to see whether an identical `[OMX] <worker> native Stop allowed` line is already pending for the same team. No cap on queued Stop nudges.

4. **Caller in `codex-native-hook.js` (line ~2783)** routes any `allowed` decision into `maybeNudgeLeaderForAllowedWorkerStop` without consulting team-level state (e.g. a `shutdown_started_at` marker), and swallows errors silently.

## Suggested fixes (any of these substantially mitigates)

- **Team-scoped cooldown ledger.** Add `state/team/<team>/worker-stop-nudge.json` checked in addition to the per-worker file. Suppress when *any* worker has nudged within `OMX_TEAM_WORKER_STOP_COOLDOWN_MS` (default 30s).
- **Shutdown gate.** Before building the prompt, check for a `state/team/<team>/shutdown.json` (or absence of `manifest.v2.json`). If present/absent, log `result: 'team_shut_down'` and skip the inject. Set this marker as the first step of team teardown.
- **Inject dedupe against pane capture.** `evaluatePaneInjectionReadiness` already returns `paneGuard.paneCapture`. Skip when the capture's tail already contains `[OMX] <worker> native Stop allowed` (or any same-team Stop nudge within the last N lines).
- **Cap queued Stop nudges per team to 1.** When `leaderHasActiveTask` and a same-team nudge is already pending in the queue, replace or skip rather than append.

## Why this is annoying in practice

The leader (Claude Code / Claude in this case) has to consume each queued message as a separate turn. With 10+ identical nudges pointing at a dead team, that's 10+ wasted token-billable turns where the only correct response is "No team state found, nothing to do." Users notice and assume Claude/OMX is stuck in a loop.

## Environment

- `oh-my-codex@0.18.5` installed via `npm i -g`
- Node `v24.12.0`
- Linux 6.17.0-29-generic
- Launcher: `omx --madmax`
- Run dir: `~/.omx-runs/run-20260527054610-5468`
- Workers stopped natively (codex Stop hook) while team teardown was concurrent.

## References

- `dist/scripts/notify-hook/team-worker-stop.js` lines 99–236 (function body)
- `dist/scripts/codex-native-hook.js` lines 2779–2795 (caller, swallows errors)
- `dist/scripts/codex-native-hook.js` lines 1559–1652 (`resolveTeamWorkerStopDecision` — no shutdown awareness)

Yeachan-Heo:
## Summary
- Deduplicate allowed worker Stop leader nudges at the team level.
- Add a short team lock so concurrent workers cannot send duplicate Stop nudges.
- Avoid recreating removed team state during teardown races.

## Verification
- `npm run build`
- `node --test --test-name-pattern 'worker Stop|Stop leader nudge|allowed worker Stop' dist/scripts/__tests__/codex-native-hook.test.js`
- `npm run lint`
- `git diff --check`
- `npm run check:no-unused`

Fixes #2559

—

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -33,6 +33,7 @@ import { getLegacyWikiDir, serializePage, writePage } from "../../wiki/storage.j
 import { WIKI_SCHEMA_VERSION } from "../../wiki/types.js";
 import { createUltragoalPlan, readUltragoalPlan } from "../../ultragoal/artifacts.js";
 import { getBaseStateDir } from "../../state/paths.js";
+import { maybeNudgeLeaderForAllowedWorkerStop } from "../notify-hook/team-worker-stop.js";
 
 function nativeHookScriptPath(): string {
   return join(process.cwd(), "dist", "scripts", "codex-native-hook.js");
@@ -136,8 +137,12 @@ async function withLoreGuardConfig<T>(
 
 function buildWorkerStopFakeTmux(
   tmuxLogPath: string,
-  options: { failSend?: boolean; busyLeader?: boolean } = {},
+  options: { failSend?: boolean; busyLeader?: boolean; captureText?: string; sendDelayMs?: number; removePathOnSend?: string } = {},
 ): string {
+  const rawCaptureText = options.captureText ?? (options.busyLeader ? "• Working… (esc to interrupt)" : "› ready");
+  const captureText = `'${rawCaptureText.replace(/'/g, "'\"'\"'")}'`;
+  const sendDelaySeconds = Math.max(0, options.sendDelayMs ?? 0) / 1000;
+  const removePathOnSend = options.removePathOnSend ? `'${options.removePathOnSend.replace(/'/g, "'\"'\"'")}'` : "";
   return `#!/usr/bin/env bash
 set -eu
 echo "$@" >> "${tmuxLogPath}"
@@ -165,10 +170,12 @@ if [[ "$cmd" == "display-message" ]]; then
   exit 0
 fi
 if [[ "$cmd" == "capture-pane" ]]; then
-  ${options.busyLeader ? 'echo "• Working… (esc to interrupt)"' : 'echo "› ready"'}
+  printf '%s\\n' ${captureText}
   exit 0
 fi
 if [[ "$cmd" == "send-keys" ]]; then
+  ${sendDelaySeconds > 0 ? `sleep ${sendDelaySeconds}` : ""}
+  ${removePathOnSend ? `rm -rf ${removePathOnSend}` : ""}
   ${options.failSend ? "exit 1" : "exit 0"}
 fi
 exit 0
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
MERGE_READY

OMX review gate completed for PR #2574.

Evidence:
- Changed files match the expected narrow scope only:
  - `src/scripts/notify-hook/team-worker-stop.ts`
- The fix adds team-scoped worker Stop nudge dedupe/state, lock-based concurrent suppression, shutdown/terminal-team state gating, and visible queued-nudge duplicate suppression while preserving the per-worker cooldown path.
- Regression coverage covers same-team multi-worker dedupe, concurrent dedupe, missing/shutdown skip, queued duplicate suppression, and teardown-race behavior.
- PR is open, not draft, mergeState is `CLEAN`, and live CI rollup is green for lint/typecheck/build/test/coverage/status checks.
- No user-facing contract change requiring owner confirmation.

—

Yeachan-Heo:
REQUEST_CHANGES

I reviewed the exact PR head `a7a6ed45b2442e3c2897342f298afeeb2546b0ca` against latest `dev` `1b808672a2f0b5ede51ce65953265e01bb743eb2` for the duplicate worker Stop nudge fix.

What I verified:
- Scope is narrow: `src/scripts/notify-hook/team-worker-stop.ts` and `src/scripts/__tests__/codex-native-hook.test.ts` only.
- Live CI required gate was green when checked (`CI Status` passed), and the PR head/base matched the expected refs.
- Focused local checks on the exact PR head passed after dependency linkage:
  - `npm run build`
  - `node --test --test-name-pattern 'worker Stop|Stop leader nudge|allowed worker Stop' dist/scripts/__tests__/codex-native-hook.test.js` (11/11)
  - `npm run lint -- src/scripts/notify-hook/team-worker-stop.ts src/scripts/__tests__/codex-native-hook.test.ts`
  - `npm run check:no-unused`
  - `git diff --check origin/dev...HEAD`

Blocking findings before merge:
1. `teamStopNudgeAlreadyQueued()` treats captured pane transcript as authoritative pending-queue state (`src/scripts/notify-hook/team-worker-stop.ts`). An old already-processed nudge that remains visible in pane history can suppress a later legitimate worker Stop nudge after cooldown, with no retry path.
2. Successful send/queue delivery can be reported even if worker/team cooldown persistence fails. The state writes are swallowed, so duplicate nudges can recur when the durable team cooldown was not actually recorded.
3. Deferred paths still use recursive state/event writes and can recreate torn-down team state if teardown races between the precheck and deferred recording.

Please adjust the implementation so duplicate suppression is based on durable/explicit pending state rather than generic transcript text, preserve persistence failures except for known teardown cases, and make deferred recording teardown-aware. Add regression coverage for the stale-visible-transcript-after-cooldown case and persistence/deferred teardown behavior.

No merge performed.

—

PR #2615, opened 2026-05-29
Guard hook payload intake size

Yeachan-Heo:
## Summary
- add a shared bounded raw hook payload guard for native stdin and notify argv payloads
- fail closed for oversized native non-Stop payloads while allowing oversized Stop payloads to return `{}` without full JSON parsing
- skip inactive native Stop processing when there is no OMX/git/runtime surface, avoiding `.omx/state` creation
- add regressions for oversized native Stop/UserPromptSubmit stdin and oversized notify-hook argv payloads

Fixes #2614.

## Validation
- `npm run build`
- `node --test --test-concurrency=1 dist/scripts/__tests__/codex-native-hook.test.js dist/scripts/notify-hook/__tests__/payload-guard.test.js dist/hooks/__tests__/notify-hook-modules.test.js`
- `npm run test:recent-bug-regressions:compiled`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

Issue #2614, opened 2026-05-29
Guard native and notify hooks against oversized long-session payloads


physicstellyou1928:
## Problem

In very long Codex sessions, hook payloads can become large enough to cause short CPU spikes in the hook process before the hook can decide whether it needs to do any real work.

Observed locally on Linux/Codex App with a very large conversation (`~96.1M in / 375K out`):

- `node .../oh-my-codex/dist/scripts/codex-native-hook.js` briefly consumed `100%+` CPU and then exited.
- A prior related case showed `notify-hook.js` consuming high CPU when handed a large payload/transcript-shaped input.

The likely failure mode is unbounded hook input handling:

- `notify-hook.js` receives JSON through argv and previously parsed it before applying any size guard.
- `codex-native-hook.js` receives JSON through stdin and reads/concatenates/parses the full payload before it can tell that an oversized `Stop` event can be a no-op.

## Local fix prepared

I prepared and tested a local branch named:

`fix/notify-hook-large-payload`

Local commit:

`b73cbd9 Bound hook payload intake to avoid long-session stalls`

I could not push the branch from this machine because the local `origin` remote is HTTPS and has no available GitHub credentials (`could not read Username for 'https://github.com'`). The branch is ready locally.

## Proposed changes in the local commit

- Add a bounded notify-hook raw payload guard before full JSON parsing.
- Add raw JSON string-field extraction for pre-parse logging.
- Add a bounded native-hook stdin guard.
- For oversized native `Stop` payloads, identify `hook_event_name` from the prefix and return `{}` without full JSON parsing.
- Keep non-Stop oversized native hook payloads fail-closed with a deterministic diagnostic.
- Add an inactive native `Stop` fast path for directories with no `.omx`, no git layout, and no OMX runtime hints, avoiding unnecessary `.omx/state` creation.
- Preserve existing git/document-refresh/workflow Stop audits for git repositories and OMX-managed directories.
- Add regression tests for oversized notify payloads and oversized native Stop stdin.

## Verification run locally

- `npm run build`
- `node --test dist/scripts/__tests__/codex-native-hook.test.js dist/hooks/__tests__/notify-hook-large-payload.test.js dist/hooks/__tests__/notify-hook-modules.test.js`
  - 420 tests passed
- `npm run test:recent-bug-regressions:compiled`
  - 643 tests passed
- `git fsck --full`
  - passed after repairing four local zero-byte blob objects created before push

## Notes

[...]

PR #2616, opened 2026-05-29
Compact oversized notify payload summaries

iqdoctor:
## Summary
- refine the merged #2615 notify guard so oversized legacy notify payloads still leave a bounded local turn-log hint instead of disappearing completely
- replace the earlier broad compact payload view with a narrow `extractOversizedNotifyLogSummary` helper for the oversized path only
- persist only normal turn-log-equivalent fields for managed OMX directories: capped input/output previews, input count/truncation flags, turn/thread/type, and payload size
- keep the oversized path side-effect-minimal: no full `JSON.parse`, no synthetic `input_messages`, no state writes, no native/derived event dispatch, and no fallback completion inference

## Context
#2615 correctly added the raw payload guard for #2614. This follow-up keeps that safety boundary, but preserves a small amount of useful notify evidence when Codex legacy notify payloads are oversized because of historical `input_messages` or assistant text bloat.

The producer-side Codex issue is tracked at:
- https://github.com/openai/codex/issues/25141

## Why this version is smaller
After RALPLAN + Scholastic review, the patch was narrowed from a “compact payload view” abstraction to a log-summary extractor. The net simplification removed payload-view metadata that the oversized branch must not semantically own (`client`, `mode`, `source`, `project_name`, `session_id`, scan-byte metadata, fallback-complete inference, full arrays).

## Validation
- `npm run build`
- `node --test --test-concurrency=1 dist/scripts/notify-hook/__tests__/payload-guard.test.js`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

### DIFF
--- a/src/scripts/hook-payload-guard.ts
+++ b/src/scripts/hook-payload-guard.ts
@@ -111,3 +111,214 @@ export function extractRawCodexHookEventName(rawInput: string): RawCodexHookEven
     ? raw as RawCodexHookEventName
     : null;
 }
+
+export const MAX_NOTIFY_COMPACT_SCAN_BYTES = 2 * 1024 * 1024;
+export const MAX_NOTIFY_RETAINED_TEXT_BYTES = 16 * 1024;
+
+export interface CompactNotifyPayloadView {
+  intake: "compact";
+  payloadCompacted: true;
+  rawPayloadBytes: number;
+  rawPayloadBytesScanned: number;
+  type: string;
+  cwd: string;
+  threadId: string;
+  turnId: string;
+  sessionId: string;
+  client: string;
+  mode: string;
+  producerSource: string;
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Closing this draft rather than keeping it open. #2615 intentionally made oversized notify argv payload intake fail closed before full parsing. This draft changes that boundary into compact extraction plus log writing for previously dropped oversized payloads, including retained input/output previews. That is a contract/privacy-sensitive behavior change on the hot hook path and is not something to accept from an external draft follow-up without a maintainer-owned design.

The safer path is to keep #2615's bounded fail-closed behavior. If compact oversized notify evidence becomes necessary, it should be redesigned as a maintainer-owned change with an explicit redaction/privacy contract and producer-side fix context.

—

PR #2617, opened 2026-05-29
Summarize oversized notify logs without payload views

iqdoctor:
## Summary
- refine the merged #2615 notify guard so oversized legacy notify payloads still leave a bounded local turn-log hint instead of disappearing completely
- replace the earlier broad compact payload view with a narrow `extractOversizedNotifyLogSummary` helper for the oversized path only
- persist only normal turn-log-equivalent fields for managed OMX directories: capped input/output previews, input count/truncation flags, turn/thread/type, and payload size
- keep the oversized path side-effect-minimal: no full `JSON.parse`, no synthetic `input_messages`, no state writes, no native/derived event dispatch, and no fallback completion inference

## Context
#2615 correctly added the raw payload guard for #2614. This follow-up keeps that safety boundary, but preserves a small amount of useful notify evidence when Codex legacy notify payloads are oversized because of historical `input_messages` or assistant text bloat.

The producer-side Codex issue is tracked at:
- https://github.com/openai/codex/issues/25141

## Why this version is smaller
After RALPLAN + Scholastic review, the patch was narrowed from a “compact payload view” abstraction to a log-summary extractor. The net simplification removed payload-view metadata that the oversized branch must not semantically own (`client`, `mode`, `source`, `project_name`, `session_id`, scan-byte metadata, fallback-complete inference, full arrays).

## Validation
- `npm run build`
- `node --test --test-concurrency=1 dist/scripts/notify-hook/__tests__/payload-guard.test.js`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

### DIFF
--- a/src/scripts/hook-payload-guard.ts
+++ b/src/scripts/hook-payload-guard.ts
@@ -111,3 +111,177 @@ export function extractRawCodexHookEventName(rawInput: string): RawCodexHookEven
     ? raw as RawCodexHookEventName
     : null;
 }
+
+export const MAX_NOTIFY_COMPACT_SCAN_BYTES = 2 * 1024 * 1024;
+export const MAX_NOTIFY_RETAINED_TEXT_BYTES = 16 * 1024;
+
+export interface OversizedNotifyLogSummary {
+  cwd: string;
+  type: string;
+  threadId: string;
+  turnId: string;
+  latestInputText: string;
+  inputMessageCount: number | null;
+  inputMessageCountPartial: boolean;
+  inputMessagesTruncated: boolean;
+  outputPreview: string;
+  rawPayloadBytes: number;
+}
+
+function truncateUtf8(value: string, maxBytes: number): string {
+  if (utf8ByteLength(value) <= maxBytes) return value;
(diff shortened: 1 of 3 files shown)


iqdoctor:
Scholastic gate: PASS.

Ontology check: this patch now treats the oversized legacy notify data as a **log summary source**, not as a recoverable canonical hook payload. That avoids the earlier category mistake of giving the oversized path payload-level semantics (`mode`, `source`, fallback completion, synthetic `input_messages`, etc.).

Logic check: the branch remains strictly side-effect-minimal: bounded scan only, managed-cwd gate, capped previews, no state writes, no native event dispatch, and no full JSON parse. This preserves #2615's safety boundary while restoring only low-sensitivity local turn-log observability.

Yeachan-Heo:
Closing this for the same maintainership boundary as #2616. The patch is narrower, but it still changes the #2615 oversized notify argv path from fail-closed/drop to bounded extraction plus `.omx/logs` writes, including `input_preview` / `output_preview` derived from an oversized payload. That is a privacy/contract-sensitive behavior change on a hot hook path, not a safe external follow-up to merge.

Keeping the fail-closed boundary is intentional. If we later want oversized notify observability, it needs a maintainer-owned design with explicit redaction semantics and producer-side context, not a retry PR that reopens payload-derived logging.

—

PR #2619, opened 2026-05-29
Clarify native ralplan tracker diagnostics

iqdoctor:
## Summary
- make strict Autopilot ralplan consensus gate failures more actionable when native review evidence is not present in `subagent-tracking.json`
- include the actual tracker path checked in missing-session/thread diagnostics
- clarify that external/collab subagent reviews are not tracker-backed OMX native lanes

## Investigation
This came from a real Autopilot run where Architect/Critic review content existed, but `omx state write` correctly refused the ralplan -> ultragoal/complete transition. The reviews were produced by the external collab subagent tool, not by OMX-tracked native subagent lanes, so they were not recorded under the current session in `.omx/state/subagent-tracking.json`.

The gate behavior is intentionally strict and should stay strict. The problem was the operator-facing diagnosis: the error only said the tracker session/thread was missing, without showing which tracker was checked or explaining that non-OMX/collab reviews do not satisfy the native-lane requirement.

## Validation
- `npm run build`
- `env -u OMX_ROOT -u OMX_STATE_ROOT -u OMX_TEAM_STATE_ROOT -u OMX_SESSION_ID node --test --test-concurrency=1 dist/state/__tests__/operations.test.js --test-name-pattern="native ralplan reviews are not present"`
- `node --test --test-concurrency=1 dist/ralplan/__tests__/runtime.test.js --test-name-pattern="consensus|native"`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

PR #2620, opened 2026-05-29
Reset terminal autopilot state on reactivation

iqdoctor:
## Summary
- reset Autopilot mode state when `$autopilot` is invoked again after a terminal phase
- avoid preserving stale `current_phase: complete` / completed handoff artifacts during a new activation
- add a regression test for reactivating Autopilot after terminal state with stale top-level and nested handoffs

## Investigation
Disk evidence from the real run showed two separate issues:

1. `omx state write` correctly refused my synthetic ralplan consensus because the reviews were not present in OMX `subagent-tracking.json` for the current session. That is addressed separately by #2619 with clearer diagnostics.
2. A later `$autopilot` activation reused a terminal `autopilot-state.json`: it became `active: true` while still carrying `current_phase: complete` and stale handoff evidence. That is a real reactivation bug fixed here.

## Validation
- `npm run build`
- `node --test --test-concurrency=1 dist/hooks/__tests__/keyword-detector.test.js --test-name-pattern="resets Autopilot mode state"`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

### DIFF
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -647,6 +647,84 @@ describe('keyword detector skill-active-state lifecycle', () => {
     }
   });
 
+
+  it('resets Autopilot mode state when reactivated after a terminal phase', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-autopilot-terminal-reset-'));
+    try {
+      const stateDir = join(cwd, '.omx', 'state');
+      const sessionId = 'sess-autopilot-terminal-reset';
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
Thanks for the focused fix, but this is not merge-ready. The reactivation reset is incomplete: it only treats `complete`/`completed`/`failed`/`cancelled`/`canceled` as terminal, while stopped/user-stopped Autopilot terminal state can also carry stale handoff artifacts and should be reset under the stated scope.

Because this is an external PR that needs design/code changes, I am closing it rather than leaving it in request-changes limbo. A fresh follow-up can reopen with the terminal-state predicate covering stopped/user-stopped cases and regression coverage for that path.

—

PR #2622, opened 2026-05-29
Surface HUD authority tick failures

iqdoctor:
## Summary
- surface HUD authority tick child watcher failures with status/signal plus stderr/stdout details
- print notify-fallback watcher missing-script and fatal errors to stderr before exiting
- add regressions for stderr, silent nonzero, missing notify-script, and fatal authority-only paths

## Validation
- npm run build
- node --test dist/hud/__tests__/authority.test.js dist/hud/__tests__/index.test.js dist/hooks/__tests__/notify-fallback-watcher.test.js --test-name-pattern='authority|notify script missing|fatal watcher|keeps rendering when the authority tick fails'
- npm run lint
- npm run check:no-unused
- git diff --check

## Notes
- HUD watch still keeps rendering after authority tick failures; the change makes failures actionable instead of generic `status 1` bling.

### DIFF
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -647,6 +647,173 @@ describe('keyword detector skill-active-state lifecycle', () => {
     }
   });
 
+
+  it('fully resets terminal Autopilot mode state when reactivated', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-autopilot-terminal-reset-'));
+    const stateDir = join(cwd, '.omx', 'state');
+    const sessionId = 'sess-autopilot-terminal-reset';
+    try {
+      await mkdir(join(stateDir, 'sessions', sessionId), { recursive: true });
+      await writeFile(join(stateDir, 'sessions', sessionId, SKILL_ACTIVE_STATE_FILE), JSON.stringify({
+        version: 1,
+        active: true,
+        skill: 'autopilot',
+        keyword: '$autopilot',
+        phase: 'complete',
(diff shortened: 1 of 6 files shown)


iqdoctor:
Root cause: the HUD authority tick was failing in a child `notify-fallback-watcher`, but `runHudAuthorityTick` spawned that child with stdio ignored. When the child exited with status 1, stderr/stdout were discarded, so the HUD could only show the generic message `HUD watch authority tick failed: hud authority tick failed with status 1`.

So the HUD itself was not crashing; it intentionally kept rendering after authority tick failures. The bug was that the real child failure cause was hidden, making the message look like unexplained blinking noise.

This PR makes those failures actionable by piping child stdout/stderr, including status/signal in the authority error, and making the watcher print missing-script/fatal causes to stderr before `exit(1)`. Regression tests cover stderr failures, silent nonzero exits, missing notify-script, and fatal authority-only watcher paths.

Yeachan-Heo (approved):
Reviewed via `omx-pr-2622-hud-tick-review` and follow-up CI check. Diff scope is limited to HUD authority tick failure surfacing and notify fallback watcher reporting. Focused isolated validation passed, and CI is now green. No keyword-detector reintroduction was found; those files were base drift only.

Evidence:
- PR diff: `src/hud/authority.ts`, `src/hud/__tests__/authority.test.ts`, `src/scripts/notify-fallback-watcher.ts`, `src/hooks/__tests__/notify-fallback-watcher.test.ts`
- Local isolated worktree `/tmp/omx-pr2622-review`: build passed; focused HUD/notify tests passed
- CI: all required lanes pass/skip appropriately

Approved.

—

PR #2624, opened 2026-05-29
Fix HUD autopilot session mirror fallback

iqdoctor:
## Summary
- restore HUD visibility for active autopilot when canonical skill-active state is mirrored at root but the session skill-active file is not materialized yet
- allow HUD to render canonical workflow phase from skill-active state without resurrecting terminal mode detail
- add regression coverage for session-mirrored root autopilot state

## Validation
- `npm run build`
- `node dist/scripts/run-test-files.js dist/hud/__tests__/state.test.js`
- `node dist/scripts/run-test-files.js dist/hooks/__tests__/keyword-detector.test.js`
- `git diff --check`

PR #2625, opened 2026-05-29
Fix HUD autopilot session mirror fallback

iqdoctor:
## Summary
- restore HUD visibility for active autopilot when canonical skill-active state is mirrored at root but the session skill-active file is not materialized yet
- allow HUD to render canonical workflow phase from skill-active state without resurrecting terminal mode detail
- add regression coverage for session-mirrored root autopilot state

This is a minimal branch recreated from current `upstream/dev` after #2624 was closed for targeting `main` and carrying unrelated branch delta.

## Validation
- `npm run build`
- `node dist/scripts/run-test-files.js dist/hud/__tests__/state.test.js dist/hooks/__tests__/keyword-detector.test.js`
- `git diff --check`

### DIFF
--- a/src/hud/__tests__/state.test.ts
+++ b/src/hud/__tests__/state.test.ts
@@ -751,6 +751,36 @@ describe('readAllState canonical skill precedence', () => {
     });
   });
 
+  it('surfaces session-mirrored root autopilot state when the HUD session file has not been materialized yet', async () => {
+    await withTempRepo('omx-hud-root-mirror-autopilot-', async (cwd) => {
+      const rootStateDir = join(cwd, '.omx', 'state');
+      const sessionId = 'sess-autopilot-root-mirror';
+      const sessionDir = join(rootStateDir, 'sessions', sessionId);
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (changes requested):
Thanks for recreating this as a minimal `dev` branch, but I am not merging this version. The happy-path HUD fallback works, but the root fallback is still too broad: when the per-session `skill-active-state.json` is absent, `readVisibleSkillActiveStateFromPaths()` can surface root `active_skills` entries without a matching `session_id` as visible for the current session. That can leak/resurrect a stale/global canonical workflow owner into a session that has not actually materialized that skill state.

Please tighten the fallback to only session-bound mirrored entries for this path, or add an explicit freshness/ownership guard that proves the root entry belongs to the current HUD session. Include a regression where root `skill-active-state.json` has an active workflow entry with no/different `session_id` and the current session has no session skill-active file; HUD must not surface that stale owner.

Closing this external PR rather than leaving it in request-changes limbo.

—

PR #2678, opened 2026-06-01
[codex] Ignore native hook routing text in keyword detection

dale-design:
## Summary
- Ignore native OMX hook routing text before running UserPromptSubmit keyword detection.
- Keep explicit user requests such as `ultraqa 하자` working normally.
- Add regression tests for the hook-text false positive and explicit ultraqa activation path.

## Root cause
Stop/UserPromptSubmit routing messages such as `OMX ultraqa is still active...` can be carried into later prompt payloads. The keyword detector treated those hook-generated lines as user-authored workflow requests, reactivating `ultraqa` even when the latest human prompt was only asking about the loop.

## Validation
- `node --test /Users/deugcheonhan/.npm-global/lib/node_modules/oh-my-codex/dist/scripts/__tests__/codex-native-hook.test.js` -> PASS (54/54) after applying the equivalent installed-package patch.
- Smoke: hook-generated `OMX ultraqa is still active...` text no longer activates `ultraqa`; explicit `ultraqa 하자` still activates it.

Note: Full source `npm run build` was not run locally because the npm-installed package copy does not include dev dependency binaries such as `tsc`; the compiled installed-package regression suite passed with the same logic.


Yeachan-Heo:
Thanks for the contribution. I’m closing this because PRs to this repo must target `dev`, not `main`. Please reopen/resubmit against `dev` if the change is still applicable after syncing with the current branch.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2680, opened 2026-06-01
fix: preserve tagged HUD panes during cleanup

### LINKED ISSUE #2679: HUD duplicate cleanup can preserve legacy pane over leader-tagged pane
## Summary

When prompt-submit HUD reconciliation sees both a legacy same-session `hud --watch` pane without `OMX_TMUX_HUD_LEADER_PANE` and a current leader-tagged HUD pane, cleanup currently preserves whichever matching pane appears first in the tmux snapshot. If tmux lists the legacy pane first, reconciliation can keep the weaker legacy owner metadata and kill the stronger leader-tagged pane.

## Impact

This makes subsequent HUD reconciliation less deterministic and can contribute to repeated HUD/footer churn in sessions that have accumulated legacy HUD panes.

## Expected behavior

When a leader pane scope is known, same-owner HUD lookup should prefer panes whose `OMX_TMUX_HUD_LEADER_PANE` matches the current leader, while still allowing legacy same-session panes as fallback cleanup candidates.

## Reproduction shape

A same-window tmux snapshot containing:

1. leader pane `%1`
2. legacy same-session HUD pane `%2` with `OMX_SESSION_ID=sess-a` but no leader tag
3. current HUD pane `%3` with `OMX_SESSION_ID=sess-a` and `OMX_TMUX_HUD_LEADER_PANE=%1`

`findHudWatchPaneIds(panes, '%1', { sessionId: 'sess-a', leaderPaneId: '%1' })` should return `%3` before `%2`.

## Proposed fix

Sort matching HUD panes so exact leader-tagged matches come before same-session legacy matches. Add a regression test for the mixed legacy/current ordering.

XIO1016:
## Summary
- Prefer exact leader-tagged HUD watch panes over legacy same-session panes during duplicate cleanup.
- Add a regression for mixed owner metadata so reconciliation preserves the stronger owner tag.
- Isolate the compat doctor contract test from ambient `USE_OMX_EXPLORE_CMD` unless a case explicitly opts in.

## Validation
- `npm run build`
- `USE_OMX_EXPLORE_CMD=1 node --test dist/compat/__tests__/doctor-contract.test.js`
- `USE_OMX_EXPLORE_CMD=1 node --test dist/hud/__tests__/tmux.test.js dist/hud/__tests__/reconcile.test.js`
- `npm run lint`
- `npm test`

Fixes #2679

### DIFF
--- a/src/compat/__tests__/doctor-contract.test.ts
+++ b/src/compat/__tests__/doctor-contract.test.ts
@@ -42,10 +42,14 @@ function resolveCompatTarget(): { command: string; argsPrefix: string[] } {
 
 function runCompatTarget(cwd: string, argv: string[], envOverrides: Record<string, string> = {}): CompatRunResult {
   const target = resolveCompatTarget();
+  const env = { ...process.env, ...envOverrides };
+  if (!Object.hasOwn(envOverrides, 'USE_OMX_EXPLORE_CMD')) {
+    delete env.USE_OMX_EXPLORE_CMD;
+  }
   const result = spawnSync(target.command, [...target.argsPrefix, ...argv], {
     cwd,
     encoding: 'utf-8',
-    env: { ...process.env, ...envOverrides },
+    env,
   });
   return { status: result.status, stdout: result.stdout || '', stderr: result.stderr || '', error: result.error?.message };
 }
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Closing this draft because the linked issue #2679 is already fixed on `dev` by #2687 (`ce11bbd6557a47a72a999751877bdcf0eb9c1990`). Keeping this open would duplicate the merged HUD cleanup path.

Thanks for surfacing the original repro shape; it is now covered in the merged follow-up.

—

PR #2694, opened 2026-06-01
Keep deep-interview questions visible in short tmux panes

iqdoctor:
## Summary
- Open a full tmux window when the estimated question frame would be clipped in the current pane
- Keep the normal split-pane path when there is enough height
- Add regression coverage for both sizing paths

## Verification
- npm run build
- env -u OMX_ROOT -u OMX_SESSION_ID node --test dist/question/__tests__/renderer.test.js
- npx biome lint src/question/renderer.ts src/question/__tests__/renderer.test.ts
- Live tmux smoke check in a short pane: question UI opened in a new window instead of a tiny split

### DIFF
--- a/src/question/__tests__/renderer.test.ts
+++ b/src/question/__tests__/renderer.test.ts
@@ -214,7 +214,8 @@ describe('launchQuestionRenderer', () => {
         strategy: 'inside-tmux',
         execTmux: (args) => {
           calls.push(args);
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for the PR. I’m closing this one because it targets `main`; OmX PRs need to target `dev` so they go through the normal integration and CI path.

If you want to continue this change, please reopen it as a clean PR against `dev`.

—

PR #2695, opened 2026-06-01
Keep deep-interview questions visible in short tmux panes

iqdoctor:
## Summary
- Open a full tmux window when the estimated question frame would be clipped in the current pane
- Keep the normal split-pane path when there is enough height
- Add regression coverage for both sizing paths

## Verification
- npm run build
- env -u OMX_ROOT -u OMX_SESSION_ID node --test dist/question/__tests__/renderer.test.js
- npx biome lint src/question/renderer.ts src/question/__tests__/renderer.test.ts
- Live tmux smoke check in a short pane: question UI opened in a new window instead of a tiny split

## Branch policy
- Target base branch: main
- branch is now based on main

### DIFF
--- a/src/question/__tests__/renderer.test.ts
+++ b/src/question/__tests__/renderer.test.ts
@@ -205,6 +205,123 @@ describe('question window topology selection', () => {
 
     assert.ok(estimateQuestionRenderFootprint(record, 20) > estimateQuestionRenderFootprint(record, 80));
   });
+
+  it('sizes multi-question records from the largest visible screen rather than summing every question', () => {
+    const longQuestion = {
+      id: 'q-2',
+      question: 'x'.repeat(220),
+      options: [{
+        label: 'Only option',
+        value: 'only',
+        description: 'y'.repeat(180),
+      }],
+      allow_other: false,
+      multi_select: false,
+      type: 'single-answerable',
+    };
+    const shortQuestion = {
+      id: 'q-1',
+      question: 'Short question?',
+      options: [{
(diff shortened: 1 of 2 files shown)


iqdoctor:
Dogfooding note: I’m dogfooding autopilot with bugs, but I do **not** dogfood deep-interview. The  path had a critical bug, and this PR fixes it.

This is now in progress, and after the full dogfooding pass I recommend cutting a minor release quickly.

iqdoctor:
Dogfooding note: I’m dogfooding autopilot with bugs, but I do **not** dogfood deep-interview. The autopilot:deep-interview path had a critical bug, and this PR fixes it.

This is now in progress, and after the full dogfooding pass I recommend cutting a minor release quickly.

iqdoctor:
Addressed the review feedback in the current branch:\n\n- Full-window decision now uses the capped split budget, not the raw pane height.\n- The new-window path now targets the return session/window, so bridge-based launches stay in the right tmux context.\n- Multi-question sizing now measures the visible wizard frame(s) and review frame, instead of summing future questions.\n- The footprint estimate is based on rendered frames, so wrapped option rows are counted at terminal width.\n\nI also marked the review threads resolved in GitHub. All review comments on this PR are now closed.

Yeachan-Heo:
Closing this PR because it targets `main`. This repository's contribution flow requires PRs to target `dev`; `main` is not an acceptable base for feature/fix PRs.

Please reopen or recreate against `dev` if the change is still needed.

—

PR #2703, opened 2026-06-01
Keep repeated tmux HUD reconciliation scoped to the emitting pane

iqdoctor:
## Summary

This tests-only follow-up hardens the HUD tmux regression around shared left/right panes. It proves that prompt-submit reconciliation stays pane-local even when the neighboring pane already has its own HUD, and that repeated reconciliation keeps resizing the current pane's HUD instead of adopting or killing the neighbor.

## What changed

- Added a deterministic left/right regression in `src/hud/__tests__/reconcile.test.ts`.
- Covered both create and repeat-reconcile paths for each pane.
- Asserted that each pane keeps its own `targetPaneId` and that the neighboring HUD is untouched.
- Kept the diff tests-only; no production HUD code changed.

## Verification

- `npx biome lint src/hud/__tests__/reconcile.test.ts`
- `npm run build`
- `env -u OMX_ROOT -u OMX_SESSION_ID node --test dist/hud/__tests__/reconcile.test.js`

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -429,6 +429,135 @@ describe('reconcileHudForPromptSubmit', () => {
     assert.equal(Object.hasOwn(created[0]?.options ?? {}, 'fullWidth'), false);
   });
 
+  it('keeps repeated left/right prompt-submit HUD recreation scoped to each pane when the neighboring pane already has a HUD', async () => {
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
MERGE_READY head=0ab02d2fc513c34aa6c54a60cce22513854517d4 evidence=one-test-file-diff;focused-hud-tests-pass;typecheck-temp-out-pass;diff-check-pass;ci-clean
—

PR #2709, opened 2026-06-02
fix: deliver omx question env via export prefix under cmux (tmux -e shim workaround)

Bongseop-Kim:
## Summary

When OMX runs **inside cmux**, the `omx question` UI pane "disappears immediately after launch" (`zsh: command not found: -e`). This PR is a **defensive compatibility workaround**: under cmux, OMX delivers env vars to the spawned question pane via a shell-quoted `export ... &&` prefix instead of tmux's `split-window -e KEY=VALUE` option. On real tmux, behavior is **unchanged** (`-e` is still used).

The root cause is in **cmux's tmux-compat layer**, not OMX — OMX uses standard tmux syntax correctly. A matching cmux-side issue should be filed against its `__tmux-compat` shim. This change only makes OMX's own `cmux omx` integration keep working until/unless cmux fixes the shim.

## Step 1 — Investigation findings

**How the question UI launches its pane** — `src/question/renderer.ts`:
- `buildQuestionUiTmuxArgs()` (was a private helper) builds the env + command argv that is appended after the tmux split/new-session flags.
- It is consumed by `launchQuestionRenderer()` for two strategies:
  - `inside-tmux`: `tmux split-window -v -l <h> [-t <target>] -P -F '#{pane_id}' -c <cwd> <commandArgs>`
  - `detached-tmux`: `tmux new-session -d -P -F '#{session_name}' -s <name> -c <cwd> <commandArgs>`

**Current (pre-fix) env passing** — `src/question/renderer.ts`:
```ts
return [
  ...(options.sessionId ? ['-e', `OMX_SESSION_ID=${options.sessionId}`] : []),
  ...(options.returnTarget ? ['-e', `OMX_QUESTION_RETURN_TARGET=${options.returnTarget}`, '-e', 'OMX_QUESTION_RETURN_TRANSPORT=tmux-send-keys'] : []),
  process.execPath,
  ...resolveQuestionUiProcessArgs(recordPath, options),
];
```

**Other tmux `-e` env-passing site (NOT changed — see Scope):** `src/cli/index.ts` `buildDetachedSessionBootstrapSteps()` (`new-session ... -c <cwd> [-e KEY=VALUE]... <detachedLeaderCmd>`). The remaining `-e` occurrences in the repo (`src/cli/index.ts:4829/5106`, `src/notifications/*`) are `node -e` / `osascript -e`, unrelated to tmux.

**cmux detection** — none existed in OMX before this PR. Added `isRunningUnderCmux()` in `src/utils/platform-command.ts`, detecting the `CMUX_SOCKET_PATH` / `CMUX_SOCKET` env vars cmux exports for every session it manages. Confirmed by inspecting the live cmux runtime: the shim at `~/.cmuxterm/.../tmux` execs `cmux __tmux-compat "$@"` and even fakes `tmux -V` → `tmux 3.4`, so version sniffing cannot detect the missing `-e` support — env detection is required.

**psmux / Windows:** untouched. The fix is gated on cmux detection and only rewrites the tmux env-passing path.

## Root cause

The cmux `tmux` shim does not implement tmux's `split-window -e KEY=VALUE` option (valid since tmux 3.0). It consumes `-c <cwd>` (turning it into `cd -- '<cwd>' &&`) but leaks the `-e` flags into the space-joined pane shell command, so the pane's shell tries to run `-e` as a command.

## Before / after (exact shell text OMX sends to the pane)

**Before (broken, under cmux):**
```
cd -- '<cwd>' && -e OMX_SESSION_ID=abc123 -e OMX_QUESTION_RETURN_TARGET=%5 -e OMX_QUESTION_RETURN_TRANSPORT=tmux-send-keys node .../omx.js question --ui --state-path <record>
# zsh: command not found: -e   → question pane exits instantly
```

**After (under cmux):**
```
cd -- '<cwd>' && export OMX_SESSION_ID='abc123' OMX_QUESTION_RETURN_TARGET='%5' OMX_QUESTION_RETURN_TRANSPORT='tmux-send-keys' && '<node>' '<omx.js>' 'question' '--ui' '--state-path' '<record>'
# pane stays up; env vars present in the pane
```

**Real tmux (unchanged):** still emits `-e OMX_SESSION_ID=... -e ... <node> <omx.js> question --ui ...`.

Verified the `export` form round-trips env values containing `=`, `%`, spaces, and embedded single quotes through both `/bin/sh` and `/bin/zsh`.

## Scope

Scoped to the **question UI** pane-spawn path (the verified failure). The team/HUD detached-leader path (`buildDetachedSessionBootstrapSteps` in `src/cli/index.ts`) also uses tmux `-e`, but it passes a single pre-composed shell command string and is a central team surface; rewriting it carries higher regression risk and is **left as a follow-up** (it can reuse the new `isRunningUnderCmux()` helper). Flagging explicitly per the focused-diff goal.

## Tests

- `src/utils/__tests__/platform-command.test.ts` — `isRunningUnderCmux` detection (CMUX_SOCKET_PATH / CMUX_SOCKET, blank-value, non-cmux).
- `src/question/__tests__/renderer.test.ts`:
  - real-tmux mode still emits `-e OMX_SESSION_ID=...` with raw (unquoted) command tokens;
  - cmux mode emits **no bare `-e`**, env via `export`, and the first executed token is the real command (`node`), never `-e`;
  - values with `=`, `%`, spaces, and an embedded single quote are single-quoted and round-trip;
  - integration via `launchQuestionRenderer` (cmux env) asserts the `split-window` call preserves `-c`/pane flags and contains no bare `-e`.

Gate: `npm run build`, `npm run lint` (692 files clean), and `npm test` all run.

Document-refresh: not-needed | compatibility workaround for cmux's tmux shim; no operator-facing CLI contract or documented behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project are documented in this file.
 
 ## [Unreleased]
+
+### Fixed
+
+- **`omx question` UI pane works under cmux** — when OMX runs inside cmux, its `tmux` binary is a shim (`~/.cmuxterm/.../tmux` -> `cmux __tmux-compat`) that does not implement tmux's `split-window -e KEY=VALUE` env option, so the `-e` flags leaked into the spawned pane's shell command (`zsh: command not found: -e`) and the question pane exited immediately. Under cmux, OMX now delivers env vars to the question pane via a shell-quoted `export ... &&` prefix instead of `-e`. Real tmux is unchanged (still uses `-e`). This is a defensive compatibility workaround; the root cause is in cmux's tmux-compat layer.
+
 ## [0.18.8] - 2026-06-01
 
 Patch release for the post-`0.18.7` runtime reliability train: HUD/session ownership under native session drift, Autopilot replay and context-snapshot hardening, plugin hook/cache correctness, Team startup/disablement safety, native subagent guidance, and release/CI evidence improvements.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project are documented in this file.
 
 ### Fixed
 
-- **`omx question` UI pane works under cmux** — when OMX runs inside cmux, its `tmux` binary is a shim (`~/.cmuxterm/.../tmux` -> `cmux __tmux-compat`) that does not implement tmux's `split-window -e KEY=VALUE` env option, so the `-e` flags leaked into the spawned pane's shell command (`zsh: command not found: -e`) and the question pane exited immediately. Under cmux, OMX now delivers env vars to the question pane via a shell-quoted `export ... &&` prefix instead of `-e`. Real tmux is unchanged (still uses `-e`). This is a defensive compatibility workaround; the root cause is in cmux's tmux-compat layer.
+- **`omx question` UI pane works under cmux** — when OMX runs inside cmux, its `tmux` binary is a shim (`~/.cmuxterm/.../tmux` -> `cmux __tmux-compat`) that does not implement tmux's `split-window -e KEY=VALUE` env option, so the `-e` flags leaked into the spawned pane's shell command (`zsh: command not found: -e`) and the question pane exited immediately. Under cmux, OMX now delivers env vars to the question pane via a shell-neutral, single-quoted `env KEY=VALUE ...` prefix on a single shell-command argument instead of `-e`. This stays correct on the cmux shim, on a real tmux that inherits cmux env vars, and across POSIX and non-POSIX (e.g. fish) pane shells. Real tmux without cmux env is unchanged (still uses `-e`). This is a defensive compatibility workaround; the root cause is in cmux's tmux-compat layer.
 
 ## [0.18.8] - 2026-06-01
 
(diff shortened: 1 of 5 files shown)


Yeachan-Heo (approved):
Latest-head re-review for `9ca456141539f348bde913f3c4de355f091f2f64`: MERGE_READY.

Evidence checked:
- Inspected full `origin/dev...HEAD` and focused `ef0aea3b..HEAD` delta. The latest delta only changes the cmux env-prefix path from `export ... &&` to a single shell-command `env KEY='VALUE' ...` form plus matching tests/changelog.
- Verified the cmux path no longer emits bare `-e`; env assignments use constant keys with single-quoted values, and command argv tokens are single-quoted inside the one shell-command argument.
- Verified `/bin/sh` and `zsh` parse the generated `env KEY='...' node ...` form correctly, including values with spaces, `%`, `=`, and embedded single quotes.
- Verified non-cmux/real tmux still uses `split-window -e KEY=VALUE` argv and stays raw/unquoted; detached/windows/inline strategies continue to use their existing env handling.
- Local checks passed: `npm ci`, `npm run build`, `npm run check:no-unused`, `npm run lint`, targeted compiled `renderer` + `platform-command` tests (`73/73`), and `git diff --check`.
- CI run `26803213019` completed successfully for this exact head.

Signed-off-by: Yeachan-Heo <54757707+Yeachan-Heo@users.noreply.github.com>

PR #2732, opened 2026-06-04
Keep malformed Stop hook stdin on Stop-safe JSON

### LINKED ISSUE #2727: Stop hook emits invalid Stop JSON shape for malformed stdin without event name
## Symptom

The native Stop hook can emit JSON that is parseable but invalid for Codex's Stop-hook response schema when stdin is malformed before `hook_event_name` can be recovered. Recent failure logs showed malformed stdin consisting only of:

```text
{
```

That path can surface as:

```text
hook returned invalid stop hook JSON output
```

## Cause

Malformed stdin with no recoverable event name currently falls back to the generic non-Stop fail-closed shape (`continue: false`). That shape is valid for non-Stop hooks but invalid for Stop hooks, where Codex expects a Stop-compatible schema such as `decision: "block"`.

## Expected behavior

- Native Stop CLI malformed-input paths should emit exactly one Stop-schema JSON object.
- Explicitly identifiable non-Stop malformed input should keep the non-Stop fail-closed schema.
- The hook should avoid foreground stderr noise in these recovery paths.

## Proposed fix

Default unknown malformed stdin to a Stop-safe `decision: "block"` payload while preserving the generic `continue: false` payload when an explicit non-Stop event name is recoverable.

## Local verification

Clean branch based on `upstream/main` passed:

- `npm run build`
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` (418 passing)
- `npm run lint`
- `npx tsc --noEmit`
- `git diff --check`

br4vesirrobin:
Closes #2727

## Summary

- default malformed native-hook stdin with no recoverable event name to Stop-safe block JSON
- preserve non-Stop fail-closed schema when an explicit non-Stop event name is recoverable
- add regression coverage for unknown malformed stdin (`{`) and explicit malformed `UserPromptSubmit`

## Verification

- `npm run build`
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` (418 passing)
- `npm run lint`
- `npx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -441,8 +441,32 @@ describe("codex native hook dispatch", () => {
     );
   });
 
-  it("emits schema-safe JSON stdout when CLI stdin is malformed", () => {
+  it("emits Stop-safe JSON stdout when CLI stdin is malformed before event detection", () => {
     const stdout = runNativeHookCli("{");
+
+    const output = parseSingleJsonStdout(stdout) as {
+      decision?: string;
+      reason?: string;
+      stopReason?: string;
+      systemMessage?: string;
+      hookSpecificOutput?: unknown;
+    };
+
+    assert.equal(output.decision, "block");
+    assert.equal(
+      output.reason,
+      "OMX native hook received malformed JSON input. Preserve runtime state, inspect the emitting hook payload yourself, and retry with valid JSON.",
+    );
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
Thanks for moving this to dev. I’m closing this PR because the same issue is already being handled on the maintainer branch in #2731, which has been rebased onto the current dev branch and is going through the merge path now.

—

Issue #2763, opened 2026-06-09
Release validation: v0.18.11 is not present on npm


simongonzalezdc:
## Summary

Small release-validation note. GitHub release `v0.18.11` is published, but npm still appears to expose `0.18.10` as latest and `oh-my-codex@0.18.11` is not present.

I am not suggesting a code change here. This looks like a publication/dist-tag proof check.

## Evidence

- GitHub release: https://github.com/Yeachan-Heo/oh-my-codex/releases/tag/v0.18.11
- `npm view oh-my-codex version` -> `0.18.10`
- `npm view oh-my-codex dist-tags --json` -> `{ "latest": "0.18.10" }`
- `npm view oh-my-codex@0.18.11 version` -> no matching version found
- Release readiness at the tag still lists GitHub release proof and npm proof as pending: https://github.com/Yeachan-Heo/oh-my-codex/blob/v0.18.11/docs/qa/release-readiness-0.18.11.md#ci--publication-evidence

## Expected

Either npm has `0.18.11`, or the GitHub release/readiness evidence reflects the current publication state.

## Scope

Checked from public GitHub and npm surfaces only. No code change requested.

Yeachan-Heo:
Confirmed. The GitHub release/tag exists, but `oh-my-codex@0.18.11` is not on npm yet.

The Release workflow reached the npm publish step and failed while creating the Sigstore/Rekor transparency log entry:

- `TLOG_CREATE_ENTRY_ERROR`
- `request to https://rekor.sigstore.dev/api/v1/log/entries failed, reason: read ECONNRESET`

So this looks like a publish/provenance retry gap rather than a code/build failure. Keeping this open until the npm publication proof is green.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Yeachan-Heo:
Published and verified.

- `npm view oh-my-codex version` => `0.18.11`
- `npm view oh-my-codex@0.18.11 version` => `0.18.11`
- Manual publish workflow: https://github.com/Yeachan-Heo/oh-my-codex/actions/runs/27188963990

Closing this release validation issue as fixed.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2764, opened 2026-06-09
ci(release): retry npm publish on Rekor outage

simongonzalezdc:
## Summary

Related to #2763. Bellman confirmed the `v0.18.11` release failed during npm provenance publication with `TLOG_CREATE_ENTRY_ERROR` / Rekor `ECONNRESET`, after the release build/native/smoke/package gates had already passed. This PR keeps provenance as the primary publish path and adds a narrow Rekor/TLOG fallback inline in the existing release workflow.

## Changes

- Wrap npm publish in an explicit `Publish to npm` bash step.
- Skip cleanly if the package version is already visible on npm during a rerun.
- Publish with provenance first.
- Retry `npm publish --access public` only when the captured publish log matches a Sigstore/Rekor transparency-log outage.
- Preserve non-Rekor publish failures as hard failures.

## Validation

- [x] `npm run lint`
- [x] `npm run build`
- [ ] `npm test` attempted locally; full suite currently exits 1 in existing Git local-exclude setup tests outside this workflow-only diff. PR CI test gates are green.
- [x] `git diff --check`
- [x] `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); puts "YAML OK"'`
- [x] `sed -n '327,341p' .github/workflows/release.yml | sed 's/^          //' | bash -n`
- [x] `omx doctor` not applicable; this does not change setup/config behavior.

## Checklist

- [x] Docs updated when needed; not needed for this release workflow-only change

Document-refresh: not-needed | release workflow-only change; no runtime, prompt, CLI, or product-doc behavior changed.

## Scope note

This hardens future tag-triggered release publishes. It does not by itself make the already-created `v0.18.11` npm proof green; that still needs maintainer-controlled release action such as rerunning from an updated ref/tag or publishing manually.

## Related

Related to #2763

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -320,6 +320,24 @@ jobs:
           console.log(`[version-sync] OK package=${pkg.version} workspace=${workspace.workspace?.package?.version} tag=${tag ?? ''}`.trim());
           EOF
       - run: npm pack --dry-run
-      - run: npm publish --access public --provenance
+      - name: Publish to npm
+        shell: bash
+        run: |
+          PACKAGE_NAME=$(node -p "require('./package.json').name")
+          VERSION=$(node -p "require('./package.json').version")
+          if npm view "${PACKAGE_NAME}@${VERSION}" version >/dev/null 2>&1; then
+            echo "::warning::Version ${VERSION} already published to npm, skipping publish"
+            exit 0
+          fi
+
+          set -o pipefail
+          if npm publish --access public --provenance 2>&1 | tee npm-publish.log; then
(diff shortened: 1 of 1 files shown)


Yeachan-Heo:
Closing this draft workaround because the release gap has been resolved by a maintainer-controlled manual publish workflow and `oh-my-codex@0.18.11` is now live on npm.

We are not taking this PR as-is because changing the default release workflow to fall back from provenance publishing needs a separate maintainer-owned release policy decision, not a one-off external workaround.

Thanks for surfacing the release failure in #2763.

—

PR #2800, opened 2026-06-11
fix(best-practice-research): enforce terminal read-only boundary

### LINKED ISSUE #2799: $best-practice-research can continue into implementation instead of stopping at research handoff
## Summary

`$best-practice-research` appears intended to be a bounded research / planning-input workflow, but in practice an agent can continue into implementation after producing the research recommendation.

This creates a workflow-boundary mismatch: the skill says to gather repo-local facts, gather official/upstream evidence, synthesize a recommendation / handoff, and stop. It should route upward if the task becomes implementation.

## Observed behavior

In a Codex/OMX session, I invoked `$best-practice-research` for a repo-local infrastructure task that needed current upstream guidance. The agent correctly inspected repo context and gathered external evidence, but then proceeded to create implementation files and modify GitOps/Talos repo state instead of stopping at the research handoff.

No live cluster/host mutation occurred, but local repo changes were created even though the invoked workflow was research-only.

## Expected behavior

After `$best-practice-research` has enough evidence to support a recommendation, it should stop with the documented output contract:

- Direct recommendation
- Evidence used
- Version/date context
- Repo-local context
- Boundaries / non-goals
- Handoff

If the next step is implementation, the workflow should explicitly hand off to `$ralplan`, `$ralph`, `$team`, or normal execution, rather than silently continuing.

## Why this matters

Users may invoke `$best-practice-research` specifically because they want planning input before approving implementation. If the agent starts editing files, the skill behaves more like an execution workflow than a research wrapper.

This is especially risky for infrastructure / ops repos where the distinction between:

- Git-tracked planning artifacts,
- local draft edits,
- GitOps-managed Kubernetes desired state, and
- manually applied host changes

is important.

## Suggested fix

Consider adding a stronger enforcement point around the skill stop rule, for example:

1. Make the skill output contract terminal by default.
2. Add explicit language such as: “Do not write files, create commits, or modify repository state under this skill.”
3. Require an explicit handoff phrase before implementation, e.g. “Implementation is out of scope for `$best-practice-research`; invoke `$ralplan`, `$ralph`, `$team`, or normal execution to continue.”
4. If hooks can detect active skill state, block write tools while `$best-practice-research` is active unless the user explicitly switches workflows.

## Acceptance criteria

- Invoking `$best-practice-research` for a task with implementation implications produces a cited recommendation and stops.
- The agent does not modify repo files under this skill unless the user explicitly switches to an implementation workflow.
- The final handoff clearly names the appropriate next workflow for planning or execution.

Yeachan-Heo:
## Summary

`$best-practice-research` is intended as bounded research / planning input, but its contract did not stop it from continuing into implementation after producing recommendations. This change makes the skill **terminal and read-only by default**.

### What changed
- Added a **Terminal By Default** section to the skill contract: the skill gathers evidence, produces a cited recommendation plus handoff, then stops. It must not write/edit files, create or amend commits, run mutating commands, or modify repository state under this skill — even when the question has clear implementation implications.
- Strengthened **Stop Rules** with an explicit "this skill never implements" rule; work resumes only when the user explicitly switches to a named planning/implementation workflow.
- Updated the **Handoff** output contract to name the appropriate next workflow: `$ralplan` for planning; `$ultragoal`/`$team`/`executor` for execution.
- Mirrored the change to the plugin copy (`plugins/oh-my-codex/skills/best-practice-research/SKILL.md`), kept byte-identical to canonical.
- Added regression tests in `best-practice-research-skill.test.ts` covering the terminal/no-mutation contract and the named-workflow handoff.

### Acceptance criteria
- Invoking `$best-practice-research` for a task with implementation implications produces a cited recommendation and stops.
- The agent does not modify repo files under this skill unless the user explicitly switches workflows.
- The final handoff clearly names the appropriate next workflow for planning/execution.

### Verification
- `npm run build` (clean)
- Targeted tests pass: `best-practice-research-skill.test.js`, `research-workflow-boundaries.test.js`
- Plugin mirror verified in sync (`npm run sync:plugin:check`)

Closes #2799

—

### DIFF
--- a/plugins/oh-my-codex/skills/best-practice-research/SKILL.md
+++ b/plugins/oh-my-codex/skills/best-practice-research/SKILL.md
@@ -12,6 +12,10 @@ Use this skill when a task depends on current external best practices, version-a
 
 Produce a cited, reusable best-practice answer or handoff that separates current external evidence from repo-local facts and dependency-selection decisions. For pre-planning investigation, this is the ordinary first research wrapper: gather official/upstream evidence, then hand it to `$ralplan` or the caller as planning input. Do not present `$best-practice-research` as a final architecture component or as a validator-gated research loop.
 
+## Terminal By Default
+
+This skill is terminal and read-only by default. It gathers evidence and produces a cited recommendation with a handoff, then stops. Do not write or edit files, create or amend commits, run mutating commands, or otherwise modify repository state under this skill — even when the question has clear implementation implications. When implementation is warranted, stop and hand off rather than continuing: name `$ralplan` for planning and `$ultragoal`, `$team`, or `executor` for execution, and resume only after the user explicitly switches to that workflow.
+
 ## Activate When
 
 - The user asks for best practices, recommended approach, current guidance, official recommendations, standards, or version-aware external behavior.
@@ -71,7 +75,7 @@ Produce a cited, reusable best-practice answer or handoff that separates current
 <what this research does not decide>
 
 ### Handoff
-<planning/execution/test implications>
+<planning/execution/test implications; name the next workflow — `$ralplan` for planning, `$ultragoal`/`$team`/`executor` for execution — and note that this skill stops here unless the user explicitly switches workflows>
 ```
 
 ## Stop Rules
@@ -79,5 +83,6 @@ Produce a cited, reusable best-practice answer or handoff that separates current
 - Stop after a source-backed recommendation is reusable by the caller.
 - Stop and route upward if the task becomes dependency comparison, broad architecture, or implementation.
 - Do not continue researching when remaining work would only polish wording rather than change the recommendation.
+- This skill never implements. After delivering the recommendation and handoff, stop; do not modify repo files or repo state. Resume only when the user explicitly switches to a planning or implementation workflow named in the handoff.
 
 Task: {{ARGUMENTS}}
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Review complete — PR #2800

Scope is clean: exactly the canonical `$best-practice-research` skill, plugin mirror copy, and regression test file changed. The new Terminal By Default / Stop Rules language directly addresses #2799 by making the workflow terminal and read-only by default, and the handoff now names the appropriate planning/execution workflows instead of silently continuing into implementation.

Verification checked:
- local targeted test in review worktree: `src/hooks/__tests__/best-practice-research-skill.test.ts` passed
- canonical/plugin mirror relationship verified by test coverage
- all GitHub CI checks are green
- PR targets `dev`, is mergeable, and has no code-change blockers

Verdict: technically merge-ready, but **OWNER_CONFIRMATION_REQUIRED** before merge because this changes a user-facing skill contract / workflow boundary. No code changes requested.

—

iqdoctor:
Follow-up note after merge: this PR improves the direct `$best-practice-research` false-positive class by making the skill terminal/read-only instead of drifting into implementation.

I also agree with the remaining Codex P2 nuance: when `$best-practice-research` is invoked as an evidence lane by an already-active caller workflow (`$ralplan`, `$deep-interview`, `$team`, etc.), it should remain read-only but return evidence to that caller rather than requiring a fresh user workflow switch. I opened follow-up bead `omx-i39.1` under the same Autopilot/Ralplan guard-correctness epic and related it to the false-positive guard tickets, so we can continue that refinement without blocking this merged PR.

Issue #2597, opened 2026-05-28
[Bug] OMX detached tmux sessions retain/grow huge pane history after interrupted/stuck Codex TUI, causing multi-GB tmux memory footprin


zzNaLogiczz:
## Summary

I observed very high memory usage in a cmux/terminal environment. After investigation, `cmux` itself was not the main memory owner. The memory was held by the `tmux` server used by OMX-managed detached Codex sessions.

<img width="346" height="27" alt="Image" src="https://github.com/user-attachments/assets/a44f0edd-d46f-411a-bd47-3425e5aa1710" />

Two old OMX detached sessions had abnormal tmux pane history sizes:

- one pane had `history_bytes` around **67 GB**
- another pane had `history_bytes` around **8.8 GB**
- both panes had only `hist_lines=1`
- both values were still increasing over time, even though these sessions were old and detached

This appears related to OMX-managed detached tmux sessions where Codex TUI was interrupted or stuck but remained alive and continued writing terminal output/status updates.

## Environment

```text
oh-my-codex: v0.18.6
codex-cli: 0.134.0
tmux: 3.6a
OS: macOS / darwin arm64
Node.js: v26.0.0
Launch command: omx --madmax --xhigh
Launch policy: default / detached tmux
```

## Actual behavior

The `tmux` server had a multi-GB footprint:

```bash
tmux_pid=$(tmux display-message -p '#{pid}')
footprint -summary -pid "$tmux_pid"
```

Observed:

```text
tmux [62427]: Footprint: 7281 MB
phys_footprint_peak: 25 GB
MALLOC_REALLOC: 7211 MB
```

`vmmap` also showed:

```text
Physical footprint:         7.1G
Physical footprint (peak):  25.4G
Writable regions: ... swapped_out=7.1G
MALLOC_REALLOC:             ~7.0G
```

For comparison, `cmux` itself was much smaller:

```text
cmux footprint: 372 MB
cmux peak:      681 MB
```

## Abnormal pane history

Command used:

```bash
tmux list-panes -a -F '#{session_name}:#{window_index}.#{pane_index} pane=#{pane_id} pid=#{pane_pid} hist_lines=#{history_size} hist_bytes=#{history_bytes} hist_all=#{history_all_bytes} cmd=#{pane_current_command}' \
  | sort -t= -k5,5nr \
  | head
```

Observed:

```text
[...]

PR #2805, opened 2026-06-11
fix(cli): tolerate dead leader pane in detached history prune hook

lifrary:
## Summary

The detached-session history prune hook introduced in #2598 registers, per session:

```
client-detached[<hash>]  if-shell -F '#{==:#{session_attached},0}' 'clear-history -t %N'
```

When the leader pane is already dead but the session is still alive (e.g. the leader process crashed during launch while the HUD pane lingered), every `client-detached` event makes the hook run `clear-history` against a pane that no longer exists. Hook commands execute without a client, so tmux routes the failure into its config-cause queue and presents it to the **next attaching client** as a full-screen error view:

```
(null):0: can't find pane: %0
```

This repeats on every subsequent detach, so the user sees the error overlay every time they open/attach **any** session on that tmux server — including sessions that have nothing to do with omx. Observed in the wild on tmux 3.6a / macOS after an `omx` launch crashed mid-bootstrap, leaving the session, its HUD pane, and the stale hook behind.

This is the adjacent failure mode of the scenario #2597 describes (interrupted/stuck Codex TUI in a detached session): there the leader pane stayed alive and grew history; here the leader pane died, and the prune hook itself becomes the noise source.

## Changes

- `buildDetachedHistoryPruneHookCommand` now routes the prune action through `run-shell -b "tmux clear-history -t %N >/dev/null 2>&1 || true"`, matching the repo's existing best-effort hook idiom (`buildBestEffortShellCommand` / `buildNestedTmuxShellCommand` in `src/team/tmux-session.ts`, used by the HUD resize hooks). A missing leader pane becomes a silent no-op while live panes are still pruned; the `#{session_attached}` guard is unchanged. Happy to switch to reusing those helpers directly if you prefer the import over the inline string.
- This is deliberately fixed in the hook command rather than by unregistering the hook when the leader pane dies: the triggering scenario is a crashed launch, and crash paths cannot be relied on to run cleanup. The hook has to tolerate a missing target on its own — the same reasoning as the best-effort resize hooks.
- Updated the two existing command-string assertions (`index.test.ts`, `launch-fallback.test.ts`) and added a regression test pinning the error-suppressed form.

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run test:node` (repo runner, 356 files) — 5891 pass / 1 skipped / 0 fail, exit 0
- [x] Touched suites via raw `node --test`: `index.test.js` 278/278, `launch-fallback.test.js` 22/22
- [ ] `omx doctor` (when setup/config behavior changes) — N/A, no setup/config change

Additional verification:

- Regression test confirmed to fail against the unpatched source (checked out base `src/cli/index.ts`, rebuilt, observed the intended assertion failures, restored the fix).
- Live A/B on a real tmux 3.6a scratch server (fresh `-L` socket; `%0` killed after a split to simulate the dead leader):
  - old hook + dead pane → next attach shows the `(null):0: can't find pane: %0` error view (bug reproduced)
  - new hook + dead pane → nothing queued, attach is clean
  - new hook + live pane → `#{history_size}` went 473 → 0, i.e. the prune still works through all three quoting layers (`set-hook` → `if-shell` → `run-shell`)

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — N/A
- [x] Backward-compatibility impact considered — hook slot and registration step are unchanged; only the action string differs and takes effect on the next session launch

## Related

Introduced by #2598 (the fix for #2597); this covers the case where the leader pane is gone by the time the hook fires.


iqdoctor:
## Consolidated review verdict: COMMENT

**Final recommendation:** COMMENT  
**Blocking findings:** none  
**Architecture status:** WATCH  
**Scholastic gate:** WATCH

This edited comment consolidates the original code-review verdict and the Scholastic / ontology follow-up. The separate Scholastic follow-up remains as audit trail, but is superseded by this consolidated verdict.

### Findings and recommendations

- **Architecture WATCH — async/background hook semantics**  
  `src/cli/index.ts:1178-1182` changes the detached history prune from direct `clear-history` to `run-shell -b "tmux clear-history ..."`. This fixes the dead-leader-pane noise, but changes timing from immediate check+clear to deferred/background clear.  
  **Recommendation:** document the intended best-effort semantics or add an invariant/behavioral test if the attach/detach race matters.

- **Architecture + Scholastic WATCH — broad suppression category**  
  `src/cli/index.ts:1182` uses `>/dev/null 2>&1 || true`, making the intended “missing/dead leader pane is a silent no-op” into “all clear-history failures are silent no-ops.” This is the main ontology finding: it broadens one expected causal class into every failure class.  
  **Recommendation:** narrow suppression to expected missing-pane failures where feasible, or add debug-only telemetry for non-missing-pane failures while preserving no user-visible hook spam.

- **Architecture WATCH — tmux command resolution consistency**  
  `src/cli/index.ts:1182` hardcodes ambient `tmux` inside a shell string. This may diverge from the repo’s resolved/quoted tmux invocation conventions in other paths.  
  **Recommendation:** prefer the existing resolved/quoted tmux command pattern, or explicitly document why this hook intentionally depends on ambient `tmux`.

- **Test coverage WATCH — string shape vs behavior**  
  `src/cli/__tests__/index.test.ts:4827-4848` and `src/cli/__tests__/launch-fallback.test.ts:760-763` validate the emitted command string, but not the operational distinction between expected dead-pane no-op and unexpected suppressed failures.  
  **Recommendation:** add a behavioral/negative-path test if the hook contract is “dead pane only is silent.”

### Scholastic impact on verdict

Scholastic gate **does not change** the overall review verdict. It adds WATCH evidence around semantic broadening, hidden assumptions, and modality separation, but found no category error or BLOCK-level issue.

### Verification evidence reviewed

- Live PR metadata/diff fetched for PR #2805; PR is open, non-draft, base `dev`, head `fix/detached-history-prune-dead-pane`, mergeStateStatus `BEHIND`.
- Changed files reviewed: `src/cli/index.ts`, `src/cli/__tests__/index.test.ts`, `src/cli/__tests__/launch-fallback.test.ts`.
- Independent code-reviewer lane: 0 issues, recommended APPROVE.
- Independent architect lane: WATCH for async hook semantics, broad suppression, PATH/resolution consistency, and string-only tests.
- Independent Scholastic / ontology gate: WATCH; no BLOCK; no verdict change.
- Local checks: `npx tsc --noEmit` passed; `git diff --check dev...HEAD` passed.
- PR status checks reviewed as green for format, clippy, Rust tests, lint, typecheck, build, Node test shards, coverage gate, and CI status.

No code changes were made in this review lane.

iqdoctor:
## Scholastic / ontology review gate: WATCH

**Verdict:** WATCH

**Findings:**
- **Semantic broadening / modal conflation:** `src/cli/index.ts:1182` makes the intended “missing leader pane is a silent no-op” behavior into “all `tmux clear-history` failures are silent no-ops” via `>/dev/null 2>&1 || true`. That flattens distinct failure categories: expected dead-pane failure vs PATH/tooling/malformed-target failures.
- **Hidden assumption:** the hook shell can always resolve `tmux` from ambient PATH (`run-shell -b "tmux clear-history ..."`) rather than using the repo's normal resolved tmux invocation conventions.
- **Contract/test boundary:** tests at `src/cli/__tests__/index.test.ts:4827-4848` and `src/cli/__tests__/launch-fallback.test.ts:760-763` assert the emitted command shape, but do not behaviorally distinguish the intended dead-pane no-op from other suppressed failure modes.

**Minimal repair proposals:**
- Narrow suppression to the expected missing-pane case where feasible, or add debug-only telemetry for non-missing-pane failures while preserving no user-visible hook spam.
- Prefer the repo's existing resolved/quoted tmux command pattern for nested shell calls, or document why this hook intentionally depends on ambient `tmux`.
- Add a behavioral/negative-path test if the hook contract is meant to be “dead pane only is silent.”

**Evidence reviewed:** live PR metadata/diff, existing review comment, changed files `src/cli/index.ts`, `src/cli/__tests__/index.test.ts`, `src/cli/__tests__/launch-fallback.test.ts`, and local `git diff --check dev...HEAD` (clean).

**Impact on overall review verdict:** no change. Overall verdict remains **COMMENT** (non-blocking WATCH item, no BLOCK).

iqdoctor:
## Consolidated review summary

**Final recommendation:** `COMMENT`  
**Blocking findings:** none

### Combined findings + recommendations (with file:line evidence)

- **`src/cli/index.ts:1178-1182`** — Functional intent is good: detached-session hook now uses `if-shell` to run history prune only when detached, and avoids direct fail-fast on dead leader pane.
  - ✅ Aligns with the PR goal of tolerating dead/absent leader pane during detached finalization.

- **`src/cli/index.ts:1178-1182`** — **Non-blocking watch item:** `run-shell -b` changes timing from inline execution to backgrounded async execution; behavior changes from immediate check+clear to deferred clear.
  - **Impact:** narrow race window if attach state changes quickly around finalization.
  - **Recommendation:** if this race is important, consider documenting or adding an explicit invariant test path that confirms harmlessness (no false positives or missed prune cases).

- **`src/cli/index.ts:1178-1182`** — **Non-blocking watch item:** `>/dev/null 2>&1 || true` suppresses all `clear-history` failures, not just “dead/missing pane” failures.
  - **Impact:** unexpected failures (permission/command/tooling issues) are hidden.
  - **Recommendation (minimal):** narrow suppression to the missing/dead pane case only, or emit telemetry/debug logging on non-expected failures while still returning non-blocking behavior.

- **`src/cli/index.ts:1178-1182`** — **Non-blocking consistency watch item:** shell fragment hardcodes `tmux` instead of reusing the repo’s resolved/quoted tmux invocation convention (where applicable elsewhere).
  - **Impact:** behavior can diverge from environments relying on alternate tmux discovery/quoting patterns.
  - **Recommendation:** keep the resilient semantics and align command construction with existing project command-resolution patterns.

- **`src/cli/index.ts:3788-3806`** — Hook registration location is coherent (`buildDetachedSessionFinalizeSteps`) and keeps the detached-session path explicit, which is correct for this fix.
  - **Recommendation:** none required.

- **`src/cli/__tests__/index.test.ts:4827-4848` and `src/cli/__tests__/launch-fallback.test.ts:760-763`** — Current assertions validate emitted command string shape.
  - **Impact:** operational semantics (actual no-op vs failure masking under async/background execution) are not covered.
  - **Recommendation:** add a behavioral/negative-path test for the detached-prune hook, e.g., simulating missing leader pane and unexpected `clear-history` failure to confirm intended non-blocking and scoped behavior.

### Verification evidence

- PR CI checks reported green (format, clippy, Rust tests, lint, typecheck, build, Node shards, coverage, CI status).
- Local checks run:
  - `git diff --check dev...HEAD` — passed.
  - `npx tsc --noEmit` — passed.
  - A direct Node TS test invocation failed in this environment due repository expectation around compiled `dist` for `.js`-suffixed imports (pre-existing setup nuance), not due this diff.

### Note

This comment **supersedes and combines** the earlier separate review comments/lane outputs and **does not change the verdict**: **non-blocking `COMMENT`**.

PR #2830, opened 2026-06-15
fix: make deep-interview/RALPLAN Bash write detector target-aware (#2829)

### LINKED ISSUE #2829: Bug: deep-interview PreToolUse Bash detector flags read-only grep and variable artifact redirects
## Summary

The deep-interview/RALPLAN PreToolUse guard can still produce false positives for Bash commands after the `apply_patch` artifact-path fix (#2809/#2818).

I observed this while diagnosing `oh-my-codex@0.18.11` in an attached tmux Codex session. I also checked current `dev` source and the same detector shape appears to remain there.

This report is intentionally narrow: it is not asking to relax the planning-mode write boundary. It is about making the Bash write-intent detector target-aware enough to avoid blocking read-only diagnostics and allowed artifact writes.

## Environment

- `oh-my-codex`: 0.18.11 from npm
- Runtime surface: attached tmux Codex session with native hooks enabled
- Platform: macOS / Darwin
- Active workflow: `$deep-interview`

## Case A: read-only diagnostics containing the token `apply_patch`

A read-only command such as:

```sh
grep -n "apply_patch" dist/scripts/codex-native-hook.js
```

can be blocked with the deep-interview planning boundary message.

Expected: allowed, because this is a read-only search.

Actual: blocked/misclassified as write intent because the detector checks for `\bapply_patch\b` anywhere in the shell command text.

Current `dev` source still appears to contain this broad check in `commandHasDeepInterviewWriteIntent(...)`:

```ts
return /\bapply_patch\b/.test(command)
  || extractDeepInterviewCommandRedirectTargets(command).length > 0
  // ...
```

## Case B: allowed artifact write through a same-command variable redirect

This shape can also be blocked:

```sh
SNAP=".omx/context/example.md"
cat > "$SNAP" <<'EOF'
content
EOF
```

Expected: allowed, because the resolved target is under the documented allowed deep-interview artifact prefix `.omx/context/`.

Actual: blocked, because `extractDeepInterviewCommandRedirectTargets(...)` extracts the literal redirect target `$SNAP` rather than resolving the simple same-command literal assignment.

## Why this matters

`$deep-interview` needs to block implementation writes, but it also needs to permit safe planning artifacts and normal read-only diagnostics. These false positives can stall development preparation and can even interfere with debugging the hook itself.

## Related issues checked

- #2809 / #2818 cover `apply_patch` freeform patch payloads writing allowed artifacts, and are marked fixed on `dev` via #2810.
- #2726 covers stderr redirects such as `2>/dev/null`, fixed on `dev` via #2730.
- I did not find an open issue for read-only `grep "apply_patch"` false positives or variable-based allowed artifact redirects.
- #2822 may currently block npm users from receiving fixes already present in `dev`, but the two Bash detector cases above appear separate from the `apply_patch` payload parser fix.

## Suggested fix

1. Make the shell `apply_patch` check invocation-aware. For example, match only when `apply_patch` is the command being invoked at the start of a command segment, not when it appears as a grep/search argument.
2. For redirect and `tee` targets, optionally resolve simple same-command literal assignments such as `SNAP=.omx/context/a.md; cat > "$SNAP"` before applying the allowed-prefix check.
3. If variable expansion is intentionally out of scope, document that planning-mode artifact writes must use literal redirect paths, and improve the block message so agents can recover without investigating the hook.

## Workarounds

- Use literal redirect paths, not variables, when writing allowed `.omx/*` artifacts during planning/interview modes.
- Avoid read-only diagnostic commands that include the literal token `apply_patch` while deep-interview/RALPLAN guards are active.

Yeachan-Heo:
## Summary

Fixes two false positives in the deep-interview/RALPLAN PreToolUse Bash write-intent detector reported in #2829. The detector is now target-aware: read-only diagnostics and allowed artifact writes pass, while the planning write boundary stays intact.

### Case A — read-only diagnostics mentioning `apply_patch`
Previously `commandHasDeepInterviewWriteIntent` matched `\bapply_patch\b` anywhere in the command, so a read-only search like `grep -n "apply_patch" dist/scripts/codex-native-hook.js` was misclassified as write intent.

New `commandInvokesApplyPatch` only treats `apply_patch` as write intent when it is the command being invoked at a command position (statement start, or after a `|`/`;`/`&`/`(` boundary, optionally via `sudo`/`command`/`exec`). Real `apply_patch` invocations remain blocked.

### Case B — allowed artifact writes via same-command literal variable redirect
`SNAP=".omx/context/x.md"; cat > "$SNAP"` was blocked because the redirect-target extractor returned the literal `$SNAP`.

New `extractCommandLiteralAssignments` + `resolveCommandRedirectTarget` resolve redirect/tee targets against simple same-command literal assignments before the allowed-prefix check. Values involving expansion (`$`, backticks) are skipped, so unresolved/dynamic targets and redirects to disallowed paths stay conservatively blocked.

## Scope / safety
- Does **not** relax the planning-mode write boundary.
- Real `apply_patch` invocations and redirects to non-artifact paths remain blocked.
- Shared between deep-interview and RALPLAN guards (same helpers), so both modes get the fix.

## Tests
- Added focused tests in `src/scripts/__tests__/codex-native-hook.test.ts`:
  - allows read-only grep mentioning `apply_patch`; blocks a real `apply_patch` Bash invocation.
  - allows variable redirect resolving to `.omx/context/`; blocks variable redirects to disallowed and unresolved targets.
- Full `codex-native-hook.test.ts` suite passes (425/425).
- `biome lint`, `tsc`, `tsc -p tsconfig.no-unused.json`, and `verify:plugin-bundle` all pass.

Closes #2829

—

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -5951,6 +5951,22 @@ exit 0
       );
       assert.equal(allowedSubshellGrep.outputJson, null);
 
+      // Double-quoted spans that merely mention the literal token, expand a
+      // parameter, or run a substitution that is not `apply_patch` stay allowed
+      // — the quoted-substitution fix must not over-block read-only diagnostics.
+      const allowedQuotedMention = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-grep-quoted-mention",
+          tool_input: { command: 'echo "${apply_patch} $(echo apply_patch)"' },
+        },
+        { cwd },
+      );
+      assert.equal(allowedQuotedMention.outputJson, null);
+
       const blockedApplyPatch = await dispatchCodexNativeHook(
         {
           hook_event_name: "PreToolUse",
@@ -6006,6 +6022,11 @@ exit 0
         { id: "tool-di-apply-patch-subshell-env", command: `(env apply_patch <<'EOF'${heredocBody}\n)` },
         { id: "tool-di-apply-patch-command-substitution", command: `x=$(apply_patch <<'EOF'${heredocBody}\n)` },
         { id: "tool-di-apply-patch-brace-group", command: `{ apply_patch <<'EOF'${heredocBody}\n}` },
+        // Command substitution runs even inside double quotes, so quoting the
+        // already-blocked `$(…)` / `` `…` `` form must not bypass the guard.
+        { id: "tool-di-apply-patch-quoted-command-substitution", command: `echo "$(apply_patch <<'EOF'${heredocBody}\n)"` },
+        { id: "tool-di-apply-patch-quoted-backtick", command: `echo "\`apply_patch <<'EOF'${heredocBody}\`"` },
+        { id: "tool-di-apply-patch-quoted-command-substitution-prefixed", command: `echo "patched: $(apply_patch <<'EOF'${heredocBody}\n) done"` },
       ];
       for (const form of blockedRealApplyPatchForms) {
         const blockedForm = await dispatchCodexNativeHook(
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -5976,6 +5976,35 @@ exit 0
         { cwd },
       );
       assert.equal((blockedEnvWrapperApplyPatch.outputJson as { decision?: string } | null)?.decision, "block");
+
+      const heredocBody = "\n*** Begin Patch\n*** Add File: src/leak.ts\n+export const x = 1;\n*** End Patch\nEOF";
+      const blockedRealApplyPatchForms: Array<{ id: string; command: string }> = [
+        { id: "tool-di-apply-patch-env-i", command: `env -i apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-env-unset", command: `env -u FOO apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-env-i-assignment", command: `env -i FOO=bar apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-assignment-env", command: `FOO=bar env apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-exec-env-assignment", command: `exec env FOO=bar apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-absolute-path", command: `/usr/bin/apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-relative-path", command: `./apply_patch <<'EOF'${heredocBody}` },
+      ];
+      for (const form of blockedRealApplyPatchForms) {
+        const blockedForm = await dispatchCodexNativeHook(
+          {
+            hook_event_name: "PreToolUse",
+            cwd,
+            session_id: "sess-di-grep",
+            tool_name: "Bash",
+            tool_use_id: form.id,
+            tool_input: { command: form.command },
+          },
+          { cwd },
+        );
+        assert.equal(
+          (blockedForm.outputJson as { decision?: string } | null)?.decision,
+          "block",
+          `expected deep-interview to block real apply_patch form: ${form.command}`,
+        );
+      }
     } finally {
       await rm(cwd, { recursive: true, force: true });
     }
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -5950,6 +5950,32 @@ exit 0
         { cwd },
       );
       assert.equal((blockedApplyPatch.outputJson as { decision?: string } | null)?.decision, "block");
+
+      const blockedEnvAssignmentApplyPatch = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-apply-patch-env-assignment",
+          tool_input: { command: "FOO=bar apply_patch <<'EOF'\n*** Begin Patch\n*** Add File: src/leak.ts\n+export const x = 1;\n*** End Patch\nEOF" },
+        },
+        { cwd },
+      );
+      assert.equal((blockedEnvAssignmentApplyPatch.outputJson as { decision?: string } | null)?.decision, "block");
+
+      const blockedEnvWrapperApplyPatch = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-apply-patch-env-wrapper",
+          tool_input: { command: "env FOO=bar apply_patch <<'EOF'\n*** Begin Patch\n*** Add File: src/leak.ts\n+export const x = 1;\n*** End Patch\nEOF" },
+        },
+        { cwd },
+      );
+      assert.equal((blockedEnvWrapperApplyPatch.outputJson as { decision?: string } | null)?.decision, "block");
     } finally {
       await rm(cwd, { recursive: true, force: true });
     }
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -5903,6 +5903,123 @@ exit 0
     }
   });
 
+  it("allows read-only diagnostics mentioning apply_patch while deep-interview blocks real apply_patch invocations", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-pretool-deep-interview-grep-apply-patch-"));
+    try {
+      const stateDir = join(cwd, ".omx", "state");
+      const sessionDir = join(stateDir, "sessions", "sess-di-grep");
+      await mkdir(sessionDir, { recursive: true });
+      await writeJson(join(stateDir, "session.json"), { session_id: "sess-di-grep", cwd });
+      await writeJson(join(sessionDir, "skill-active-state.json"), {
+        version: 1,
+        active: true,
+        skill: "deep-interview",
+        phase: "planning",
+        session_id: "sess-di-grep",
+        active_skills: [{ skill: "deep-interview", phase: "planning", active: true, session_id: "sess-di-grep" }],
+      });
+      await writeJson(join(sessionDir, "deep-interview-state.json"), {
+        active: true,
+        mode: "deep-interview",
+        current_phase: "intent-first",
+        session_id: "sess-di-grep",
+      });
+
+      const allowedGrep = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-grep",
+          tool_input: { command: 'grep -n "apply_patch" dist/scripts/codex-native-hook.js' },
+        },
+        { cwd },
+      );
+      assert.equal(allowedGrep.outputJson, null);
+
+      const blockedApplyPatch = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-apply-patch-invoke",
+          tool_input: { command: "apply_patch <<'EOF'\n*** Begin Patch\n*** Add File: src/leak.ts\n+export const x = 1;\n*** End Patch\nEOF" },
+        },
+        { cwd },
+      );
+      assert.equal((blockedApplyPatch.outputJson as { decision?: string } | null)?.decision, "block");
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
+  it("allows deep-interview same-command literal variable redirects to artifacts while blocking variable redirects outside them", async () => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-pretool-deep-interview-var-redirect-"));
+    try {
+      const stateDir = join(cwd, ".omx", "state");
+      const sessionDir = join(stateDir, "sessions", "sess-di-var-redirect");
+      await mkdir(sessionDir, { recursive: true });
+      await writeJson(join(stateDir, "session.json"), { session_id: "sess-di-var-redirect", cwd });
+      await writeJson(join(sessionDir, "skill-active-state.json"), {
+        version: 1,
+        active: true,
+        skill: "deep-interview",
+        phase: "planning",
+        session_id: "sess-di-var-redirect",
+        active_skills: [{ skill: "deep-interview", phase: "planning", active: true, session_id: "sess-di-var-redirect" }],
+      });
+      await writeJson(join(sessionDir, "deep-interview-state.json"), {
+        active: true,
+        mode: "deep-interview",
+        current_phase: "intent-first",
+        session_id: "sess-di-var-redirect",
+      });
+
+      const allowedVarRedirect = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-var-redirect",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-var-redirect-allow",
+          tool_input: { command: 'SNAP=".omx/context/example.md"\ncat > "$SNAP" <<\'EOF\'\ncontent\nEOF' },
+        },
+        { cwd },
+      );
+      assert.equal(allowedVarRedirect.outputJson, null);
+
+      const blockedVarRedirect = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-var-redirect",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-var-redirect-block",
+          tool_input: { command: 'SNAP="src/leak.ts"\ncat > "$SNAP" <<\'EOF\'\ncontent\nEOF' },
+        },
+        { cwd },
+      );
+      assert.equal((blockedVarRedirect.outputJson as { decision?: string } | null)?.decision, "block");
+
+      const blockedUnresolvedVarRedirect = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-var-redirect",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-var-redirect-unresolved",
+          tool_input: { command: 'cat > "$SNAP" <<\'EOF\'\ncontent\nEOF' },
+        },
+        { cwd },
+      );
+      assert.equal((blockedUnresolvedVarRedirect.outputJson as { decision?: string } | null)?.decision, "block");
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
   it("allows deep-interview apply_patch artifact writes from freeform patch text while blocking outside paths", async () => {
     const cwd = await mkdtemp(join(tmpdir(), "omx-native-hook-pretool-deep-interview-apply-patch-"));
     try {
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -5938,6 +5938,19 @@ exit 0
       );
       assert.equal(allowedGrep.outputJson, null);
 
+      const allowedSubshellGrep = await dispatchCodexNativeHook(
+        {
+          hook_event_name: "PreToolUse",
+          cwd,
+          session_id: "sess-di-grep",
+          tool_name: "Bash",
+          tool_use_id: "tool-di-grep-subshell",
+          tool_input: { command: '(grep -n "apply_patch" dist/scripts/codex-native-hook.js)' },
+        },
+        { cwd },
+      );
+      assert.equal(allowedSubshellGrep.outputJson, null);
+
       const blockedApplyPatch = await dispatchCodexNativeHook(
         {
           hook_event_name: "PreToolUse",
@@ -5986,6 +5999,13 @@ exit 0
         { id: "tool-di-apply-patch-exec-env-assignment", command: `exec env FOO=bar apply_patch <<'EOF'${heredocBody}` },
         { id: "tool-di-apply-patch-absolute-path", command: `/usr/bin/apply_patch <<'EOF'${heredocBody}` },
         { id: "tool-di-apply-patch-relative-path", command: `./apply_patch <<'EOF'${heredocBody}` },
+        { id: "tool-di-apply-patch-subshell", command: `(apply_patch <<'EOF'${heredocBody}\n)` },
+        { id: "tool-di-apply-patch-subshell-spaced", command: `( apply_patch <<'EOF'${heredocBody}\n)` },
+        { id: "tool-di-apply-patch-double-subshell", command: `((apply_patch <<'EOF'${heredocBody}\n))` },
+        { id: "tool-di-apply-patch-pipe-subshell", command: `true | (apply_patch <<'EOF'${heredocBody}\n)` },
+        { id: "tool-di-apply-patch-subshell-env", command: `(env apply_patch <<'EOF'${heredocBody}\n)` },
+        { id: "tool-di-apply-patch-command-substitution", command: `x=$(apply_patch <<'EOF'${heredocBody}\n)` },
+        { id: "tool-di-apply-patch-brace-group", command: `{ apply_patch <<'EOF'${heredocBody}\n}` },
       ];
       for (const form of blockedRealApplyPatchForms) {
         const blockedForm = await dispatchCodexNativeHook(
--- a/src/scripts/codex-native-hook.ts
+++ b/src/scripts/codex-native-hook.ts
@@ -2752,11 +2752,15 @@ function isNullDeviceRedirectTarget(target: string): boolean {
 }
 
 // Matches `apply_patch` only at a shell command position (statement start, or
-// after a pipe/`;`/`&`/`(` boundary, optionally via `sudo`/`command`/`exec`),
-// so read-only diagnostics that merely mention the literal token (e.g.
-// `grep -n "apply_patch" file`) are not misread as write intent.
+// after a pipe/`;`/`&`/`(` boundary, optionally via an `env` prefix and/or
+// zero-or-more leading `NAME=VALUE` assignments and a `sudo`/`command`/`exec`
+// wrapper), so read-only diagnostics that merely mention the literal token
+// (e.g. `grep -n "apply_patch" file`) are not misread as write intent, while
+// real env-prefixed invocations (`FOO=bar apply_patch`, `env FOO=bar
+// apply_patch`) stay blocked — mirroring the git commit guard's handling of
+// leading env assignments / `env` wrapper prefixes.
 function commandInvokesApplyPatch(command: string): boolean {
-  return /(?:^|[\n;&|(]|&&|\|\|)\s*(?:sudo\s+|command\s+|exec\s+)?apply_patch\b/.test(command);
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
## Merge-gate review (restarted) — PR #2830 → `dev`

**Verdict: REQUEST_CHANGES** (one in-scope, demonstrable boundary regression; the rest of the diff is sound)

Re-ran the red-team against `dev` at base `57267863` with head `a4b456ac`. Build is green and the focused suite passes (`node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` → **425/425**, including the 3 new tests). `biome lint` on the changed file is clean.

### What passes ✅
- **Read-only `grep` mentioning `apply_patch` is allowed.** `grep -n "apply_patch" dist/...` → not at a command position → no write intent → allowed (verified via real `dispatchCodexNativeHook`).
- **Plain real `apply_patch` stays blocked.** `apply_patch <<'EOF' …` → blocked. `;`/newline/`&&`/`(`/`sudo`/`command`/`exec`/pipe boundaries all still detect it.
- **Same-command literal var redirects to artifacts are allowed safely.** `SNAP=".omx/context/x.md"; cat > "$SNAP"` → resolved and allowed; `.omx/context/*` confined.
- **Disallowed / ambiguous variable redirects stay blocked.** `SNAP="src/leak.ts"` → blocked; unresolved `cat > "$SNAP"` (no literal assignment) → target left literal → blocked. Resolution is conservative in the safe direction: values containing `$`/backtick expansion are skipped, composite targets (`"$SNAP/../etc"`) are not resolved, and `resolve()`+`relative()` normalization defeats `..` traversal. Last-assignment-wins matches shell semantics.

### Blocker 🔴 — write boundary is weakened vs. base (criterion #5)

`commandInvokesApplyPatch` only recognizes `apply_patch` after a statement boundary, optionally via `sudo`/`command`/`exec`. It does **not** account for a leading env-assignment or `env` prefix, so a real Bash `apply_patch` invocation slips past write-intent detection in deep-interview **and** RALPLAN (both share `commandHasDeepInterviewWriteIntent`/`extractDeepInterviewCommandWriteTargets`).

Reproduced end-to-end through `dispatchCodexNativeHook` (deep-interview active, phase `intent-first`):

```
BLOCKED | apply_patch <<'EOF' *** Add File: src/leak.ts …      (plain — correct)
ALLOWED | FOO=bar apply_patch <<'EOF' *** Add File: src/leak.ts …   ← regression
ALLOWED | env FOO=bar apply_patch <<'EOF' …                          ← regression
BLOCKED | FOO=bar; apply_patch <<'EOF' …                       (separator — correct)
```

Pre-PR vs post-PR detection on the same input:

| command | pre-PR (`/\bapply_patch\b/`) | post-PR (`commandInvokesApplyPatch`) |
|---|---|---|
| `FOO=bar apply_patch <<EOF` | detected → **blocked** | not detected → **allowed** |
| `env FOO=bar apply_patch <<EOF` | detected → **blocked** | not detected → **allowed** |

This is strictly worse than `dev` for those inputs: the planning agent can write implementation files (e.g. `src/leak.ts`) during the interview/plan phase. It is narrow (the native `apply_patch`/`ApplyPatch` **tool** path is unaffected, and the plain form an agent normally emits is still blocked), but it is a genuine regression of the exact guard this PR is meant to preserve. The repo already treats env-prefix as a must-handle evasion class — see the Lore commit guard's `blocks PreToolUse env-prefixed git commit …` / `env wrapper-prefixed git.exe` tests — so leaving it open here is an inconsistency, not an accepted trade-off.

### Suggested fix (minimal, mirrors the existing git-guard handling)
Allow an optional `env` and zero-or-more leading `NAME=VALUE` assignments before the command token:

```ts
function commandInvokesApplyPatch(command: string): boolean {
  return /(?:^|[\n;&|(]|&&|\|\|)\s*(?:env\s+)?(?:[A-Za-z_][A-Za-z0-9_]*=(?:"[^"]*"|'[^']*'|[^\s'"]+)\s+)*(?:sudo\s+|command\s+|exec\s+)?apply_patch\b/.test(command);
}
```

Please add a regression test for `FOO=bar apply_patch …` (blocked) alongside the existing grep/plain cases, then this is good to merge.

—

Yeachan-Heo:
## Fix pushed — env-prefixed `apply_patch` regression closed

Addressed the REQUEST_CHANGES blocker in `6a434bef` (branch `issue-2829-deep-interview-bash-detector`).

### What changed
`commandInvokesApplyPatch` now accepts an optional `env` prefix and zero-or-more leading `NAME=VALUE` assignments before the command token, mirroring the git commit guard's env-prefix evasion handling:

```ts
function commandInvokesApplyPatch(command: string): boolean {
  return /(?:^|[\n;&|(]|&&|\|\|)\s*(?:env\s+)?(?:[A-Za-z_][A-Za-z0-9_]*=(?:"[^"]*"|'[^']*'|[^\s'"]+)\s+)*(?:sudo\s+|command\s+|exec\s+)?apply_patch\b/.test(command);
}
```

This restores parity with the base `\bapply_patch\b` behavior for the two regressed inputs without relaxing the read-only grep / `.omx/context/*` redirect fixes.

### Regression coverage added
Added to the existing grep/apply_patch test in `src/scripts/__tests__/codex-native-hook.test.ts`, driven end-to-end through `dispatchCodexNativeHook` (deep-interview active, phase `intent-first`):

| command | decision |
|---|---|
| `apply_patch <<'EOF' …` | **block** (unchanged) |
| `FOO=bar apply_patch <<'EOF' …` | **block** (was the regression) |
| `env FOO=bar apply_patch <<'EOF' …` | **block** (was the regression) |
| `grep -n "apply_patch" dist/…` | allow (preserved) |
| `SNAP=".omx/context/example.md"; cat > "$SNAP"` | allow (preserved) |
| `SNAP="src/leak.ts"; cat > "$SNAP"` / unresolved `$SNAP` | block (preserved) |

### Verification
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` → **425/425 pass**, including the env-prefixed regression assertions.
- `npx @biomejs/biome lint` on both changed files → clean.
- `tsc` (full build) and `tsc -p tsconfig.no-unused.json` → clean.
- `npm run verify:plugin-bundle` → clean.

Only `src/scripts/codex-native-hook.ts` and `src/scripts/__tests__/codex-native-hook.test.ts` changed. Not merging — pushed for re-review.

—

Yeachan-Heo:
## Merge-gate re-review — PR #2830 → `dev` (head `6a434bef`)

**Verdict: REQUEST_CHANGES** — the two *named* env-prefixed forms are fixed, but the underlying class is only partially closed: several real `apply_patch` Bash invocations that base `dev` blocked still slip past the planning write guard, demonstrated end-to-end.

Re-ran the red-team on head `6a434bef` (base `dev`, merge-base `57267863`). Build, focused suite, and lint are green:
- `tsc` (full) + `tsc -p tsconfig.no-unused.json` → clean.
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` → **425/425 pass**, incl. the two new env-prefix regression assertions.
- `npx @biomejs/biome lint src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts` → clean.

### What the fix got right ✅
- **The named blocker is closed.** `FOO=bar apply_patch <<'EOF' …` and `env FOO=bar apply_patch <<'EOF' …` are now `block` (verified end-to-end via `dispatchCodexNativeHook`, deep-interview active, phase `intent-first`), with regression tests added.
- **No regression to the original #2829 fixes.** Read-only `grep -n "apply_patch" …` stays allowed; same-command literal `SNAP=".omx/context/x.md"; cat > "$SNAP"` resolves and is allowed; `SNAP="src/leak.ts"` / unresolved `$SNAP` stay blocked; the planning write boundary (`isAllowedPlanningArtifactPath` → `resolve`/`relative` + prefix confinement, `..`/absolute rejected) is untouched. RALPLAN parity holds (both modes share `commandHasDeepInterviewWriteIntent`).

### Blocker 🔴 — env-prefix fix is incomplete; boundary still weaker than base for real `apply_patch` invocations

`commandInvokesApplyPatch` now accepts an `env` *wrapper* and leading `NAME=VALUE` assignments, but **not** `env` *flags*, leading-assignment/`env` re-ordering, the `exec env …` order, or path-qualified binaries. Base `dev` used `/\bapply_patch\b/`, which blocked all of these. Reproduced end-to-end through the real `dispatchCodexNativeHook` (deep-interview active, `intent-first`), each writing `src/leak.ts`:

| command | base `\bapply_patch\b` | head `6a434bef` |
|---|---|---|
| `apply_patch <<'EOF' …` | block | **block** ✅ |
| `FOO=bar apply_patch <<'EOF' …` | block | **block** ✅ (fixed) |
| `env FOO=bar apply_patch <<'EOF' …` | block | **block** ✅ (fixed) |
| `env -i apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `env -u FOO apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `env -i FOO=bar apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `FOO=bar env apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `exec env FOO=bar apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `/usr/bin/apply_patch <<'EOF' …` | block | **allow** 🔴 |
| `./apply_patch <<'EOF' …` | block | **allow** 🔴 |

`env -i` / `env -u FOO` are the *same env-prefix class* the original blocker named — the fix handled `env NAME=VAL apply_patch` but not the equally-standard `env -flag apply_patch`. These let the planning agent write implementation files during the interview/plan phase, which is exactly what criterion #5 (planning write boundary must not be weakened vs base for real `apply_patch` invocations) forbids.

The fix's stated rationale — *"mirroring the git commit guard's handling of env-prefix evasions"* and *"restores parity with the base `\bapply_patch\b` behavior"* — does not hold. The sibling git-commit guard (`findGitCommandTokenIndex` in `src/scripts/codex-native-pre-post.ts`) is a real tokenizer that strips leading assignments, the `env` executable **with** its options (`-i`/`--ignore-environment`, `-u`/`--unset`, `--unset=`, `-C`, `-S`/`--split-string`, `--`, interleaved assignments), and recognizes path-qualified / `.exe` binaries by basename — it has explicit tests for `env -i …`, `env -u FOO …`, and `"C:/…/git.exe" commit`. The new regex mirrors none of that, so it is strictly weaker than both the guard it cites and the base it claims parity with.

### Suggested fix (mirror the guard it cites)
Detect `apply_patch` via the same tokenized command-head walk the git guard uses rather than a flat regex: after a statement boundary, skip leading `NAME=VALUE` assignments, then the `env` executable and its option grammar, then `sudo`/`command`/`exec` wrappers, and match when the resolved command token's basename is `apply_patch`. The needed predicates already exist in `codex-native-pre-post.ts` (`isInlineShellEnvAssignment`, `isEnvExecutableToken`, `envOptionConsumesNextValue`) — reuse them so the two guards stay consistent. Then extend the regression test to cover `env -i apply_patch`, `env -u FOO apply_patch`, and `/usr/bin/apply_patch` (all `block`) alongside the existing cases.

Everything else in the diff is sound; closing this residual env-flag/path-qualified gap makes it good to merge. Not merged.

—

Yeachan-Heo:
## Fix pushed — env-flag / reordered / path-qualified `apply_patch` gap closed (`8434b6ed`)

Closed the residual env-flag/path-qualified write-boundary gap from the last re-review by replacing the flat-regex `commandInvokesApplyPatch` with a tokenized command-head walk that mirrors the git commit guard (`findGitCommandTokenIndex`), reusing the same predicates (`tokenizeShellCommandWithBoundaries`, `isInlineShellEnvAssignment`, `isEnvExecutableToken`, `envOptionConsumesNextValue`, `shellCommandBasename`, `removeHereDocBodies`) so the two guards stay consistent.

### What changed
`commandInvokesApplyPatch` now lives in `src/scripts/codex-native-pre-post.ts` (exported, imported by `codex-native-hook.ts`). After each statement boundary it:
1. strips heredoc bodies first (so patch payloads can't break tokenization),
2. skips leading inline `NAME=VALUE` assignments,
3. skips the `env` executable **with its full option grammar** (`-i`/`--ignore-environment`, `-u`/`--unset`/`--unset=`, `-C`/`-S` and other value-taking options, `--`, interleaved assignments),
4. skips `sudo`/`command`/`exec` wrappers — in any order — then
5. matches when the resolved command token's **basename** is `apply_patch`.

This restores (and exceeds) parity with the base `\bapply_patch\b` behavior for every form the red-team named, without relaxing the read-only `grep` / `.omx/context/*` redirect fixes. Both deep-interview and RALPLAN get it (shared `commandHasDeepInterviewWriteIntent`).

### Regression coverage added
Extended the existing grep/apply_patch test in `src/scripts/__tests__/codex-native-hook.test.ts`, driven end-to-end through `dispatchCodexNativeHook` (deep-interview active, phase `intent-first`). Every form below now writes `src/leak.ts` → `block`:

| command | base `\bapply_patch\b` | head `8434b6ed` |
|---|---|---|
| `apply_patch <<'EOF' …` | block | **block** |
| `FOO=bar apply_patch <<'EOF' …` | block | **block** |
| `env FOO=bar apply_patch <<'EOF' …` | block | **block** |
| `env -i apply_patch <<'EOF' …` | block | **block** (was allow) |
| `env -u FOO apply_patch <<'EOF' …` | block | **block** (was allow) |
| `env -i FOO=bar apply_patch <<'EOF' …` | block | **block** (was allow) |
| `FOO=bar env apply_patch <<'EOF' …` | block | **block** (was allow) |
| `exec env FOO=bar apply_patch <<'EOF' …` | block | **block** (was allow) |
| `/usr/bin/apply_patch <<'EOF' …` | block | **block** (was allow) |
| `./apply_patch <<'EOF' …` | block | **block** (was allow) |

Preserved (no regression to #2829 fixes): read-only `grep -n "apply_patch" …` → allow; `SNAP=".omx/context/x.md"; cat > "$SNAP"` → allow; `SNAP="src/leak.ts"` / unresolved `$SNAP` → block.

### Verification
- `tsc` (full) + `tsc -p tsconfig.no-unused.json` → clean.
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` → **425/425 pass**, incl. the new env-flag/path-qualified regression assertions.
- `npx @biomejs/biome lint` on the three changed files → clean.
- `npm run verify:plugin-bundle` → clean.

Only `src/scripts/codex-native-pre-post.ts`, `src/scripts/codex-native-hook.ts`, and `src/scripts/__tests__/codex-native-hook.test.ts` changed. Not merging — pushed for re-review.

—

Yeachan-Heo:
## Final re-review — REQUEST_CHANGES

Re-reviewed at head `8434b6ed64f6d7607c65dfc0dc20161d0810f74d` (base `dev`). Built clean and ran the full `dist/scripts/__tests__/codex-native-hook.test.js` suite: **425/425 pass**, including the new `blockedRealApplyPatchForms` cases.

### ✅ Prior blockers — all fixed
Exercised `commandInvokesApplyPatch` directly (unit) and end-to-end through `dispatchCodexNativeHook` with active deep-interview state. All nine previously-flagged forms now block:

| form | result |
|---|---|
| `FOO=bar apply_patch` | block ✅ |
| `env FOO=bar apply_patch` | block ✅ |
| `env -i apply_patch` | block ✅ |
| `env -u FOO apply_patch` | block ✅ |
| `env -i FOO=bar apply_patch` | block ✅ |
| `FOO=bar env apply_patch` | block ✅ |
| `exec env FOO=bar apply_patch` | block ✅ |
| `/usr/bin/apply_patch` | block ✅ |
| `./apply_patch` | block ✅ |

### ✅ Original-issue behavior — intact
End-to-end with deep-interview active:
- read-only `grep -n "apply_patch" …` / `rg apply_patch …` → **ALLOW** ✅
- literal-var redirect `SNAP=".omx/context/x.md"; cat > "$SNAP"` → **ALLOW** ✅
- direct redirect to `.omx/context/note.md` → **ALLOW** ✅
- literal-var redirect to `src/leak.ts` → **block** ✅
- unresolved `cat > "$SNAP"` → **block** ✅
- direct redirect to `src/leak.ts` → **block** ✅

### ❌ Blocker — refactor reintroduces an `apply_patch` write bypass (subshell / command substitution)

The previous flat regex used the boundary class `[\n;&|(]`, so `apply_patch` immediately after `(` was caught. The new `commandInvokesApplyPatch` reuses `tokenizeShellCommandWithBoundaries`, which only treats `\n ; & |` as statement boundaries (`codex-native-pre-post.ts:364`) — **not** `(`, `{`, or `$(`. So `apply_patch` directly after a grouping/command-substitution opener never resolves to a command head and slips through.

Verified both at unit level (`commandInvokesApplyPatch`) and end-to-end (deep-interview active → `decision`):

| form | old regex (pre-fix) | this PR | e2e under deep-interview |
|---|---|---|---|
| `(apply_patch <<'EOF' … EOF)` | **blocked** | `false` | **ALLOW** — regression |
| `( apply_patch <<'EOF' … )` | **blocked** | `false` | **ALLOW** — regression |
| `((apply_patch <<'EOF' …))` | **blocked** | `false` | regression |
| `true \| (apply_patch <<'EOF' …)` | **blocked** | `false` | regression |
| `(env apply_patch <<'EOF' …)` | **blocked** | `false` | regression |
| `x=$(apply_patch <<'EOF' …)` | **blocked** | `false` | regression |
| `{ apply_patch <<'EOF' … ; }` | allowed | `false` | pre-existing gap (not a regression) |

`commandHasDeepInterviewWriteIntent` has no other clause that catches a heredoc-only `apply_patch` (no redirect/tee/sed-i), and both `isAllowedDeepInterviewBashWrite` and `isAllowedRalplanBashWrite` route through it — so an agent under deep-interview **or** RALPLAN can write arbitrary files via `(apply_patch <<'EOF' *** Add File: src/leak.ts … EOF)`, defeating the planning write boundary this PR exists to enforce.

Root cause is the shared tokenizer: the same blind spot affects the git-commit Lore guard (`findGitCommandTokenIndex`). Confirmed e2e with `OMX_LORE_COMMIT_GUARD=1`: `(git commit -m "fix tests")` → **ALLOW** while plain `git commit -m "fix tests"` → **block**. (For git this pre-dates the PR; for `apply_patch` this PR introduces it by dropping the `(` boundary.)

### Suggested fix
Treat grouping / command-substitution openers as command-position boundaries so the command-head walk resumes after them — e.g. in `tokenizeShellCommandWithBoundaries`, split on a leading `(`/`{` and set `startsCommand` on the following token (and handle `$(`/backtick), mirroring the old `[\n;&|(]` boundary set. Add regression cases for `(apply_patch …)`, `( apply_patch …)`, `x=$(apply_patch …)`, and `{ apply_patch …; }`. Fixing it in the shared tokenizer also closes the parallel `(git commit …)` Lore-guard hole.

**Verdict: REQUEST_CHANGES** — do not merge to `dev` until the subshell/command-substitution `apply_patch` bypass is closed.

—

Yeachan-Heo:
## Fix pushed — subshell / command-substitution `apply_patch` bypass closed (`9e322f97`)

Closed the final REQUEST_CHANGES blocker: the tokenized `commandInvokesApplyPatch` no longer slips a real `apply_patch` invoked immediately after a grouping / command-substitution opener.

### Root cause
`tokenizeShellCommandWithBoundaries` only treated `\n ; & |` as statement boundaries, so `apply_patch` after `(`, `((`, `$(`, a backtick, or a `{` group never resolved to a command head. The previous flat regex used the boundary class `[\n;&|(]`, so dropping `(` during the tokenizer refactor reintroduced the bypass.

### What changed
Extended the shared `tokenizeShellCommandWithBoundaries` so grouping / command-substitution openers act as command-position boundaries and the command-head walk resumes after them:
- `(`, `)`, and backtick are boundaries (covers `(`, `((`, `$(` command substitution, and backtick substitution).
- `{` is a boundary only as its own word (empty pending token + following whitespace), so `{ apply_patch …; }` is caught while brace expansion (`{a,b}`) and parameter expansion (`${VAR}`) are left intact.
- Quotes and heredocs are still respected, so a literal like `grep "(apply_patch"` is **not** split — this is strictly tighter than the old quote-blind regex.

Both deep-interview and RALPLAN get it (shared `commandHasDeepInterviewWriteIntent`). Fixing the shared tokenizer also closes the parallel `(git commit …)` Lore-guard hole the review noted (`parseGitCommitCommand` now resolves `(git commit …)` / `true | (git commit …)`).

### Coverage (end-to-end via `dispatchCodexNativeHook`, deep-interview active, phase `intent-first`)

| command | this PR |
|---|---|
| `(apply_patch <<'EOF' …)` | **block** (was allow) |
| `( apply_patch <<'EOF' …)` | **block** (was allow) |
| `((apply_patch <<'EOF' …))` | **block** (was allow) |
| `true \| (apply_patch <<'EOF' …)` | **block** (was allow) |
| `(env apply_patch <<'EOF' …)` | **block** (was allow) |
| `x=$(apply_patch <<'EOF' …)` | **block** (was allow) |
| `{ apply_patch <<'EOF' … }` | **block** (pre-existing gap, now closed) |

Preserved (no regression): plain / env / env-flag / path-qualified `apply_patch` stay **block**; read-only `grep -n "apply_patch" …`, subshell-wrapped `(grep -n "apply_patch" …)`, quoted `grep "(apply_patch" …`, `.omx/context/*` literal-var redirects → **allow**; disallowed / unresolved redirects → **block**.

### Verification
- `tsc` (full build) and `tsc -p tsconfig.no-unused.json` → clean.
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` → **425/425 pass**, including the new subshell/command-substitution/group block cases and the subshell-grep allow case.
- `npx @biomejs/biome lint src/scripts/codex-native-pre-post.ts src/scripts/__tests__/codex-native-hook.test.ts` → clean.
- `npm run verify:plugin-bundle` → clean.

Only `src/scripts/codex-native-pre-post.ts` and `src/scripts/__tests__/codex-native-hook.test.ts` changed. Not merging — pushed for re-review.

—

Yeachan-Heo:
## Final red-team re-review @ `9e322f97` — verdict: REQUEST_CHANGES

The r5 push **does** close the exact blocker from the prior review, cleanly and additively, and CI is green (all required checks pass). But a focused red-team of the final state surfaced a **command-substitution bypass that survives the new tokenizer** — a regression vs `dev` and a contradiction of this PR's own "block subshell/command-substitution apply_patch" claim. It must be addressed before merge.

### What r5 fixed (verified ✅)
Direct probes of `commandInvokesApplyPatch` + full `dispatchCodexNativeHook` (deep-interview active) — all **BLOCK**:
`(apply_patch …)`, `( apply_patch …)`, `((apply_patch …))`, `true | (apply_patch …)`, `(env apply_patch …)`, `x=$(apply_patch …)`, `{ apply_patch …; }`, plus path-qualified (`/usr/bin/…`, `./…`), env-flag (`env -i`, `env -u FOO`, `env --unset=`, `env -C`, `env --`), reordered (`FOO=bar env apply_patch`, `exec env FOO=bar apply_patch`), and `sudo`/`command`/`exec` wrappers. 28/28 true-positive forms blocked.
No false positives (26/26 safe): `grep -n "apply_patch" f`, `grep "(apply_patch" f`, `${apply_patch}`, `{a,b}`, `VAR=apply_patch echo`, comments, `find -name apply_patch`, etc.
`(git commit …)`/`((…))`/`x=$(git commit …)` now detected; `git help commit`, `git config alias.ci commit`, `git tag commit`, quoted forms correctly **not** flagged. Focused suite: **425/425 pass**.

### Blocker: quoted command substitution bypasses the write guard
Full-dispatch repro with deep-interview active (`decision`):

```
x=$(apply_patch <<'EOF' … EOF)        -> BLOCK   (correct)
echo "$(apply_patch <<'EOF' … EOF)"   -> ALLOW   (BYPASS)
echo "`apply_patch <<'EOF' … EOF`"    -> ALLOW   (BYPASS)
```

Command substitution **executes inside double quotes**, and `apply_patch`'s file-write side effect still fires. So a one-character edit (wrap the already-blocked `$(…)` in quotes) defeats the guard. The code comment justifies skipping quoted spans as "read-only diagnostics that merely mention the literal token" — that premise is wrong for `"$(…)"` / `` "`…`" ``: those are execution, not mentions.

This is a **regression vs `dev`**: the previous `/\bapply_patch\b/` matched (and blocked) all of these. The tokenizer's precision win (no longer false-positiving on `grep "apply_patch"`) is good, but it dropped a genuine write form that is the direct quoted twin of an explicitly-fixed one.

Same class also bypasses (lower priority, noted for completeness — all blocked by `dev`, allowed now): `eval "apply_patch …"`, `bash -c "apply_patch …"`, `sh -c '…'`, `… | xargs apply_patch`, `env -S 'apply_patch …'`. (The `(git commit …)` guard has the analogous `"$(git commit …)"` gap, but that's best-effort Lore enforcement, lower stakes.)

### Actionable fix (does not reintroduce the grep false positive)
The quoted-mention vs quoted-execution cases are distinguishable: a double-quoted span containing `$(` or an unescaped backtick is command substitution, not a literal mention. Concretely, one of:
- In `commandInvokesApplyPatch`, when inside double quotes, recognize `$(` / `` ` `` as a nested command-position boundary and resume the command-head walk (mirrors the unquoted handling you already added) — leaves `grep "(apply_patch"` (no `$(`/backtick) intact.
- And/or treat indirect-exec heads (`eval`, `bash -c`/`sh -c`, `xargs`, `env -S`) carrying an `apply_patch` argument as write intent.

Add coverage to `codex-native-hook.test.ts` for `echo "$(apply_patch …)"`, `` echo "`apply_patch …`" ``, and `eval "apply_patch …"` so the quoted/indirect forms stay blocked.

### Verdict
`REQUEST_CHANGES` — the named blocker is closed and tests/CI are green, but a final security red-team can't green-light a write guard whose claimed-covered class (command substitution) is bypassed by adding quotes. Close the `"$(apply_patch …)"` / `` "`apply_patch …`" `` gap (and ideally the `eval`/`bash -c`/`xargs`/`env -S` heads), then this is good to land.

—

Yeachan-Heo:
## Fix pushed — quoted command-substitution `apply_patch` bypass closed (`4205a6a3`)

Closed the latest REQUEST_CHANGES blocker from the red-team re-review at `9e322f97`: `apply_patch` inside command substitution still executes inside double quotes, so forms like `echo "$(apply_patch …)"` and ``echo "`apply_patch …`"`` must be treated as real write attempts, not literal mentions.

### Root cause
The tokenizer treated the whole double-quoted span as an inert literal, so the command-head walk never saw the nested `$(...)` / backtick command substitution. That made a real `apply_patch` invocation slip past the planning write guard.

### What changed
- Added nested command-substitution handling inside double-quoted spans.
- Added regression coverage for quoted `$(apply_patch …)` and quoted backtick `apply_patch` forms.
- Preserved safe/literal mentions such as read-only grep and escaped/literal `apply_patch` text.

### Verification
- `npm run build` ✅
- `npx tsc -p tsconfig.no-unused.json` ✅
- `npx @biomejs/biome lint src/scripts/codex-native-pre-post.ts src/scripts/__tests__/codex-native-hook.test.ts` ✅
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` ✅ — 425/425 pass
- Direct `commandInvokesApplyPatch` probe ✅:
  - `echo "$(apply_patch …)"` → block
  - quoted backtick `apply_patch` → block
  - `echo "${apply_patch} $(echo apply_patch)"` → allow
  - `grep -n "apply_patch" file.ts` → allow

Terminal verdict: MERGE_READY pending GitHub CI for the newly pushed head.

—

Yeachan-Heo:
## Fix pushed — quoted command-substitution `apply_patch` bypass closed (`4205a6a3`)

Closed the latest REQUEST_CHANGES blocker from the red-team re-review at `9e322f97`: `apply_patch` inside command substitution still executes inside double quotes, so forms like `echo "$(apply_patch …)"` and ``echo "`apply_patch …`"`` must be treated as real write attempts, not literal mentions.

### Root cause
The tokenizer treated the whole double-quoted span as an inert literal, so the command-head walk never saw the nested `$(...)` / backtick command substitution. That made a real `apply_patch` invocation slip past the planning write guard.

### What changed
- Added nested command-substitution handling inside double-quoted spans.
- Added regression coverage for quoted `$(apply_patch …)` and quoted backtick `apply_patch` forms.
- Preserved safe/literal mentions such as read-only grep and escaped/literal `apply_patch` text.

### Verification
- `npm run build` ✅
- `npx tsc -p tsconfig.no-unused.json` ✅
- `npx @biomejs/biome lint src/scripts/codex-native-pre-post.ts src/scripts/__tests__/codex-native-hook.test.ts` ✅
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` ✅ — 425/425 pass
- Direct `commandInvokesApplyPatch` probe ✅:
  - `echo "$(apply_patch …)"` → block
  - quoted backtick `apply_patch` → block
  - `echo "${apply_patch} $(echo apply_patch)"` → allow
  - `grep -n "apply_patch" file.ts` → allow

Terminal verdict: MERGE_READY pending GitHub CI for the newly pushed head.

—

PR #2837, opened 2026-06-16
[codex] Warn about external Codex MCP process guards

MadKangYu:
## Summary

- Adds an `omx doctor` warning for macOS LaunchAgents that can terminate Codex/MCP helper processes outside OMX setup ownership.
- Detects the observed Codex app-server MCP child dedupe guard pattern and explicit XcodeBuildMCP cleanup guards.
- Adds regression coverage using an isolated temporary HOME and fake LaunchAgent/script fixtures.

## Root Cause

A local external LaunchAgent can kill Codex app-server MCP child processes while standard OMX MCP configuration still looks healthy. Before this change, `omx doctor` could report a healthy OMX install while SIGTERM or transport resets were actually caused by out-of-band process cleanup.

## Validation

- `npm run build`
- `node dist/scripts/run-test-files.js dist/cli/__tests__/doctor-warning-copy.test.js`
- `npm run lint`
- `node dist/cli/omx.js doctor`

### DIFF
--- a/src/cli/__tests__/doctor-warning-copy.test.ts
+++ b/src/cli/__tests__/doctor-warning-copy.test.ts
@@ -20,6 +20,7 @@ import {
 } from "./packaged-explore-harness-lock.js";
 import {
 	checkExploreHarness,
+	checkExternalCodexProcessGuards,
 	checkNativeHookDistSmoke,
 	classifyPostCompactHookStdout,
 } from "../doctor.js";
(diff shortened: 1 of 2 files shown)


MadKangYu:
Closing this after re-checking the change at atomic scope. The patch only teaches `omx doctor` to warn about a local macOS LaunchAgent pattern, while current local evidence shows the guard is no longer an active .plist and OMX already has MCP duplicate/orphan cleanup coverage. Keeping this upstream PR would add user-environment-specific warning logic rather than a general OMX fix.

PR #2859, opened 2026-06-18
fix(plugin): avoid Windows native hook shell wrapping

Yeachan-Heo:


Yeachan-Heo (commented):
MERGE_READY

Adversarial merge-readiness review of PR #2859 at HEAD 143a7ba45b4f1cb934be8561a999eca3d270cc09.

Evidence:
- Diff scope is limited to `plugins/oh-my-codex/hooks/codex-native-hook.mjs`, `src/cli/__tests__/codex-plugin-layout.test.ts`, and `src/scripts/sync-plugin-mirror.ts`.
- The launcher now keeps Windows direct `.exe`/`.com` commands off shell wrapping and preserves shell fallback for non-direct Windows commands such as `.cmd` shims. The pinned cache writer uses `process.execPath` plus `dist/cli/omx.js`, so this targets the reported Windows node.exe failure path without changing non-Windows spawn options.
- Stop-hook fallback/error handling, stdout filtering, and non-Stop fail-closed behavior remain unchanged around the spawn site.
- CI reported success for required lanes, including Typecheck, Build dist artifact, Node test shards, coverage gate, and CI Status.
- Local targeted verification: initial `npm run build` and plugin-boundary tests failed only because dependencies were absent from the review worktree (`@iarna/toml`, `@modelcontextprotocol/sdk`). After `npm ci`, `npm run build` passed and `npm run test:plugin-boundaries:compiled` passed: 47 tests across the plugin/layout/bin/setup/SSOT boundary files.

Non-blocker risk: the new regression test is source-shape based rather than a true Windows spawn simulation, but the behavior change is small, covered by the existing plugin-boundary executable tests on Linux, and CI is green. I do not see a merge-blocking regression, security drift, or platform contract break in this diff.

—

Issue #2328, opened 2026-05-13
Detect and protect AGENTS.md from external tool overwrites


debop:
## Summary

`~/.codex/AGENTS.md` can be overwritten by other Codex ecosystem tools. When that happens, OMX may still have `config.toml`, hooks, skills, and prompts installed, but new sessions lose the top-level OMX orchestration contract and local user guidance.

I hit this after installing `context-mode`: its Codex setup copied its own `configs/codex/AGENTS.md` into `~/.codex/AGENTS.md`, replacing the OMX-generated AGENTS content.

## Environment

- macOS
- Codex CLI: `0.130.0`
- oh-my-codex: `0.16.0`
- context-mode: `1.0.124`
- setup scope: user

`omx doctor` still reported most OMX components as OK because config/hooks/skills/prompts were present. The missing part was semantic: `AGENTS.md` no longer contained the OMX contract or my local policy sections.

## Impact

After the overwrite, future sessions could miss:

- OMX operating contract and workflow routing
- native subagent/delegation rules
- local response style and language policy
- qmd knowledge retrieval policy
- Lore commit protocol
- project-specific issue selection priorities

This is silent unless the user manually inspects `AGENTS.md` or has a custom self-audit marker check.

## Requested behavior

Please consider adding stronger protection around `AGENTS.md` ownership and integrity:

1. `omx doctor` should validate required AGENTS markers, not only file existence.
2. `omx setup` could preserve or restore marker-bounded user/local sections when regenerating AGENTS.md.
3. If `AGENTS.md` exists but no longer contains the OMX contract, doctor/setup should emit a high-severity warning with a repair command.
4. If possible, support a documented overlay/merge model for other tools that also need AGENTS.md instructions.

Potential markers to validate:

- `oh-my-codex - Intelligent Multi-Agent Orchestration`
- `AUTONOMY DIRECTIVE`
- `Lore Commit Protocol`
- `OMX:MODELS:START`
- expected runtime/team overlay markers

## Local workaround

I added chezmoi-side validation so sync fails if `AGENTS.md` is missing required local markers, and self-audit now fails when the merged AGENTS contract is absent.

That protects my machine, but OMX itself would be a better place to warn users before they run with a partially installed-but-semantically-broken setup.

## Related upstream issue

I also opened a context-mode issue asking it to merge instead of replacing existing Codex AGENTS.md:

- https://github.com/mksglu/context-mode/issues/552

[...]

PR #1946, opened 2026-04-27
Preserve AGENTS.md guidance during setup refresh

Yeachan-Heo:
## Summary
- Add explicit `omx setup --merge-agents` support to preserve existing AGENTS.md guidance while inserting/refreshing OMX-managed generated sections between `<!-- OMX:AGENTS:START -->` / `<!-- OMX:AGENTS:END -->` markers.
- Keep existing default safety: unmanaged AGENTS.md files are skipped in non-interactive setup unless `--merge-agents` or `--force` is explicit, and active project sessions still block project-root AGENTS rewrites.
- Document the new flag in CLI help, README setup notes, and the setup skill/plugin mirror.
- Add regression coverage for unmarked user AGENTS merge, user content preservation, idempotent repeated merge, managed model table refresh, active-session protection, and default non-interactive safety.

Fixes #1898

## Verification
- `npm install` (restored missing dependencies in the local workspace)
- `npm run build`
- `node --test dist/cli/__tests__/setup-agents-overwrite.test.js`
- `npm run lint`
- `npm run check:no-unused`
- `npm run verify:plugin-bundle`
- `git diff --check -- README.md skills/omx-setup/SKILL.md plugins/oh-my-codex/skills/omx-setup/SKILL.md src/cli/__tests__/setup-agents-overwrite.test.ts src/cli/index.ts src/cli/setup.ts src/utils/agents-md.ts`

Note: full `npm test` was started and progressed through many suites, but the user directed not to wait on the stalling background run after the targeted gates above had passed.

## Signature
Implemented by Codex / OMX Ralph workflow.

Issue #1898, opened 2026-04-24
Add merge option for existing AGENTS.md in omx setup


cometzero:
## Summary

`omx setup --scope=project` currently has safe handling for an existing `AGENTS.md`, but it does not provide a general merge path for user-authored content.

Please add an explicit merge option so existing project guidance can be preserved while adding or refreshing the OMX-managed `AGENTS.md` scaffold.

## Current behavior

From the current setup flow:

- If an existing, unmarked `AGENTS.md` is present, non-interactive setup skips overwriting it by default.
- In an interactive run, or with `--force`, setup can overwrite the existing file after taking a backup.
- If the file contains OMX-owned markers, setup can refresh the managed model capability table only.
- There is no option to merge the generated OMX guidance into an existing user-authored `AGENTS.md` while preserving the user content in place.

## Requested behavior

Add an explicit merge mode, for example one of:

- `omx setup --scope=project --merge-agents`
- `omx setup --scope=project --agents-mode=merge`
- an interactive prompt choice that offers `skip`, `overwrite`, or `merge`

The merge mode should:

- preserve existing user-authored `AGENTS.md` content;
- insert or refresh OMX-managed sections using clear marker boundaries;
- avoid duplicating generated sections across repeated setup runs;
- keep the existing backup behavior before modifying the file;
- continue to skip by default in non-interactive runs unless the user explicitly requests merge or force;
- continue to respect the active-session guard that avoids rewriting project-root `AGENTS.md` during an active OMX session.

## Suggested acceptance criteria

- Add regression coverage for merging an unmarked, user-authored `AGENTS.md` with the generated OMX scaffold.
- Verify user-authored content remains after merge.
- Verify repeated merge runs are idempotent and do not duplicate generated sections.
- Verify existing OMX-managed model table refresh still works.
- Verify active-session protection still prevents project-root `AGENTS.md` mutation.
- Document the new option in setup/help docs.

## Motivation

Many repositories already have project-specific `AGENTS.md` guidance. Today users must choose between preserving the file unchanged or replacing it with the generated OMX scaffold. A merge option would make `omx setup` safer and more useful for existing projects that want to adopt OMX without losing their local instructions.

Yeachan-Heo:
Thanks for the detailed request. This is understandable, but it adds a new `omx setup` option/merge mode and changes the setup/user-guidance contract, so I’m holding it for owner/product confirmation before implementation.

If accepted, the likely shape should stay narrow:
[...]

Issue #1782, opened 2026-04-20
[Bug] omx setup overwrites AGENTS.md without creating a backup


MilesCranmer:
## Summary

`omx setup` prompts before overwriting an existing `AGENTS.md`, but if the user accepts, the original file is destroyed with no backup. User content is unrecoverable.

## Environment

- OS: macOS (Darwin 25.4.0, arm64)
- Node.js version: v25.9.0
- `oh-my-codex` version: v0.12.6
- Codex CLI version: 0.121.0

## Steps to reproduce

1. Create `~/.codex/AGENTS.md` with custom content (e.g., project-specific style rules, code principles)
2. Run `omx setup`
3. Accept the overwrite prompt: `Overwrite existing AGENTS.md at "/Users/mcranmer/.codex/AGENTS.md"? [y/N]: y`
4. Original content is gone with no backup created

## Expected behavior

Before overwriting, `omx setup` should move the existing file to a backup (e.g., `.AGENTS.md.bkup` or `.AGENTS.md.bkup{N}` with an incrementing index), and print the backup path so the user knows where to find it.

## Actual behavior

The original file is overwritten in place. No backup is created. If the user didn't manually save their content beforehand, it's permanently lost.

## Proposed fix

Before the overwrite write, `mv` the existing `AGENTS.md` to `.AGENTS.md.bkup{N}` (where N increments past any existing backups), and print something like:

```
Backed up existing AGENTS.md to .AGENTS.md.bkup
To restore: mv ~/.codex/.AGENTS.md.bkup ~/.codex/AGENTS.md
```

Such a message could also be printed to the user when running `omx uninstall`, for them to do as they like with their backed up copy.

Yeachan-Heo:
Investigating this now in a fix worktree. Reproducing the current overwrite path and patching setup so an existing `AGENTS.md` is preserved via backup instead of being destroyed in-place.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2892, opened 2026-06-18
feat(setup): merge AGENTS.md by default, add --no-merge-agents opt-out

vikaswalia:
## Problem

`omx setup` overwrites an existing root `AGENTS.md` by default. On install/update this silently drops user-authored project guidance. This has been reported repeatedly: #1521, #1671, #1782, #2328.

The non-destructive merge already exists behind `--merge-agents` (added in #1898 / PR #1946), but it is opt-in, so the *default* path is still the destructive one.

## Change

Make the marker-based merge the **default** behavior of `omx setup`, with explicit opt-outs:

- **Default (no flags):** merge — refresh only the OMX-managed block between `<!-- OMX:AGENTS:START -->` / `<!-- OMX:AGENTS:END -->`, preserve everything outside it. On a file with no markers yet, the existing `upsertManagedAgentsBlock` appends the block instead of clobbering.
- **`--no-merge-agents`:** opt out — restores the previous behavior (refresh managed sections only, or prompt before overwriting).
- **`--force`:** unchanged — full replace after backup.
- **`--merge-agents`:** still accepted (now redundant; also re-enables merge alongside `--force`).

The decision is a small pure helper, `resolveMergeAgents(flags)`, wired into the CLI option parsing in `src/cli/index.ts`. The `setup()` logic and its existing tests are untouched — only the CLI default routes to the merge path now.

## Why default-merge

- Losing user content should require an explicit choice, not be the fallback.
- The merge path is already idempotent and marker-based, so it is safe to run on every setup/update.
- Principle of least surprise: a routine `omx update` should never drop project guidance.

## Backward compatibility

Fully-generated `AGENTS.md` files (which already carry the OMX markers) keep refreshing as before. Only the previously-destructive "overwrite a customized file without a flag" case changes — it now merges. `--merge-agents` continues to work.

## Tests / verification

- New `src/cli/__tests__/agents-merge-mode.test.ts` — covers the default and every flag combination.
- `biome lint` clean, `tsc` builds.
- Existing `setup-agents-overwrite` suite passes unchanged (23/23).

## Docs

Updated `--help`, `README.md`, and `plugins/oh-my-codex/skills/omx-setup/SKILL.md` to describe merge-as-default and `--no-merge-agents`.

## Notes

Targeting `dev` per CONTRIBUTING. Happy to adjust the opt-out flag name (e.g. `--overwrite-agents`) or move the default into `setup()` core if you'd prefer it there.


Yeachan-Heo:
Closing this because it changes the default `omx setup` behavior and AGENTS.md handling semantics, which is a maintainer-owned workflow contract. This is not something we can take as an external default-behavior flip in its current form.

If we revisit this area, it needs a narrower maintainer-owned design and compatibility plan rather than changing the default install/update behavior directly.

—

PR #2861, opened 2026-06-18
fix(hud): stabilize standalone two-pane HUD ownership

Yeachan-Heo:
## Summary

Closes #2860.

- Treat standalone HUD topology as scoped to the owning leader pane geometry, not always full-window width.
- Preserve the prior full-width recreate behavior for full-width leaders and detached/unknown leader geometry.
- Add a regression for two live standalone leaders in one tmux window keeping independent pane-scoped HUDs stable without cross-owner kill/recreate churn.

## Verification

- `npm run build`
- `PATH=/tmp/omx-issue2860-fakebin:$PATH env -u TMUX -u TMUX_PANE node --test dist/hud/__tests__/reconcile.test.js dist/hud/__tests__/tmux.test.js dist/hud/__tests__/hud-tmux-injection.test.js` — 122/122 pass. The fake tmux shim only supplies roomy `display-message` geometry so tests are not polluted by this agent's own cramped tmux pane.
- Live bounded tmux smoke with manual two-pane session: created two leader panes and two pane-scoped HUD panes, ran `omx hud --reconcile-tmux` for both leaders, and confirmed the original HUD panes stayed alive (`%7`, `%8`) with two HUD panes total.

## Notes

- Checked for open PRs mentioning #2860/standalone HUD flicker before opening; none were present.
- #2828 is Team shutdown/HUD ownership and #2859 is Windows hook shell wrapping; neither already fixes this standalone same-window pane-scoped reconcile path.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2978, opened 2026-06-26
fix(hud): make standalone-tmux HUD reconcile non-blocking and self-healing

### LINKED ISSUE #2977: [Bug] Managed-mode HUD reconcile deadlocks on tmux wait-for lock and spawns 1000+ stuck processes (v0.18.15) — escalation of #2860
## Summary

On **v0.18.15** (the release that closed #2860 via #2861), the standalone-tmux HUD reconcile loop no longer just *flickers* — under **3+ concurrent `omx` leaders** sharing one tmux server it escalates into a **`tmux wait-for` lock deadlock** that spawns **1000+ stuck `sh` wrapper processes**, wedges tmux input, and requires manual `kill -9` by PID to recover. #2861 fixed the cosmetic 2-pane flicker; this is the same mechanism failing harder.

## Environment

- oh-my-codex **v0.18.15** (latest tag)
- Linux (Debian 13), tmux 3.x, Node 24
- Several `omx --madmax --xhigh` leaders running concurrently in one tmux server (different windows / projects), Codex in `--dangerously-bypass-approvals-and-sandbox`
- Managed mode (NOT `--direct`)

## Reproduction

1. One tmux server, multiple windows.
2. Start **3+** `omx` leaders (managed mode) across windows/projects.
3. Cause layout churn (resize the client, add/kill a pane, or just let the per-leader HUD panes coexist and reconcile).
4. The bottom HUD panes begin rapid create/remove (as in #2860), then **lock up** and processes accumulate without bound.

## Root cause

Managed mode installs tmux hooks `window-layout-changed[<id>]` and `client-resized[<id>]` that run `omx hud --reconcile-tmux`. The reconcile **mutates the layout** (creates/removes HUD panes) → that mutation **re-fires the same hooks** → another reconcile → unbounded feedback loop.

The hook command (`src/hud/tmux.ts`, `buildHudLayoutReconcileHookCommand`) wraps the reconcile in a **blocking** `tmux wait-for -L <hookName>_layout … wait-for -U`:

```
leaderAlive && ( wait-for -L lock ; unregister ; reconcile ; status=$? ; wait-for -U lock ; exit ) || unregister
```

Two failure modes:

1. **Unbounded queue** — `wait-for -L` *blocks*. Under rapid hook fires, each fire spawns a fresh `sh` wrapper that blocks acquiring the lock, so they pile up monotonically (observed **1380** stuck `sh` processes, tmux server saturated, client input unresponsive).
2. **Permanent deadlock** — if a reconcile process is killed (or exits) between `-L` and `-U`, the lock is **never released**, and every subsequent reconcile blocks on `wait-for -L` forever.

`reconcileHudForPromptSubmit` (`src/hud/reconcile.ts`) also has **no re-entrancy guard**, so nothing prevents N parallel reconcile processes from racing on the pane snapshot.

## Why it's hard to recover

- `pkill -f` does **not** match the stuck wrappers — their `sh -c env TMUX=… tmux …` cmdline is large and the matching token sits past pkill's cmdline read-limit. Recovery required `ps -eo pid,args | grep <lock> | awk '{print $1}' | xargs kill -9`.
- Unsetting the hooks (`set-hook -gu` / per-window) does not drain the already-queued waiters.
- A pane silently dropping into tmux `view-mode` during the churn additionally eats all client keystrokes (looks like "tmux is frozen"); and a wedged client needs `tmux detach-client` + reattach.

## Workaround

Running `omx --direct` (no tmux/HUD management) avoids it entirely.

## Proposed fix

Make the shell wrapper **non-blocking** and move concurrency control into the reconcile via a **non-blocking, self-healing cross-process file lock**:

1. `src/hud/tmux.ts` — drop the blocking `tmux wait-for -L/-U` serialization from the layout-reconcile hook command.
2. `src/hud/reconcile.ts` — acquire a non-blocking file lock at the top of `reconcileHudForPromptSubmit`; a competing reconcile returns a new `skipped_concurrent` status instead of piling up. A lock left by a crashed/killed holder is stolen once it is stale (> 10s) or its pid is no longer alive; the lock is released in a `finally` so a thrown reconcile never leaks it.

I have a PR ready with this fix + regression tests (full hud suite green, including #2861's two-pane test). Linking shortly.

Thanks for oh-my-codex.

funaileo:
## Summary

Closes #2977.

The standalone-tmux HUD reconcile fork-bombs under 3+ concurrent leaders. The `window-layout-changed`/`client-resized` hooks run `omx hud --reconcile-tmux`, which mutates the layout (split/kill HUD panes) and **re-fires the same hooks** → a feedback loop. The hook command wrapped the reconcile in a **blocking** `tmux wait-for -L/-U` serialization (`src/hud/tmux.ts`, `buildHudLayoutReconcileHookCommand`), so:

- under layout churn the fires **queue on the lock** into 1000+ stuck `sh` wrappers (observed 1380), saturating the tmux server and wedging client input; and
- a reconcile **killed between `-L` and `-U`** leaks the lock → every later reconcile blocks forever (permanent deadlock).

#2861 fixed the cosmetic 2-pane flicker but left this blocking-lock pile-up under more leaders.

## Fix

Make the shell wrapper **non-blocking** and gate concurrency inside the reconcile with a **non-blocking, self-healing cross-process file lock**.

- **`src/hud/tmux.ts`** — drop the blocking `tmux wait-for -L/-U` from the layout-reconcile hook command. It now just runs `leaderAlive && (unregister; reconcile) || unregister`, so the wrapper stays cheap and never blocks. (Comment added explaining the pile-up / deadlock it replaces.)
- **`src/hud/reconcile.ts`** — `reconcileHudForPromptSubmit` acquires a non-blocking file lock (keyed by session id / pane / `global`, under `os.tmpdir()`) right after the existing `skipped_not_omx_owned_tmux` / `skipped_no_entry` guards. A competing reconcile returns the new `skipped_concurrent` status instead of piling up. A lock left by a crashed/killed holder is **stolen** once it is stale (`RECONCILE_LOCK_STALE_MS = 10_000`) or its pid is no longer alive; a corrupt/unreadable lock file is treated as stealable rather than throwing. The lock is released in a `finally`, so a thrown reconcile never leaks it, and a SIGKILL'd holder self-heals on the next run.
  - New `deps`: `now?`, `lockDir?`, `processAlive?` (for tests). New status: `skipped_concurrent`.

## Tests

- **`src/hud/__tests__/reconcile.test.ts`** — three regressions sharing a real tmpdir `lockDir`: (1) a live lock → second reconcile returns `skipped_concurrent` and never touches panes; (2) a stale lock (old `ts` / dead `processAlive`) is stolen and the reconcile proceeds; (3) the lock file is gone after a normal reconcile (finally cleanup).
- **`src/hud/__tests__/tmux.test.ts`** — the layout-hook assertion flips from "contains `wait-for`" to "does not contain `wait-for`" (the blocking serialization is intentionally gone).

## Verification

- `npm run build` (tsc) — clean.
- `node --test dist/hud/__tests__/*.test.js` — **330 pass / 0 fail** (includes #2861's two-pane stability test, still green).
- `biome lint` on the changed files — clean; `npm run check:no-unused` — clean.

## Note / known edge

The common acquire path is atomic (`writeFileSync` with `wx` — exactly one winner). The *steal* path (stale/dead-pid recovery) does an `unlink` + rewrite that is not fully atomic, so two processes racing to steal the **same already-abandoned** lock could let **one** extra reconcile through during crash recovery. That is bounded to a single redundant run — never the unbounded pile-up this fixes — so I kept the simpler form; happy to tighten to an atomic rename-based steal if you'd prefer.

To keep the diff minimal the existing reconcile body is wrapped in `try { … } finally { … }` without re-indenting (biome + strict tsc pass). Glad to reflow for readability if preferred.

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -1907,4 +1907,105 @@ fi
     assert.equal(resized[0]?.paneId, '%2');
     assert.equal(resized[0]?.lines, HUD_TMUX_HEIGHT_LINES);
   });
+
+  it('skips a concurrent reconcile while another holds the cross-process lock', async () => {
+    // The tmux layout/resize hooks re-fire the reconcile while it is mutating the
+    // layout. A second reconcile sharing the same lock dir must back off instead
+    // of racing the first one's split/kill.
+    const lockDir = await mkdtemp(join(tmpdir(), 'omx-hud-reconcile-lock-'));
+    try {
+      const created: string[] = [];
+      const killed: string[] = [];
+
+      const sharedDeps = {
+        env: { TMUX: '1', TMUX_PANE: '%1', OMX_SESSION_ID: 'sess-lock', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+        lockDir,
+        listCurrentWindowPanes: () => [{ paneId: '%1', currentCommand: 'codex', startCommand: 'codex' }],
+        createHudWatchPane: () => {
+          created.push('create');
+          return '%9';
+        },
+        killTmuxPane: (paneId: string) => {
+          killed.push(paneId);
+          return true;
+        },
+        resizeTmuxPane: () => true,
+        unregisterHudResizeHook: noOpUnregisterHudResizeHook,
+        registerHudResizeHook: noOpRegisterHudResizeHook,
+        resolveOmxCliEntryPath: () => '/repo/dist/cli/omx.js',
+      };
+
+      // Pre-seed the lock as if a first reconcile (same key) were still running.
+      const lockKey = 'sess-lock';
+      const lockPath = join(lockDir, `omx-hud-reconcile-${lockKey}.lock`);
+      await writeFile(lockPath, JSON.stringify({ pid: process.pid, ts: Date.now() }), { flag: 'wx' });
+
+      const result = await reconcileHudForPromptSubmit('/repo', sharedDeps);
+
+      assert.equal(result.status, 'skipped_concurrent');
+      assert.equal(result.paneId, null);
+      assert.deepEqual(created, []);
+      assert.deepEqual(killed, []);
+      // The pre-existing live lock is left intact for its holder.
+      assert.equal((await readFile(lockPath, 'utf8')).length > 0, true);
+    } finally {
+      await rm(lockDir, { recursive: true, force: true });
+    }
+  });
+
+  it('steals a stale reconcile lock and proceeds', async () => {
+    const lockDir = await mkdtemp(join(tmpdir(), 'omx-hud-reconcile-lock-'));
+    try {
+      const created: string[] = [];
+
+      const lockPath = join(lockDir, 'omx-hud-reconcile-sess-stale.lock');
+      // A lock written long ago by a holder that is no longer alive.
+      await writeFile(lockPath, JSON.stringify({ pid: 999999, ts: 1 }), { flag: 'wx' });
+
+      const result = await reconcileHudForPromptSubmit('/repo', {
+        env: { TMUX: '1', TMUX_PANE: '%1', OMX_SESSION_ID: 'sess-stale', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+        lockDir,
+        now: () => 1_000_000, // well beyond the stale threshold relative to ts=1
+        processAlive: () => false,
+        listCurrentWindowPanes: () => [{ paneId: '%1', currentCommand: 'codex', startCommand: 'codex' }],
+        createHudWatchPane: () => {
+          created.push('create');
+          return '%9';
+        },
+        resizeTmuxPane: () => true,
+        unregisterHudResizeHook: noOpUnregisterHudResizeHook,
+        registerHudResizeHook: noOpRegisterHudResizeHook,
+        resolveOmxCliEntryPath: () => '/repo/dist/cli/omx.js',
+      });
+
+      assert.equal(result.status, 'recreated');
+      assert.equal(result.paneId, '%9');
+      assert.equal(created.length, 1);
+    } finally {
+      await rm(lockDir, { recursive: true, force: true });
+    }
+  });
+
+  it('removes the reconcile lock after a normal reconcile completes', async () => {
+    const lockDir = await mkdtemp(join(tmpdir(), 'omx-hud-reconcile-lock-'));
+    try {
+      const result = await reconcileHudForPromptSubmit('/repo', {
+        env: { TMUX: '1', TMUX_PANE: '%1', OMX_SESSION_ID: 'sess-clean', [OMX_TMUX_HUD_OWNER_ENV]: '1' },
+        lockDir,
+        listCurrentWindowPanes: () => [{ paneId: '%1', currentCommand: 'codex', startCommand: 'codex' }],
+        createHudWatchPane: () => '%9',
+        resizeTmuxPane: () => true,
+        unregisterHudResizeHook: noOpUnregisterHudResizeHook,
+        registerHudResizeHook: noOpRegisterHudResizeHook,
+        resolveOmxCliEntryPath: () => '/repo/dist/cli/omx.js',
+      });
+
+      assert.equal(result.status, 'recreated');
+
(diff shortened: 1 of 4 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Thanks for the quick fix. I ran the fallback merge-gate review after the GJC review lane aborted before tool execution. This should not merge as-is.

Evidence checked:
- PR is still `BEHIND` `dev` (4 behind / 4 ahead). `git merge-tree` against current `origin/dev` did not show conflict markers, but the branch still needs updating before final merge consideration.
- Local verification passed after `npm ci`: targeted compiled HUD tests (`dist/hud/__tests__/reconcile.test.js`, `dist/hud/__tests__/tmux.test.js`) passed 100/100, `npm run check:no-unused` passed, and Biome lint on the four changed files passed.
- Blocker: the lock steal condition treats age alone as sufficient. A live holder whose lock is older than `RECONCILE_LOCK_STALE_MS` is stolen even when `processAlive(pid)` reports the holder is still alive. I confirmed this with a compiled-code probe: preseed a lock `{ pid: process.pid, ts: 1 }`, set `now() => 20002`, `processAlive(pid) => true`, and the second reconcile still proceeds with `status: "recreated"`.

That means a legitimate reconcile lasting over 10s can overlap with a second reconcile, reintroducing the exact multi-process layout race this PR is meant to prevent. Please change stale stealing so a lock is only stealable when the holder is not alive, or when the lock is stale and liveness cannot be trusted / metadata is corrupt, and add a regression asserting that a live stale holder returns `skipped_concurrent`.

—

Yeachan-Heo (changes requested):
Verdict: REQUEST_CHANGES

Review scope: PR #2978 at `261546f09b8dbe79a12b21f4d1ff9a66afe17cf5` against `origin/dev` (`5ea96e6cf8163ee3c24c7d497fc3389536262241`).

Inspected:
- `src/hud/reconcile.ts`
- `src/hud/tmux.ts`
- `src/hud/__tests__/reconcile.test.ts`
- `src/hud/__tests__/tmux.test.ts`

Focused checks:
- The previous blocker is not fixed. `acquireReconcileLock` still marks an existing lock stealable when `now() - ts > RECONCILE_LOCK_STALE_MS || !processAlive(pid)`, so lock age alone can steal from a live holder.
- The new regression test only covers a live holder with a fresh timestamp. It does not cover the required stale timestamp + live PID case.
- A direct live-stale probe against the compiled PR head seeded `{ pid: process.pid, ts: 1 }`, used `now: () => 1_000_000`, and `processAlive: () => true`; `reconcileHudForPromptSubmit` returned `recreated`, called pane listing/creation/resizing, and removed the original lock. Expected behavior for a live holder is `skipped_concurrent` with the lock/layout untouched.
- Dead stale lock recovery and non-blocking hook behavior are covered by the targeted tests, but the live-stale steal condition reintroduces the overlapping reconcile race the prior review blocked.

Verification run:
- `git rev-parse HEAD` → `261546f09b8dbe79a12b21f4d1ff9a66afe17cf5`
- `git merge-base HEAD origin/dev` → `5ea96e6cf8163ee3c24c7d497fc3389536262241`
- `npm run build` → passed
- `node dist/scripts/run-test-files.js dist/hud/__tests__/reconcile.test.js dist/hud/__tests__/tmux.test.js` → passed (`reconcile`: 52/52, `tmux`: 48/48)
- Additional direct live-stale lock probe → failed the required invariant: returned `recreated` and mutated layout instead of `skipped_concurrent`.

Required fix: only steal a parseable lock when the holder is not alive, or when stale and holder liveness cannot be trusted because metadata is corrupt/unusable. Add the missing regression for stale timestamp + live PID (including EPERM-equivalent liveness) asserting `skipped_concurrent`, no pane list/create/kill/resize, and lock contents preserved.

—

Yeachan-Heo:
Closing this iteration after the current-head review still reproduced the live-stale lock steal blocker. The linked bug remains tracked in #2977; a follow-up patch needs the stale timestamp + live PID regression and the liveness-aware steal condition described in the review.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2979, opened 2026-06-27
Fix HUD reconcile live stale lock handling

### LINKED ISSUE #2977: [Bug] Managed-mode HUD reconcile deadlocks on tmux wait-for lock and spawns 1000+ stuck processes (v0.18.15) — escalation of #2860

Yeachan-Heo:
## Summary
- Replace blocking tmux wait-for layout reconcile serialization with a non-blocking file lock inside HUD reconciliation.
- Preserve parseable stale locks when the holder PID is still live or liveness is unknown/EPERM-equivalent, returning skipped_concurrent without pane/layout mutation.
- Recover stale reconcile locks only when the recorded holder is dead, and keep existing HUD stale/corrupt recovery paths covered by regression tests.

## Verification
- npm run build
- node dist/scripts/run-test-files.js dist/hud/__tests__/reconcile.test.js dist/hud/__tests__/tmux.test.js

Fixes #2977

—

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -1,7 +1,9 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
+import { existsSync } from 'node:fs';
 import { chmod, mkdtemp, mkdir, readFile, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
+import { dirname } from 'node:path';
 import { join } from 'node:path';
 import { OMX_TMUX_HUD_OWNER_ENV, reconcileHudForPromptSubmit } from '../reconcile.js';
 import { HUD_TMUX_HEIGHT_LINES, HUD_TMUX_ULTRAGOAL_HEIGHT_LINES, HUD_TMUX_MIN_LAUNCH_WINDOW_HEIGHT_LINES } from '../constants.js';
@@ -9,6 +11,18 @@ import { OMX_TMUX_HUD_LEADER_PANE_ENV } from '../tmux.js';
 
 const noOpRegisterHudResizeHook = () => true;
 const noOpUnregisterHudResizeHook = () => true;
+
+async function writeHudReconcileLock(cwd: string, owner: Record<string, unknown>): Promise<string> {
+  const lockPath = join(cwd, '.omx', 'state', 'hud-reconcile.lock');
+  await mkdir(dirname(lockPath), { recursive: true });
+  await mkdir(lockPath, { recursive: true });
+  await writeFile(join(lockPath, 'owner.json'), `${JSON.stringify(owner, null, 2)}\n`);
+  return lockPath;
+}
+
+async function readLockOwner(lockPath: string): Promise<Record<string, unknown>> {
+  return JSON.parse(await readFile(join(lockPath, 'owner.json'), 'utf-8')) as Record<string, unknown>;
+}
 
 describe('reconcileHudForPromptSubmit', () => {
   it('skips reconciliation outside tmux', async () => {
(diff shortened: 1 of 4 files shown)


Yeachan-Heo:
MERGE_READY_LOW_RISK_MAINTAINER_JUDGMENT

Reviewed head SHA: `d8a52caa70e1f4434e50a40791222f88d9194e7d`.

Evidence reviewed:
- Diff against `origin/dev`: `src/hud/reconcile.ts`, `src/hud/tmux.ts`, `src/hud/__tests__/reconcile.test.ts`, `src/hud/__tests__/tmux.test.ts`.
- The reconcile lock now treats a stale lock with a recorded live PID, or unknown/EPERM-equivalent liveness, as active concurrency and returns `skipped_concurrent` before listing panes or touching HUD layout.
- The stale-lock recovery path still reaps only after stale age plus dead PID, and revalidates token/mtime/age after the rename before deleting and reacquiring. Invalid/corrupt/untrusted owner metadata falls back to stale directory mtime and recovers only when the lock directory is actually stale.
- Hook-level `tmux wait-for` was removed, but the process-aware filesystem lock preserves the live-holder invariant that #2978 missed.

Verification:
- GitHub PR checks for this SHA are green: lint, typecheck, build dist artifact, Node test lanes, coverage gate, Ralph persistence gate, and final CI status succeeded; docs/rust/full-source lanes were skipped by CI policy.
- Local `npm run build`: initially failed because dependencies were absent in this review worktree; after `npm ci`, `npm run build` passed.
- Local targeted HUD verification passed: `node dist/scripts/run-test-files.js dist/hud/__tests__/reconcile.test.js dist/hud/__tests__/tmux.test.js` -> 52/52 reconcile tests passed and 48/48 tmux tests passed.
- Additional local adversarial stale-lock probe passed for corrupt owner JSON, invalid/untrusted owner fields, and dead-holder PID recovery.

Merge note: GitHub still reports `mergeStateStatus: BEHIND`, and local `origin/dev` has advanced beyond the PR merge-base. The fix itself is merge-ready, but update the branch and re-run required checks before merge if branch protection or maintainer policy requires up-to-date CI. Issue #2977 can close after this PR is merged, assuming the post-update checks remain green.

PR #2984, opened 2026-06-27
Fix pending PermissionRequest fallback notifications

mdgld:
## Summary
- watch pending `require_escalated` tool calls in `notify-fallback-watcher`
- emit a fallback notification when a PermissionRequest wait crosses a configurable threshold
- add focused regression coverage plus a dedicated verification npm script

## Verification
- `npx @biomejs/biome lint src/scripts/notify-fallback-watcher.ts src/hooks/__tests__/notify-fallback-watcher.test.ts package.json`
- `node --test --test-name-pattern 'pending PermissionRequest' dist/hooks/__tests__/notify-fallback-watcher.test.js`

## Notes
- `npx tsc --pretty false` still reports unrelated pre-existing errors on `src/team/__tests__/tmux-session.test.ts` in this upstream checkout.


Yeachan-Heo:
Thanks for the PR. Initial maintainer gate state:

- This PR currently targets `main`; please retarget it to `dev`. This repo takes contribution PRs through `dev` first.
- After it targets `dev`, fork CI can be approved and we can run the normal maintainer review gate.

No merge review is running yet because the base branch is not reviewable for this repo flow.

—

Yeachan-Heo:
Thanks — this patch targets the right kind of runtime pain, but PRs here need to target `dev`, not `main`. I retargeted this to `dev`; review will happen against that base.

—

br4vesirrobin:
## OMX Codex review loop verdict

Architect verdict: **BLOCK until the two pending-permission fallback correctness issues are applied to the PR branch.**

Reviewer verdict: **Changes requested.** I confirmed the two unresolved Codex threads are valid:
- P1: the synthetic pending-permission notification reuses the real turn id as an `agent-turn-complete` event, which can poison `notify-hook` turn-completion dedupe and suppress the actual completion notification.
- P2: `runPendingPermissionRequestTick()` only looks at the oldest pending request, so an already-notified older request can starve newer pending escalations.

Executor verdict: **Local fix prepared and verified, but not pushed due permissions.** In my PR review worktree I fixed both by:
- using a synthetic `pending-permission-<callId>` turn id while preserving the original turn id as `pending-turn-id`; and
- scanning for the first unnotified eligible pending request instead of returning on the oldest already-notified request.

Verification evidence:
- PASS `npm run verify:notify-fallback-pending-permission` — build passed and pending-permission watcher tests passed (3/3), including the new multiple-pending-request regression.
- PASS `npx tsc --noEmit`.
- PASS `npx biome lint src/scripts/notify-fallback-watcher.ts src/hooks/__tests__/notify-fallback-watcher.test.ts package.json`.
- PASS GitHub CI for current PR head: all listed required lanes are passing; PR Base Guidance is skipped.

GitHub/permission evidence:
- PR is open, non-draft, mergeable, and currently behind `dev`.
- `maintainerCanModify=true`, but attempted dry-run push to `mdgld/oh-my-codex:fix/pending-permission-fallback-watcher` failed with HTTP 403 for the authenticated account, so I did **not** push or merge.
- Unresolved review threads remain on the PR until the branch is updated.

No merge performed.

Yeachan-Heo (changes requested):
REQUEST_CHANGES_CLOSE_EXTERNAL

Reviewed head `b5eb03ab2adb6e134e6d431546932f3ae87644b7` against `origin/dev` (`1967462d11ec92c4f9b0f76c792af490c045ad4f` locally; mergeStateStatus is still `BEHIND`). CI is green, and local targeted verification passed after installing dependencies:

- `npm run verify:notify-fallback-pending-permission` PASS: build succeeded and both `pending PermissionRequest` tests passed.
- `npx @biomejs/biome lint src/scripts/notify-fallback-watcher.ts src/hooks/__tests__/notify-fallback-watcher.test.ts package.json` PASS.

Blocking issue: the fallback notification and persisted watcher state include a command preview derived directly from the escalated tool-call arguments (`args.cmd || args.command`). That means a pending PermissionRequest for something like `curl -H "Authorization: Bearer ..."`, `npm publish --//registry...:_authToken=...`, or any shell command containing credentials can be copied into `last-assistant-message` and `pending_permission_request.command_preview`. This is exactly the sensitive raw/private payload surface the fallback path should avoid. Truncating to 160 chars is not redaction, and state/log/notification fanout can still leak secrets.

The behavior should be changed so the pending-permission fallback reports only non-sensitive metadata (for example tool name + age + maybe a redacted/constant reason), or applies the repo's established secret-redaction policy before storing or notifying any preview. Regression coverage should include a command containing an obvious token/password and assert the notify payload/state/log do not contain it.

Branch update was not attempted because the current head has a blocking privacy issue; the branch is also behind `dev` and will need updating after the fix.

—

Yeachan-Heo:
Closing this iteration after maintainer review.

The current branch still has a blocking payload-hygiene issue in the pending PermissionRequest fallback path: command preview text is derived from escalated tool-call arguments and can end up in notification/state surfaces. Truncation is not redaction, so this cannot be merged as-is.

If you want to continue, please open a fresh PR against `dev` with the fallback notification/state path avoiding raw command/tool-argument previews and with regression coverage for that case.

—

PR #2988, opened 2026-06-28
fix: preserve exact planner and architect contracts (follow-up to dev)

iqdoctor:
This follow-up branch retargets the original fix against `dev`.

### What changed
- Preserved exact planner and architect contracts.
- Kept worker overrides separate from inherited leader model state.
- Kept planner pinned to the intended high model under the `madmax` mini configuration.

### Why
The original PR targeted the wrong base branch. This follow-up carries the same fix onto `dev` so it can merge cleanly.

### Validation
- `npm run build`
- `node dist/scripts/run-test-files.js dist/team/__tests__/runtime.test.js dist/cli/__tests__/index.test.js dist/team/__tests__/model-contract.test.js`

### DIFF
--- a/docs/reference/omx-config-schema-routing.md
+++ b/docs/reference/omx-config-schema-routing.md
@@ -141,7 +141,7 @@ These overrides do not change built-in defaults in source. They are user/project
 For a named role, effective model precedence is:
 
 1. `.omx-config.json` `agentModels[role]`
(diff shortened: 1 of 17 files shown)


iqdoctor:
Superseded by the single cumulative follow-up PR #2989 against dev.

PR #2989, opened 2026-06-28
fix: preserve exact planner and architect contracts (cumulative follow-up to dev)

iqdoctor:
This is the single cumulative follow-up against `dev`.

It combines the exact-model fix work that was previously split across two PRs and keeps the leader/worker model boundary explicit.

### What changed
- Preserved exact planner and architect contracts.
- Kept worker overrides separate from inherited leader model state.
- Retained the planner high-model pin under `madmax` mini.
- Added the extra CLI/runtime coverage needed to keep inherited leader model handling separate from worker env.

### Validation
- `npm run build`
- `node dist/scripts/run-test-files.js dist/cli/__tests__/index.test.js dist/team/__tests__/runtime.test.js dist/team/__tests__/model-contract.test.js dist/agents/__tests__/native-config.test.js dist/agents/__tests__/definitions.test.js dist/utils/__tests__/agents-model-table.test.js`

PR #2990, opened 2026-06-28
Fix exact-role worker CLI resolution in scale-up

iqdoctor:
## Follow-up to PR #2989

This PR is a follow-up to merged PR #2989 on `dev`.
It fixes the scale-up path so exact-role workers recompute their CLI from the final per-worker launch args instead of stale shared launch args.

## What changed
- Recompute worker CLI from the resolved per-worker launch args during scale-up.
- Carry inherited leader model routing into exact-role resolution.
- Extract the exact-role CLI resolver into the shared tmux-session layer so scaling and runtime use the same pure helper.
- Keep the final worker bootstrap aligned with the resolved role contract.
- Add a regression test that covers a scale-down gap / monotonic worker-index case.

## Why
A stale shared-args path could preserve inherited non-Codex model routing too early and select the wrong worker CLI during scale-up.
The previous implementation also conflated worker identity with contiguous lookup position, which could fail after scale-down.

## Verification
- `npm run test:team:worker-runtime-identity`
- `npm run lint`

### DIFF
--- a/src/team/__tests__/worker-runtime-identity.test.ts
+++ b/src/team/__tests__/worker-runtime-identity.test.ts
@@ -288,4 +288,107 @@ process.on('SIGTERM', () => process.exit(0));
       await rm(fakeBinDir, { recursive: true, force: true });
     }
   });
+
+  it('scaleUp recomputes worker CLI from exact-role-resolved launch args instead of inherited non-Codex model routing', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-runtime-identity-scale-exact-'));
+    const fakeBinDir = await mkdtemp(join(tmpdir(), 'omx-runtime-identity-scale-exact-bin-'));
+    const tmuxStubPath = join(fakeBinDir, 'tmux');
+    const tmuxLogPath = join(fakeBinDir, 'tmux.log');
+    const previousPath = process.env.PATH;
+
+    try {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
MERGE_READY

Reviewed exact head 5e8693d3335a987345cac786bed22e12f47c7a04 for the scale-up exact-role worker CLI resolution change.

Evidence:
- PR metadata/head/base checked: non-draft, CLEAN/MERGEABLE, required CI lanes green or intentionally skipped by path lane.
- Diff scope is limited to src/team/scaling.ts and src/team/__tests__/worker-runtime-identity.test.ts.
- scaleUp now resolves launch args per runtimeRole, honors exact-role model selection, then derives worker_cli from those resolved launch args instead of reusing an agentType-wide plan. The resolved worker_cli is consistently used for startup command generation, identity persistence, and pane notification/fallback paths.
- Regression coverage adds a scaleUp case where an architect exact-role worker must stay on codex despite inherited claude model routing, and asserts identity/startup/tmux paths do not route through claude/gemini.
- Local targeted npm run test:team:worker-runtime-identity could not complete in this checkout because build dependency resolution failed on missing @iarna/toml/@modelcontextprotocol packages before the PR test ran; GitHub CI is green and covers the required lanes.

No concrete blocker found.

—

PR #2991, opened 2026-06-28
Follow-up: exact-role scale-up CLI resolver

iqdoctor:
## Follow-up to PR #2990

This PR is a follow-up to merged PR #2990 on `dev`.
It tightens the exact-role scale-up fix so the worker CLI resolver stays shared, the scale-up path stays on the final per-worker launch args, and the regression covers the scale-down gap case.

## What changed
- Recompute worker CLI from the resolved per-worker launch args during scale-up.
- Carry inherited leader model routing into exact-role resolution.
- Extract the exact-role CLI resolver into the shared tmux-session layer so scaling and runtime share one pure helper.
- Keep the final worker bootstrap aligned with the resolved role contract.
- Add a regression test that covers a scale-down gap / monotonic worker-index case.

## Why
A stale shared-args path could preserve inherited non-Codex model routing too early and select the wrong worker CLI during scale-up.
The earlier implementation also conflated worker identity with contiguous lookup position, which could fail after scale-down.

## Verification
- `npm run test:team:worker-runtime-identity`
- `npm run lint`

### DIFF
--- a/src/team/__tests__/worker-runtime-identity.test.ts
+++ b/src/team/__tests__/worker-runtime-identity.test.ts
@@ -335,7 +335,7 @@ process.on('SIGTERM', () => process.exit(0));
         subject: 'architecture follow-up',
         description: 'exact-role scale-up regression',
         status: 'pending',
-        owner: 'worker-2',
+        owner: 'worker-3',
         role: 'architect',
       }, cwd);
 
@@ -345,6 +345,7 @@ process.on('SIGTERM', () => process.exit(0));
       config.tmux_session = 'omx-team-exact-role-cli';
       config.leader_pane_id = '%11';
       config.workers[0]!.pane_id = '%21';
+      config.next_worker_index = 3;
       await saveTeamConfig(config, cwd);
 
       const manifestPath = join(cwd, '.omx', 'state', 'team', 'exact-role-cli', 'manifest.v2.json');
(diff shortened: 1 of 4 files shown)


Yeachan-Heo (commented):
MERGE_READY

Red-team final review at `f227d5cd9db0447f1b579d26473b6e5551622773`: inspected the exact diff and the relevant resolver paths. The resolver move from `runtime.ts` to `tmux-session.ts` is a straight relocation, and the scale-up call now indexes the CLI map against the newly-added worker batch (`i + 1`, `count`) while preserving monotonic worker names via `next_worker_index`. That fixes the exact-role scale-up case where worker names can skip but the CLI map is scoped to the scale-up request rather than the total historical worker index.

The prior concern about `env.OMX_TEAM_WORKER_CLI_MAP` versus `env[OMX_TEAM_WORKER_CLI_MAP_ENV]` is non-blocking: both are equivalent public property access for this literal env key, and error messages still use the constant. Exact-role behavior is covered by the updated worker-runtime identity regression asserting the architect scale-up worker resolves to `codex` despite an inherited Claude model.

Local focused verification was attempted, but this checkout lacks installed dependencies/build artifacts (`@iarna/toml`, `@modelcontextprotocol/sdk`, missing `dist/scripts/run-test-files.js`). Given the already-green GitHub checks, I do not see a blocker.

—

PR #2998, opened 2026-06-28
Fail closed on live root planning pointer conflicts

### LINKED ISSUE #2992: PreToolUse planning-write guard fails open for concurrent OMX sessions sharing one OMX_ROOT
## Summary

The planning-phase **PreToolUse** write guard (deep-interview / ralplan / autopilot-planning) **fails open** for OMX sessions that are *not* the one registered in the root `session.json`, when multiple OMX leader sessions share a single state root (`OMX_ROOT` / `OMX_TEAM_STATE_ROOT`). In that case implementation writes are allowed during a planning phase that should block them.

This is the failure PR #2987 aimed at, via a different mechanism: #2987 closed the `terminalAutopilotRunState` early-return, but the guard can still be skipped earlier, during **session-id resolution**.

> Evidence below is from **code-path + on-disk state analysis**, not a captured failing trace. The on-disk state shows the topology that makes the guard unreachable.

## Mechanism (verified by inspection)

For `PreToolUse`, `dispatchCodexNativeHook` resolves the session, then reads session-scoped active-skill / mode state:

1. `resolveInternalSessionIdForPayload` (`src/scripts/codex-native-hook.ts:3083`) reads **only the single root `session.json`** to translate the native (Codex thread) id → canonical OMX id. It **never consults `subagent-tracking.json`.**
2. When several OMX **leader** sessions share one `OMX_ROOT`, the root `session.json` registers **only one** of them. For any other session, `payloadSessionId` (the Codex thread id) matches neither `session_id`, `native_session_id`, nor `owner_omx_session_id`, so the resolver returns the **raw thread id**.
3. The guard then calls `readVisibleSkillActiveStateForStateDir(stateDir, <raw-thread-id>)`, which builds `sessions/<raw-thread-id>/skill-active-state.json` (`src/state/skill-active.ts:263`). That path **does not exist** — the real state lives under `sessions/<omx-session-id>/`.
4. `readVisibleSkillActiveStateFromPaths` (`src/state/skill-active.ts:340`) returns `null` when the session-scoped file is absent — **no root fallback.**
5. `readActiveDeepInterviewStateForPreToolUse` (`src/scripts/codex-native-hook.ts:3523`) hits `if (!canonicalState) return null;` and the guard emits no block. Same shape in `readActiveRalplanStateForPreToolUse` (`:3572`). Net effect: **write allowed.**

## On-disk evidence (one observed run)

- `OMX_ROOT/.omx/state/session.json` → `session_id: 019f0dae-…`, `cwd: /…/mg3` (one session).
- A concurrent session in the same root: Codex thread `019f0d63-…`, whose real state is under `sessions/omx-1782636081982-ttkolh/` and which `subagent-tracking.json` records as `kind: leader`.
- `subagent-tracking.json` **does contain** the needed mapping: `sessions[omx-1782636081982-ttkolh].threads["019f0d63-…"]`. The resolver simply doesn't read it.

So the data needed to resolve the session correctly is already persisted; the guard just doesn't consult it.

## Proposed fix (the sound one)

In `resolveInternalSessionIdForPayload`, after the root `session.json` checks fail, **consult `subagent-tracking.json`**: find the entry whose `sessions[*].threads[*].thread_id` (or `leader_thread_id`) equals the payload thread id, and return **that entry's own OMX `session_id`** (not the root canonical id). Then the existing session-scoped reads resolve correctly.

### Please avoid these tempting-but-unsound fixes
- **Root-level fallback** in `readVisibleSkillActiveStateFromPaths` / the guards: in a multi-session root the root state belongs to a *different* session, so this makes session A's planning phase block session B's tool calls (cross-session false-blocks).
- **Skipping the `cwd` authority check**: ineffective here (the session *ids* differ regardless of cwd) and it would bind the wrong session as authoritative.
- **Blanket "fail closed"** on unresolved state: would over-block plain Codex sessions that have no workflow. Any fail-closed must be **scoped** to a thread that resolves to a session whose state shows an active interview/ralplan.

## Affected files
- `src/scripts/codex-native-hook.ts` — `resolveInternalSessionIdForPayload`, `readActiveDeepInterviewStateForPreToolUse`, `readActiveRalplanStateForPreToolUse`
- `src/state/skill-active.ts` — `readVisibleSkillActiveStateFromPaths`

## Related
- #2987 (kept planning writes blocked — closed a different escape on the same guard)

Yeachan-Heo:
## Summary
- fail closed during PreToolUse planning-write checks when a different live root `session.json` owner is detected and the current native session cannot be authoritatively resolved
- preserve normal owner-session behavior and allowed planning artifact writes
- add focused regressions for live root pointer mismatch and authoritative owner pass-through

Fixes #2992

## Verification
- `npm ci`
- `npm run build && node --test dist/scripts/__tests__/codex-native-hook.test.js --test-name-pattern='live root session pointer|live-root'`
- `npm run lint`
- `git diff --check -- src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts`

—

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -83,6 +83,22 @@ async function writeNativeMappedSessionState(
   });
 }
 
+async function writeLiveNativeMappedSessionState(
+  cwd: string,
+  stateDir: string,
+  sessionId: string,
+  nativeSessionId: string,
+): Promise<void> {
+  await mkdir(join(stateDir, "sessions", sessionId), { recursive: true });
+  await writeJson(join(stateDir, "session.json"), {
+    session_id: sessionId,
+    native_session_id: nativeSessionId,
+    cwd,
+    pid: process.pid,
+    platform: "darwin",
+  });
+}
+
 async function writeSessionSkillActiveState(
   stateDir: string,
   sessionId: string,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
MERGE_READY

Evidence:
- PR #2998 is open, non-draft, merge state is CLEAN, and there are no PR comments or reviews requesting changes.
- Diff is scoped to PreToolUse planning-write defense-in-depth: it detects a different still-live root `session.json` owner, blocks implementation writes only when the live owner has active deep-interview/ralplan/autopilot-planning state, and preserves authoritative owner/planning-artifact writes.
- This matches #2992's revised scope: fail closed under collapsed root isolation, without adding the demoted `subagent-tracking.json` multi-session resolver.
- CI rollup is green: lint, typecheck, build, Node test lanes, coverage gate, Ralph persistence, and CI Status succeeded; non-applicable docs/Rust lanes are skipped.

—

PR #2999, opened 2026-06-28
Protect deep-interview planning gate state writes

Yeachan-Heo:
## Summary
- deny protected planning gate/phase-control state filenames under `.omx/state` during deep-interview/ralplan gated planning
- keep legitimate planning artifacts writable under `.omx/context`, `.omx/interviews`, `.omx/plans`, `.omx/specs`, and non-control `.omx/state` files
- block `omx state write/clear` mutations during gated planning while preserving state reads/questions

Fixes #2993

## Tests
- `npm run build`
- `node --test dist/scripts/__tests__/codex-native-hook.test.js --test-name-pattern='deep-interview artifact and state writes|deep-interview apply_patch artifact writes|Ralplan read-only inspection'` (Node ran the full file: 465 passing)
- `npm run lint`
- `git diff --check`

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2997, opened 2026-06-28
Fix madmax inherited root isolation

Yeachan-Heo:
## Summary
- Fixes madmax/worktree auto-isolation so stale inherited `OMX_ROOT`, `OMX_STATE_ROOT`, or `OMXBOX_ACTIVE` no longer suppresses a new top-level boxed root.
- Preserves legitimate child-process reuse only when the inherited boxed root metadata and `OMX_MADMAX_DETACHED_CONTEXT` match the current launch context.
- Adds focused regressions for stale env isolation and matching-context child reuse.

Closes #2995.

## Verification
- `npm run build && node --test dist/cli/__tests__/index.test.js --test-name-pattern='madmax state isolation'`
- `git diff --check -- src/cli/index.ts src/cli/__tests__/index.test.ts`
- `npm run lint`

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #2996, opened 2026-06-28
Fix autopilot phase HUD contract

Yeachan-Heo:
Closes #2994.

## Summary
- Documented the Autopilot child-phase state/HUD contract for code-review and ultragoal.
- Keeps Autopilot visible while supervised child phases advance HUD/current-phase state.
- Surfaces keyword-activated ultragoal HUD state before durable goals exist.

## Verification
- build ✅
- lint ✅
- no-unused ✅
- catalog check ✅
- targeted hook/HUD/state tests ✅
  - includes `node dist/scripts/run-test-files.js dist/state/__tests__/operations.test.js dist/state/__tests__/workflow-transition.test.js` (70+11 pass)

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Issue #2995, opened 2026-06-28
madmax/worktree launch silently reuses an inherited OMX_ROOT (isolation skipped on OMXBOX_ACTIVE/OMX_ROOT presence) → sessions collapse into one state root


iqdoctor:
## Summary

A new `omx --madmax` / `--worktree` launch can **silently reuse an inherited `OMX_ROOT`** instead of creating its own isolated state root, collapsing what should be independent per-session roots into one shared root. This violates the intended invariant of **one live OMX session per `OMX_ROOT`** (sequential within a worktree; parallel work = separate `omx --madmax --worktree …` launches, each isolated), and it is the upstream trigger for the planning-guard fail-open (#2992) and the HUD wrong-session phase (#2994).

## Mechanism (verified)

`shouldAutoIsolateMadmaxLaunch` (`src/cli/index.ts`) gates isolation purely on **env-variable presence**:

```ts
if (command !== "launch" && command !== "exec") return false;
if (env.OMX_NO_BOX === "1" || env.OMXBOX_ACTIVE === "1") return false;
if (env.OMX_ROOT || env.OMX_STATE_ROOT) return false;
return launchArgs.some(a => a === MADMAX_FLAG || a === MADMAX_SPARK_FLAG);
```

If the launching shell already carries `OMXBOX_ACTIVE=1` or `OMX_ROOT`/`OMX_STATE_ROOT` — e.g. inherited from a previous `omx --madmax` run in the same terminal/tmux pane — the gate returns `false`, `activateMadmaxIsolationIfNeeded` does **not** mint a fresh `runDir`, and the new session inherits and writes into the **stale** root.

The check cannot distinguish two ontologically different situations:
- a **child process** of an already-isolated session legitimately reusing *its own* root (correct), versus
- a **new top-level session** launched in a shell that merely *inherited* a different run's env (incorrect).

Both look identical to a presence test on `OMXBOX_ACTIVE` / `OMX_ROOT`.

Notably, infrastructure to tell them apart already exists but is not consulted here: `createMadmaxIsolatedRoot` records a per-run `detached_launch_context` (`buildMadmaxDetachedLaunchContextKey`, `OMX_MADMAX_DETACHED_CONTEXT`), and `isMadmaxDetachedGuardEnabled` already pairs `OMXBOX_ACTIVE` with that token. The isolation gate could validate the inherited context against the run-dir's recorded token rather than trusting raw presence.

## Evidence (one observed run)

[...]

PR #3000, opened 2026-06-28
refactor(planning-guard): compact #2997/#2998 helpers + review findings for #2996/#2999

iqdoctor:
## Summary

Follow-up to the four merged planning-guard PRs (#2996 / #2997 / #2998 / #2999, addressing issues #2992–#2995). This PR contains **only behavior-preserving compaction** of the code those PRs introduced, plus a detailed analysis of two deeper coherence problems I found while reviewing the set — which are intentionally **not** changed here because they invert just-merged behavior and need a design decision (see *Findings → recommended follow-up*).

Base: `dev`.

## What this PR changes (compaction, no behavior change)

1. **`src/cli/index.ts` (#2997 cleanup).** Removed the redundant `arg.startsWith("-w=")` clause in `launchArgRequestsDisposableWorktree`. `-w=<name>` is a strict subset of the existing `arg.startsWith("-w") && arg.length > 2` clause, so the separate check could never change the result. Added a one-line comment explaining the coverage.

2. **`src/scripts/codex-native-hook.ts` (#2998 cleanup).** Collapsed the two near-identical root-pointer fail-closed builders (`buildDeepInterviewRootPointerConflictBlock` / `buildRalplanRootPointerConflictBlock`) into a single `buildRootPointerConflictBlock(activeState, planningModeLabel, planningPhaseDescription)`. The two named wrappers are retained so call sites are untouched and the emitted `reason` / `additionalContext` strings are byte-identical to before — existing assertions in `codex-native-hook.test.ts` continue to hold.

### Validation
- `npm run build` green.
- Changed paths pass: `cli/__tests__/index.test.js` isolation/worktree pattern **26/26**, `scripts/__tests__/codex-native-hook.test.js` planning-guard/conflict pattern **96/96**.
- The only failing test in `index.test.js`, `releaseTmuxExtendedKeysLease restores when all remaining holders are dead`, **also fails on clean `origin/dev`** (it depends on PID-liveness semantics not available in this sandbox) and is unrelated to this change.

## Why the merged set works — and where it doesn't (review findings)

The four PRs correctly fix the primary chain: #2997 restores per-launch root isolation (the real root cause of #2995 — a `--madmax`/`--worktree` launch reusing an inherited `OMX_ROOT`), and #2998 adds the fail-closed defense for collapsed roots. #2995 looks closeable on the back of #2997.

However, #2999 (gate-write protection) and #2996 (phase contract) interact in a way that **inverts the safety posture** for the deep-interview / ralplan planning phases. I'm flagging this rather than patching it, because the fix changes the behavior these PRs just established and should be a deliberate design call.

**Finding 1 — the gate-validated transition path is blocked, the gate-bypassing one is open.**
- The state backend `state_write` (`src/state/operations.ts`) enforces the autopilot phase gate on **every** write, for both transports — `omx state write` (CLI) and `omx_state` (MCP) — via `canAdvanceAutopilotDeepInterviewToRalplan` / `canAdvanceAutopilotRalplanToUltragoal` (ordering + completion/skip evidence).
- #2999 blocks the **Bash `omx state write|clear`** path (and raw edits to the protected state files) during deep-interview/ralplan. So the model's *gate-validated* CLI transition is blocked, while the MCP path to the same validated backend stays open — an inconsistent transport asymmetry.
- Meanwhile #2996's keyword-handoff path, `persistAutopilotSupervisedChildPhaseState` (`src/hooks/keyword-detector.ts`), writes `current_phase` **directly via `writeFile`, without invoking the gate**. So advancing by invoking `$ralplan` / `$ultragoal` bypasses exactly the gate that `state_write` enforces.
- Net effect: the protection blocks the *validated* path and leaves the *unvalidated* paths (keyword handoff, MCP) open. A CLI-first autopilot loop (which `codex-native-hook.ts:2034` and `skills/autopilot/SKILL.md` both still recommend, and which must persist `handoff_artifacts`/gate evidence to `autopilot-state.json`) can stall, because the only evidence-persistence path left open during planning is MCP.

**Finding 2 — `deep_interview_gate.skip_authorized_by_user` has no provenance enforcement.**
- `isSkipGate` (`src/autopilot/deep-interview-gate.ts:319`) accepts a skip only when `skip_authorized_by_user === true` (plus reason/timestamp/source/session). But that field is **only ever read** — no trusted code path sets it (repo-wide, the single occurrence is the read at :319). A model can therefore set it itself, so the "user-authorized" requirement is forgeable. #2999's transport block is effectively papering over this on the Bash/file transports while MCP remains open.

### Recommended follow-up (separate, design-reviewed PR)
Resolve both with a single coherent rule instead of transport-level blocks:
1. Make the keyword-handoff transition (`persistAutopilotSupervisedChildPhaseState`) defer to the same `canAdvance…` gate as `state_write`, so no transition path bypasses the gate.
2. Enforce provenance for `skip_authorized_by_user` in `state_write` (only honor a false→true transition that is backed by a real user-authorization record, e.g. a deep-interview `omx question` answer), so a skip cannot be self-attested on any transport.
3. Once 1+2 hold, relax #2999's blanket Bash `omx state write|clear` block (keep the raw-file protection), which restores the CLI-first loop and removes the CLI/MCP asymmetry.

I'm happy to open that as a follow-up PR if the direction looks right; it deliberately isn't bundled here so this cleanup stays low-risk and the design change gets its own review.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -2005,7 +2005,8 @@ function launchArgRequestsDisposableWorktree(arg: string): boolean {
   return arg === "--worktree" ||
     arg === "-w" ||
     arg.startsWith("--worktree=") ||
-    arg.startsWith("-w=") ||
+    // Covers both `-w=<name>` and `-w<name>`; an explicit `-w=` check would be a
+    // strict subset of this clause, so it is omitted as redundant.
     (arg.startsWith("-w") && arg.length > 2);
 }
 
--- a/src/scripts/codex-native-hook.ts
+++ b/src/scripts/codex-native-hook.ts
@@ -3819,34 +3819,36 @@ function blocksDeepInterviewImplementationWrite(payload: CodexHookPayload, cwd: 
     || !candidates.every((candidate) => isAllowedDeepInterviewArtifactPath(cwd, candidate));
 }
 
-function buildDeepInterviewRootPointerConflictBlock(activeState: Record<string, unknown>): Record<string, unknown> {
+// Shared builder for the "live root session pointer owned by another session"
+// fail-closed block. Deep-interview and ralplan/autopilot only differ in the
+// human-readable mode label and phase-description fragment; the decision,
+// structure, and guidance are identical.
+function buildRootPointerConflictBlock(
+  activeState: Record<string, unknown>,
+  planningModeLabel: string,
+  planningPhaseDescription: string,
+): Record<string, unknown> {
   const phase = formatPhase(activeState.current_phase ?? activeState.currentPhase, "planning");
-  return {
-    decision: "block",
-    reason: `Deep-interview is active in the live root session pointer (phase: ${phase}), but the current native session could not be authoritatively resolved to that owner; failing closed for planning-write protection.`,
-    hookSpecificOutput: {
-      hookEventName: "PreToolUse",
-      additionalContext:
-        "OMX detected a live root session pointer owned by another session while a deep-interview planning phase is active. "
-        + "This indicates collapsed session-root isolation. Do not perform implementation writes from this unresolved session; use the owning OMX session or restart with an isolated OMX_ROOT.",
-    },
-  };
-}
-
-function buildRalplanRootPointerConflictBlock(activeState: Record<string, unknown>): Record<string, unknown> {
-  const phase = formatPhase(activeState.current_phase ?? activeState.currentPhase, "planning");
-  const activeMode = safeString(activeState.mode).trim().toLowerCase();
-  const planningModeLabel = activeMode === "autopilot" ? "Autopilot planning" : "Ralplan";
   return {
     decision: "block",
     reason: `${planningModeLabel} is active in the live root session pointer (phase: ${phase}), but the current native session could not be authoritatively resolved to that owner; failing closed for planning-write protection.`,
     hookSpecificOutput: {
       hookEventName: "PreToolUse",
       additionalContext:
-        "OMX detected a live root session pointer owned by another session while a ralplan/autopilot planning phase is active. "
+        `OMX detected a live root session pointer owned by another session while a ${planningPhaseDescription} is active. `
         + "This indicates collapsed session-root isolation. Do not perform implementation writes from this unresolved session; use the owning OMX session or restart with an isolated OMX_ROOT.",
     },
   };
+}
+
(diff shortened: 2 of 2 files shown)


Yeachan-Heo (commented):
MERGE_READY

Adversarial review evidence:

- Changed surface is limited to `src/cli/index.ts` and `src/scripts/codex-native-hook.ts`; PR head is `d4cab9b3f37adb2b046294f98e96af65403cbd6b` on base `dev`.
- CLI predicate check: the removed `arg.startsWith("-w=")` branch is a strict subset of the retained `(arg.startsWith("-w") && arg.length > 2)` clause. Literal matrix verified after fresh build:
  - `-w=foo` => worktree enabled, named `foo`; subset proof true.
  - `-wfoo` => worktree enabled, named `foo`; handled by retained short `-w` clause.
  - `-w` and `--worktree` => detached worktree.
  - `--worktree=foo` => named worktree `foo`.
  - non-worktree flags such as `--model gpt-5`, `-h`, `--help` remain non-worktree for `parseWorktreeMode`.
  - `--` behavior is unchanged by this PR; `parseWorktreeMode(["--", "-w=foo"])` still consumes the `-w=foo` token after `--`, which is pre-existing and not introduced by this removal.
- Root-pointer conflict builder check: current shared helper expands to the same deep-interview strings and the same ralplan/autopilot strings as `dev` for `decision`, `reason`, `hookSpecificOutput.hookEventName`, and `hookSpecificOutput.additionalContext`. The mode-specific label (`Deep-interview`, `Ralplan`, `Autopilot planning`) and phase descriptions (`deep-interview planning phase`, `ralplan/autopilot planning phase`) are preserved.
- Focused local verification passed after `npm ci`: `npm run build && node dist/scripts/run-test-files.js dist/team/__tests__/worktree.test.js dist/cli/__tests__/index.test.js dist/scripts/__tests__/codex-native-hook.test.js` => 467 hook/CLI tests passed plus 18 worktree tests passed.
- GitHub CI re-check is green: build, typecheck, lint, smoke, cli-core-rest, hooks-notify-platform, team-state-runtime, coverage gate, Ralph persistence gate, and CI Status all pass. Skipped Rust/docs/full-source lanes are unchanged lane gating.

No PR-blocking regression, contract drift, missing-test blocker, planning-guard fail-open/fail-closed error, or interaction issue with the recently merged planning-guard PRs was found. Do not merge from this review comment; this is review evidence only.

—

Issue #2993, opened 2026-06-28
deep-interview gate/phase-control state is model-writable during the gated phase (planning write-prefix too broad)


iqdoctor:
## Summary

The deep-interview PreToolUse guard treats the entire `.omx/state` prefix as a writable "planning artifact" location (`DEEP_INTERVIEW_ALLOWED_WRITE_PREFIXES`, `src/scripts/codex-native-hook.ts:3131`). But that prefix also contains the **phase-control state the guard itself depends on** — `autopilot-state.json` (`deep_interview_gate`, `current_phase`), `skill-active-state.json`, `deep-interview-state.json`. So the model can persist its own gate transition while still in the gated phase, making the guard's enabling condition **self-mutable**.

> This is a **hardening** request, not "the model cheating." Writing a `deep_interview_gate.skip_reason` is a *designed* transition in the autopilot protocol (clear bounded tasks may skip with an explicit persisted reason). The concern is narrower: a phase **boundary** is enforced by state that the gated actor can rewrite, under an allow-listed path, with no separate authorization.

## Detail

```ts
// src/scripts/codex-native-hook.ts:3131
const DEEP_INTERVIEW_ALLOWED_WRITE_PREFIXES = [
  ".omx/context",
  ".omx/interviews",
  ".omx/specs",
  ".omx/state",     // <-- also holds gate/phase-control files
] as const;
```

`autopilot-state.json` lives at `.omx/state/sessions/<id>/autopilot-state.json` — under `.omx/state`, hence an allowed write during deep-interview. Observed in one run:

```json
"deep_interview_gate": {
  "status": "skipped",
  "skip_reason": "…we can inspect and fix locally without further requirement clarification."
}
```

The same breadth applies to the ralplan prefix list (`RALPLAN_ALLOWED_WRITE_PREFIXES`).

## Proposed options (pick one)

1. **Narrow the prefix**: allow `.omx/state/sessions` *metadata* and `.omx/state/session.json`, but **exclude** phase-control files (`autopilot-state.json`, `skill-active-state.json`, `deep-interview-state.json`, `ralplan-state.json`) from model-writable planning prefixes.
2. **Explicit deny-list**: keep `.omx/state` allowed but deny the phase-control filenames above in the guard.
3. **Route gate transitions through a tool**, not a raw file write: require `omx state`/an explicit handoff command to set `deep_interview_gate.skipped`, so the transition is observable and policy-checked rather than a side-effect of an allowed write.

[...]

PR #3001, opened 2026-06-28
fix(planning-gate): single-source phase-gate enforcement across CLI/MCP/keyword transports

iqdoctor:
## Summary

Makes Autopilot planning-phase gate enforcement **single-source and transport-agnostic**. The merged #2999/#2996 pair inadvertently inverted the safety posture: the *gate-validated* transition path was blocked during deep-interview/ralplan, while the *gate-bypassing* paths stayed open. This PR fixes that with two coordinated changes, keeps the genuine protection from #2999, and is deadlock-free by construction.

Base: `dev` (on top of #3000).

## The problem (from reviewing #2992–#2999)

The `state_write` backend (`src/state/operations.ts`) already enforces the Autopilot phase gate on **every** write — for both `omx state write` (CLI) and `omx_state` (MCP) — via `canAdvanceAutopilotDeepInterviewToRalplan` / `canAdvanceAutopilotRalplanToUltragoal` (ordering + completion/skip evidence). Against that backdrop:

- **#2999** blocked the **Bash `omx state write|clear`** transport during planning. But the equivalent **MCP** path to the same validated backend stayed open (asymmetry), the block was **redundant** with the backend gate, and it **broke the CLI-first loop**, which must persist `handoff_artifacts`/gate evidence and advance phases via `omx state write`.
- **#2996** added a keyword-handoff phase write, `persistAutopilotSupervisedChildPhaseState` (`src/hooks/keyword-detector.ts`), which writes `current_phase` **directly via `writeFile`, without the gate**. So advancing by emitting `$ralplan` / `$ultragoal` **bypasses** exactly the gate `state_write` enforces.

Net effect: the system blocked the *validated* path and left the *unvalidated* ones (keyword handoff, MCP) open — i.e. `$ralplan` alone could skip the deep-interview gate (a re-opening of #2993 via a different door).

### Part A — defer `omx state` mutations to the gate, not the transport
`src/scripts/codex-native-hook.ts`

`isAllowedDeepInterviewBashWrite` / `isAllowedRalplanBashWrite` no longer block `omx state write|clear`; they defer to the gate-enforcing `state_write` backend (identical enforcement for CLI and MCP). The reverted `commandHasDeepInterviewWriteIntent` clause and the "not model-writable" detail message go with it.

**Kept:** the protected-file rule (`isProtectedPlanningStatePath` / `PROTECTED_PLANNING_STATE_FILE_NAMES`) still rejects raw `Write`/`Edit`/Bash-redirects to `autopilot-state.json`, `deep-interview-state.json`, `ralplan-state.json`, `skill-active-state.json` — those bypass the backend gate and are the genuinely unvalidated vector.

### Part B — gate the keyword handoff
`src/hooks/keyword-detector.ts`

`persistAutopilotSupervisedChildPhaseState` now routes its `current_phase` advance through a new `resolveGatedSupervisedChildPhase`, which evaluates the **same** `canAdvance…` gates as `state_write` for the two gated boundaries (deep-interview→ralplan, ralplan→ultragoal). If the gate is not satisfied, the current phase is held (no advance), so a `$child` keyword cannot skip the gate. Non-gated transitions are unaffected.

### Why it can't deadlock
Part A re-allows the gated `omx state write`, so the leader can persist completion/skip evidence during planning; once persisted, the advance gate passes on every transport (CLI, MCP, and now the keyword handoff). Before this PR, a CLI-first/non-MCP loop could only advance by *bypassing* the gate via the keyword path.

## Result
- Every phase-advance transport (CLI, MCP, keyword) enforces the **same** gate.
- Raw-file protection from #2999 preserved.
- CLI/MCP asymmetry and the CLI-first stall removed.

## Out of scope (needs a product decision)
`deep_interview_gate.skip_authorized_by_user` has **no provenance enforcement**: it is read in `src/autopilot/deep-interview-gate.ts` (the only occurrence repo-wide) but never set by trusted code, so a model can set it itself and satisfy `isSkipGate` on any transport. Closing this requires a genuine user-authorization channel (e.g. an `omx question` answer that the model can't fabricate) and a decision on how users authorize a deep-interview skip. Happy to follow up once the direction is set.

## Validation
- `npm run build` green.
- `scripts/__tests__/codex-native-hook.test.js`: **467/467** (the two #2999 tests that asserted the Bash-transport block were updated to assert the corrected behavior — `omx state write` allowed/deferred, protected raw-file writes still blocked).
- New test in `hooks/__tests__/keyword-detector.test.js`: asserts `$ralplan` does **not** advance `current_phase` past an unsatisfied deep-interview gate.
- Pre-existing failures in `state` / `modes` / `autopilot` / `cli` suites **reproduce identically on clean `origin/dev`** (local model-config classification + sandbox PID semantics) and are unrelated to this change.

> Note: an end-to-end Autopilot loop run (with and without MCP) is worth doing before merge — I validated the pieces at unit level but can't drive the live loop here.

### DIFF
--- a/src/autopilot/fsm.ts
+++ b/src/autopilot/fsm.ts
@@ -1,4 +1,4 @@
-const AUTOPILOT_CHILD_PHASES = [
+export const AUTOPILOT_CHILD_PHASES = [
   'deep-interview',
   'ralplan',
   'ultragoal',
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -793,6 +793,42 @@ describe('keyword detector skill-active-state lifecycle', () => {
     }
   });
 
+  it('does not let a keyword jump ahead past a planning gate (forward skip)', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-skip-ahead-'));
+    const stateDir = join(cwd, '.omx', 'state');
+    try {
+      await mkdir(stateDir, { recursive: true });
+      await recordSkillActivation({
+        stateDir,
+        text: 'please run $autopilot',
+        sessionId: 'sess-skip',
+        threadId: 'thread-skip',
+        turnId: 'turn-1',
+        nowIso: '2026-02-25T00:00:00.000Z',
+      });
+
+      // `$ultragoal` while in deep-interview skips the ralplan gate entirely; the
+      // keyword handoff must hold the current phase rather than jump ahead.
+      const result = await recordSkillActivation({
+        stateDir,
+        text: 'jump straight to $ultragoal',
+        sessionId: 'sess-skip',
+        threadId: 'thread-skip',
+        turnId: 'turn-2',
+        nowIso: '2026-02-25T00:01:00.000Z',
+      });
+
+      const autopilotState = JSON.parse(
+        await readFile(join(stateDir, 'sessions', 'sess-skip', 'autopilot-state.json'), 'utf-8'),
+      ) as { current_phase: string };
+      assert.equal(autopilotState.current_phase, 'deep-interview', 'forward skip must be held');
+      // skill-active phase is kept in sync with the held autopilot phase.
+      assert.equal(result?.phase, 'deep-interview');
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
   it('seeds dedicated planner routing in Autopilot state when main is cheap', async () => {
     const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-autopilot-planner-routing-'));
     const stateDir = join(cwd, '.omx', 'state');
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -773,7 +773,7 @@ describe('keyword detector skill-active-state lifecycle', () => {
       // A bare `$ralplan` keyword handoff must not advance current_phase across the
       // deep-interview -> ralplan gate while the gate is unsatisfied. The keyword
       // path now defers to the same gate as the state_write backend.
-      await recordSkillActivation({
+      const denied = await recordSkillActivation({
         stateDir,
         text: 'continue with $ralplan',
         sessionId: 'sess-gate',
@@ -788,6 +788,17 @@ describe('keyword detector skill-active-state lifecycle', () => {
         'deep-interview',
         'keyword handoff must not skip the deep-interview gate',
       );
+
+      // The visible canonical skill-active state must stay aligned with the held
+      // detail phase (no drift to ralplan) — both the returned state and the
+      // persisted skill-active-state.json mirror.
+      assert.equal(denied?.phase, 'deep-interview');
+      assert.deepEqual(denied?.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'deep-interview']]);
+      const canonical = JSON.parse(
+        await readFile(join(stateDir, 'sessions', 'sess-gate', SKILL_ACTIVE_STATE_FILE), 'utf-8'),
+      ) as { phase?: string; active_skills?: Array<{ skill?: string; phase?: string }> };
+      assert.equal(canonical.phase, 'deep-interview');
+      assert.deepEqual(canonical.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'deep-interview']]);
     } finally {
       await rm(cwd, { recursive: true, force: true });
     }
@@ -824,6 +835,56 @@ describe('keyword detector skill-active-state lifecycle', () => {
       assert.equal(autopilotState.current_phase, 'deep-interview', 'forward skip must be held');
       // skill-active phase is kept in sync with the held autopilot phase.
       assert.equal(result?.phase, 'deep-interview');
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
+  it('does not advance or drift canonical state past an unsatisfied ralplan -> ultragoal gate', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-ralplan-gate-'));
+    const stateDir = join(cwd, '.omx', 'state');
+    const sessionId = 'sess-ralplan-gate';
+    try {
+      await mkdir(join(stateDir, 'sessions', sessionId), { recursive: true });
+      // Supervised Autopilot already in ralplan, with no ralplan consensus evidence
+      // (the ralplan -> ultragoal gate is unsatisfied).
+      await writeFile(
+        join(stateDir, 'sessions', sessionId, SKILL_ACTIVE_STATE_FILE),
+        JSON.stringify({
+          version: 1,
+          active: true,
+          skill: 'autopilot',
+          keyword: '$autopilot',
+          phase: 'ralplan',
+          source: 'keyword-detector',
+          session_id: sessionId,
+          active_skills: [{ skill: 'autopilot', phase: 'ralplan', active: true, session_id: sessionId }],
+        }, null, 2),
+      );
+      const autopilotStatePath = join(stateDir, 'sessions', sessionId, 'autopilot-state.json');
+      await writeFile(
+        autopilotStatePath,
+        JSON.stringify({ active: true, mode: 'autopilot', current_phase: 'ralplan', session_id: sessionId }, null, 2),
+      );
+
+      const denied = await recordSkillActivation({
+        stateDir,
+        text: 'advance to $ultragoal',
+        sessionId,
+        threadId: 'thread-ralplan-gate',
+        turnId: 'turn-ralplan-gate',
+        nowIso: '2026-02-25T00:01:00.000Z',
+      });
+
+      const afterAdvance = JSON.parse(await readFile(autopilotStatePath, 'utf-8')) as { current_phase: string };
+      assert.equal(afterAdvance.current_phase, 'ralplan', 'keyword handoff must not skip the ralplan -> ultragoal gate');
+      assert.equal(denied?.phase, 'ralplan');
+      assert.deepEqual(denied?.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'ralplan']]);
+      const canonical = JSON.parse(
+        await readFile(join(stateDir, 'sessions', sessionId, SKILL_ACTIVE_STATE_FILE), 'utf-8'),
+      ) as { phase?: string; active_skills?: Array<{ skill?: string; phase?: string }> };
+      assert.equal(canonical.phase, 'ralplan');
+      assert.deepEqual(canonical.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'ralplan']]);
     } finally {
       await rm(cwd, { recursive: true, force: true });
     }
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -890,6 +890,56 @@ describe('keyword detector skill-active-state lifecycle', () => {
     }
   });
 
+  it('does not let an implementation-phase keyword skip the code-review gate to ultraqa', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-ultraqa-skip-'));
+    const stateDir = join(cwd, '.omx', 'state');
+    const sessionId = 'sess-ultraqa-skip';
+    try {
+      await mkdir(join(stateDir, 'sessions', sessionId), { recursive: true });
+      // Supervised Autopilot in an implementation phase (ultragoal). The completion
+      // gate requires code-review before ultraqa.
+      await writeFile(
+        join(stateDir, 'sessions', sessionId, SKILL_ACTIVE_STATE_FILE),
+        JSON.stringify({
+          version: 1,
+          active: true,
+          skill: 'autopilot',
+          keyword: '$autopilot',
+          phase: 'ultragoal',
+          source: 'keyword-detector',
+          session_id: sessionId,
+          active_skills: [{ skill: 'autopilot', phase: 'ultragoal', active: true, session_id: sessionId }],
+        }, null, 2),
+      );
+      const autopilotStatePath = join(stateDir, 'sessions', sessionId, 'autopilot-state.json');
+      await writeFile(
+        autopilotStatePath,
+        JSON.stringify({ active: true, mode: 'autopilot', current_phase: 'ultragoal', session_id: sessionId }, null, 2),
+      );
+
+      const denied = await recordSkillActivation({
+        stateDir,
+        text: 'run $ultraqa now',
+        sessionId,
+        threadId: 'thread-ultraqa-skip',
+        turnId: 'turn-ultraqa-skip',
+        nowIso: '2026-02-25T00:01:00.000Z',
+      });
+
+      const afterAdvance = JSON.parse(await readFile(autopilotStatePath, 'utf-8')) as { current_phase: string };
+      assert.equal(afterAdvance.current_phase, 'ultragoal', 'keyword handoff must not skip the code-review gate');
+      assert.equal(denied?.phase, 'ultragoal');
+      assert.deepEqual(denied?.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'ultragoal']]);
+      const canonical = JSON.parse(
+        await readFile(join(stateDir, 'sessions', sessionId, SKILL_ACTIVE_STATE_FILE), 'utf-8'),
+      ) as { phase?: string; active_skills?: Array<{ skill?: string; phase?: string }> };
+      assert.equal(canonical.phase, 'ultragoal');
+      assert.deepEqual(canonical.active_skills?.map((entry) => [entry.skill, entry.phase]), [['autopilot', 'ultragoal']]);
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
   it('seeds dedicated planner routing in Autopilot state when main is cheap', async () => {
     const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-autopilot-planner-routing-'));
     const stateDir = join(cwd, '.omx', 'state');
--- a/src/hooks/__tests__/keyword-detector.test.ts
+++ b/src/hooks/__tests__/keyword-detector.test.ts
@@ -748,6 +748,51 @@ describe('keyword detector skill-active-state lifecycle', () => {
     }
   });
 
+  it('does not advance the supervised child phase past an unsatisfied deep-interview gate via keyword handoff', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-keyword-supervised-gate-'));
+    const stateDir = join(cwd, '.omx', 'state');
+    try {
+      await mkdir(stateDir, { recursive: true });
+
+      // Activate Autopilot: seeds autopilot-state.json at current_phase=deep-interview
+      // with deep_interview_gate.status="required" (gate not satisfied).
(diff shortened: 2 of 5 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

Adversarial review finding: this PR closes the direct `autopilot-state.json` phase advance, but denied keyword handoffs still corrupt the visible canonical skill-active phase.

What I verified:
- Base includes the recent planning/runtime fixes 2996/2997/2998/2999/3000 (`dev` at `eae4b3988`; PR head `de0d2e85489d07f48c623078cbefde6bb80e6bcf`).
- Removing the Bash transport block for `omx state write|clear` looks safe in the tested path: the native hook now allows the CLI transport while still blocking raw writes to protected mode/canonical files, and the state backend rejects gated Autopilot phase advances for CLI/MCP state_write. The focused tests cover this in deep-interview and ralplan PreToolUse plus state operations/MCP state-server.
- `skip_authorized_by_user` remains in the existing backend gate path and I did not see this PR broaden it beyond the documented out-of-scope behavior.
- Fresh GitHub CI re-check: mergeStateStatus is CLEAN; all required CI/check runs visible in `gh pr view` are SUCCESS or intentionally SKIPPED as of 2026-06-28T18:50:47Z.

Blocker:
- `src/hooks/keyword-detector.ts` computes `effectivePhase` through the gate for the detail state (`current_phase: effectivePhase`, lines around 1149-1165), but the visible skill-active state is built earlier/later with `phase: match.skill` and `active_skills[].phase: match.skill` (around lines 1404-1429). When the gate denies `deep-interview -> ralplan`, the test only asserts `autopilot-state.json.current_phase` remains `deep-interview` (test around lines 773-790); it does not assert the canonical `skill-active-state.json` mirror.
- I reproduced the resulting drift after build with a minimal prompt-submit sequence: `$autopilot`, then denied `$ralplan`. Result: `autopilot-state.json.current_phase` stayed `deep-interview`, but `skill-active-state.json.phase` and `active_skills[0].phase` were both rewritten to `ralplan`, with `supervised_child_skill: ralplan`.
- That violates the stated safety contract that denied advancement preserves existing phase without corrupting state. It also creates user-facing/runtime ambiguity for HUD/status/list-active/prompt guidance consumers that read canonical skill-active state rather than the detail mode file.

Required fix before merge:
- On denied supervised keyword handoff, keep the visible canonical Autopilot phase aligned with the authoritative held phase, or return a transition error without mutating canonical phase fields.
- Add regression coverage that asserts both `autopilot-state.json.current_phase` and `skill-active-state.json` / `active_skills[].phase` remain unchanged after denied `deep-interview -> ralplan` and denied `ralplan -> ultragoal` keyword handoffs.

Local verification run:
- `npm install` to restore absent deps.
- `npm run build` passed.
- `node dist/scripts/run-test-files.js dist/hooks/__tests__/keyword-detector.test.js dist/scripts/__tests__/codex-native-hook.test.js dist/state/__tests__/operations.test.js dist/mcp/__tests__/state-server.test.js` passed: 737 focused tests, 0 failures.

Do not merge until the canonical visible-state drift is fixed.
—

iqdoctor:
### Addressed self-review findings (high-effort /code-review)

A recall-biased review of this PR surfaced 4 issues, now fixed in `71b2edb0`:

- **F2 (was a regression in this PR):** deferring *all* `omx state write|clear` was too broad — the `state_write` backend only gates `mode==='autopilot'` transitions, not standalone deep-interview/ralplan deactivation, so a model could self-exit a standalone planning phase. Added `commandEndsPlanningPhase`: `omx state clear` and any `omx state write` setting `active:false` are rejected at the transport boundary; non-deactivating writes still defer to the backend gate.
- **F1:** the `omx state`/`omx question` allowance short-circuited the whole command, so a chained implementation write (`printf x > src/evil.ts && omx state read`) slipped through. A disallowed chained target is now rejected before the allowance.
- **F6:** `resolveGatedSupervisedChildPhase` only gated the two adjacent boundaries, so `$ultragoal` from deep-interview skipped the gate. Now holds the current phase on any forward jump past the immediate next gated phase (mirrors the backend's non-adjacent rule).
- **F4:** on a gate-held advance, skill-active phase showed the requested phase while autopilot-state held the prior one; persist/reconcile now return the effective phase and the keyword caller uses it, keeping both stores in sync.

**Known residual (backend follow-up):** `commandEndsPlanningPhase` is transport-level defense-in-depth and is inherently heuristic. Robustly gating standalone deep-interview/ralplan deactivation/skip — and closing the forgeable `deep_interview_gate.skip_authorized_by_user` (read at deep-interview-gate.ts, never set by trusted code) — belongs in the `state_write` backend and needs a user-authorization channel + design decision.

Tests: codex-native-hook 467/467 (added deactivation-blocked / chained-write-blocked / non-deactivating-allowed cases); new keyword tests for forward-skip hold and skill-active sync.

Yeachan-Heo (changes requested):
Terminal verdict: REQUEST_CHANGES

Re-reviewed latest head `71b2edb0d0ac74ca0e85afda0258d0afd0a14823`; not merged.

What is fixed:
- The prior blocker for denied `$ralplan` handoff is fixed for the deep-interview gate. `persistAutopilotSupervisedChildPhaseState` now returns the effective held phase, and `recordSkillActivation` writes that same `effectivePhase` into visible `skill-active-state` (`src/hooks/keyword-detector.ts:1148-1158`, `src/hooks/keyword-detector.ts:1428-1473`).
- Local smoke verified denied `$ralplan` from Autopilot deep-interview keeps all three visible stores at `deep-interview`: return phase, `autopilot-state.current_phase`, and `skill-active-state.phase`/active skill phase.

Remaining blocker:
- Keyword handoff still bypasses another canonical Autopilot gate. From active Autopilot `current_phase: "ultragoal"`, a prompt containing `$ultraqa` advances both `autopilot-state.json.current_phase` and visible `skill-active-state.json.phase` directly to `ultraqa`.
- That conflicts with the backend completion gate contract: implementation phases may not jump to `ultraqa`; they must go through `code-review` first (`src/autopilot/completion-gate.ts:128-158`).
- I reproduced this locally on the built latest head:

```json
{
  "resultPhase": "ultraqa",
  "autopilotPhase": "ultraqa",
  "skillPhase": "ultraqa",
  "activeSkillPhase": "ultraqa"
}
```

Suggested fix:
- Make the keyword handoff path use the same semantic Autopilot transition validation as `state_write`, especially `validateAutopilotCompletionTransition(existing, nextState)`, and hold/return the existing phase on denial.
- Add regression coverage for implementation-phase `$ultraqa` skips, asserting both Autopilot detail state and visible skill-active state remain held.

Verification run:
- `npm run build`
- `node dist/scripts/run-test-files.js dist/hooks/__tests__/keyword-detector.test.js dist/scripts/__tests__/codex-native-hook.test.js` → 170 keyword-detector tests + 467 codex-native-hook tests passed.
- Final GitHub check refresh: head `71b2edb0d0ac74ca0e85afda0258d0afd0a14823`, `mergeStateStatus: CLEAN`, 19 checks completed, no failing checks.

—

iqdoctor:
@Yeachan-Heo thanks — the canonical skill-active drift you reproduced is fixed. Your review was on `de0d2e85`; the fix landed just after in `71b2edb0`, and `951c3844` now adds the regression coverage you required.

**Fix (the "keep aligned" option you offered):** `persistAutopilotSupervisedChildPhaseState` now returns the gate-held `effectivePhase`, `reconcileAutopilotSupervisedChildModeStates` propagates it, and the keyword caller builds the visible skill-active state from `effectivePhase` (both `phase` and `active_skills[].phase`) instead of `match.skill`. So a denied advance no longer rewrites the canonical phase.

**Regression coverage (both boundaries, both stores), as requested:**
- Denied `deep-interview -> ralplan`: asserts `autopilot-state.json.current_phase`, the returned state, and the persisted `skill-active-state.json` (`phase` + `active_skills[].phase`) all stay `deep-interview`.
- Denied `ralplan -> ultragoal` (new test, seeded ralplan with no consensus evidence): asserts `autopilot-state.json.current_phase` and the `skill-active-state.json` mirror both stay `ralplan`.

I re-ran your exact reproduction (`$autopilot` then denied `$ralplan`): `skill-active-state.json.phase` and `active_skills[0].phase` now stay `deep-interview` (no drift to ralplan).

Keyword suite: 170/171 (the one failure, `writes skill-active-state.json with deep-interview phase when autopilot keyword activates`, reproduces on clean `origin/dev` — it asserts planner-routing model classification and picks up the local model config; unrelated to this change).

Two items I deliberately kept out of scope and documented for a backend follow-up (not regressions in this PR): robust standalone deep-interview/ralplan deactivation gating (the hook's `commandEndsPlanningPhase` is transport-level defense-in-depth), and the forgeable `skip_authorized_by_user`. Both need a user-authorization channel in `state_write`.

iqdoctor:
@Yeachan-Heo fixed in `e100206c` — thanks for catching the second skip.

You were right that the keyword path needed the **same semantic validation** as `state_write`, not a positional heuristic. `resolveGatedSupervisedChildPhase` now calls `validateAutopilotCompletionTransition(existing, nextState)` (the same function operations.ts uses) and holds the current phase on any error. That covers the implementation-phase → ultraqa skip you reproduced (completion-gate.ts:153-158), plus the other completion-gate rules, as one source of truth.

I verified your exact repro on the new head: from active Autopilot `current_phase: "ultragoal"`, `$ultraqa` now holds — `autopilot-state.current_phase`, return phase, `skill-active-state.phase`, and `active_skills[0].phase` all stay `ultragoal`. The legitimate `ultragoal -> code-review` advance still proceeds (that transition returns null), and the existing code-review supervised-child test still passes.

Added regression coverage as requested: implementation-phase `$ultraqa` skip asserts both the Autopilot detail state and the visible skill-active mirror remain held at `ultragoal`.

Suites: keyword-detector 171/172 and codex-native-hook 467/467 (the single keyword failure, `writes skill-active-state.json with deep-interview phase when autopilot keyword activates`, reproduces on clean `origin/dev` — planner-routing model classification picking up local model config; unrelated).

iqdoctor:
@codex review

Targeted audit request — please focus on **transport/path parity for Autopilot phase-gate enforcement**, since two skip vectors were already found and fixed here:

1. **Keyword-handoff vs state_write parity.** `resolveGatedSupervisedChildPhase` in `src/hooks/keyword-detector.ts` now gates supervised `$child` keyword advances using `canAdvanceAutopilotDeepInterviewToRalplan`, `canAdvanceAutopilotRalplanToUltragoal`, `validateAutopilotCompletionTransition`, and a positional `isForwardChildPhaseSkip` (deep-interview/ralplan only). The `state_write` backend (`src/state/operations.ts`, the `mode === 'autopilot'` block ~lines 503-595) applies its own sequence. **Is there any current_phase transition the backend would reject that the keyword path still allows, or any legitimate transition the keyword path now wrongly holds?** Please enumerate concrete (currentPhase → requestedChildSkill) pairs and say which side diverges. Consider rework/team/ralph and waiting-for-user.

2. **Standalone (non-autopilot) deactivation.** `commandEndsPlanningPhase` in `src/scripts/codex-native-hook.ts` blocks `omx state clear` and `omx state write` with `active:false` at the transport boundary, but I documented it as heuristic. **Name any other single `omx state write` payload that, during an active standalone deep-interview/ralplan PreToolUse boundary, would end/neutralize the phase (e.g. a terminal `current_phase`/`status`, or removing the active marker) and slip past `commandEndsPlanningPhase` while `state_write` does not re-gate it.**

Please give concrete reproducible inputs and the exact file:line, not general advice. If you find nothing on either point, say so explicitly per point.

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Re-reviewed latest head `e100206cf62327e96dcbcc74383b6efcd7b5caba` after the `951c3844` and `e100206c` follow-ups. CI is now green, but there is still a transport-guard bypass in the `omx state` allowance, so I’m keeping this held.

What looks fixed:
- The keyword handoff path now reuses the same semantic gates for the earlier blockers: `resolveGatedSupervisedChildPhase` calls `canAdvanceAutopilotDeepInterviewToRalplan`, `canAdvanceAutopilotRalplanToUltragoal`, and `validateAutopilotCompletionTransition` before writing the effective phase.
- The canonical mirror drift is addressed: the keyword caller writes `phase` and `active_skills[].phase` from the gate-held `effectivePhase`, not from the requested keyword.
- Regression coverage now covers denied deep-interview → ralplan, ralplan → ultragoal, and ultragoal → ultraqa keyword handoffs, including the persisted `skill-active-state.json` mirror.
- The direct chained write example (`printf 'x' > src/evil.ts && omx state read --json`) is covered.

Remaining blocker:
- `commandEndsPlanningPhase()` only recognizes `omx state clear` and inline `omx state write` payloads containing `active:false`. But `omx state` explicitly supports `--input-file`, and the hook does not inspect that file. During deep-interview/ralplan guarded Bash, a model can run a command like:
  - `omx state write --input-file /tmp/deactivate.json --json`
  - where `/tmp/deactivate.json` contains `{"mode":"deep-interview","active":false}` or the analogous ralplan/autopilot deactivation payload.
- That command has no redirect/tee/sed/git/npm write intent, so `isAllowedDeepInterviewBashWrite()` / `isAllowedRalplanBashWrite()` returns `true` at the `!commandHasDeepInterviewWriteIntent(command)` branch before the `omx state` allowance even gets a chance to reason about it. Net effect: the PR blocks inline deactivation but leaves the documented file-input deactivation vector open.

Please tighten this by failing closed for planning-phase `omx state write --input-file` / `--input-file=...` in the hook, or otherwise inspect only safe in-workflow state payloads through the same deactivation/gate logic. Add regression coverage for both deep-interview and ralplan guarded Bash so this doesn’t reopen.

Local verification note: I could not run focused tests in the review worktree because `node_modules` was absent (`npm run build` failed before test execution on missing `@iarna/toml`, `@modelcontextprotocol/sdk`, and `zod`). GitHub CI on the PR head is green, so this hold is based on code-path review, not a test failure.

—

iqdoctor:
@Yeachan-Heo You're empirically right that the vector is open — `omx state write` does support `--input-file` (`src/cli/state.ts:5`,:12), the hook can't see the file payload, and `state_write` only gates `mode === 'autopilot'`, so standalone deactivation slips through. I verified all of that.

But before patching, I implemented the suggested fix (a backend gate in `state_write`/`state_clear` rejecting deactivation of an active standalone deep-interview/ralplan, on the resolved payload so it covers `--input`/`--input-file`/substitution/MCP at once) — and it **breaks two existing, sanctioned flows**:

1. **CLI completion** — `codex-native-hook.test.ts` › *"does not block Stop after owner-session ralplan CLI completion writes to native canonical state"* seeds an active ralplan and then does, as the legitimate completion path:
   `executeStateOperation('state_write', { mode:'ralplan', active:false, current_phase:'complete' })`
   The gate rejects it.
2. **Session clear** — `operations.test.ts` › *"does not report a legacy root mode active after clearing the current session scope"* seeds an active deep-interview and expects `state_clear` to succeed (→ `active:false, current_phase:'cleared'`). The gate rejects it.

The reason isn't the gate's *placement* — it's that **completion and escape are state-identical**. Legitimate completion *is* "deactivate the phase" (`active:false` / terminal `current_phase` / clear), through the exact same operation+payload a model would use to escape. `state_write` cannot separate them, because there's nothing in the state to separate.

The only real distinguisher is whether the interview/consensus was *actually performed* — i.e. provenance-protected evidence the model can't fabricate. But a model can write `current_phase:complete` or fake completion evidence itself, so even "allow deactivation only to a terminal phase" is bypassable. That's the **same forgeable-evidence gap** as `skip_authorized_by_user` (read in `deep-interview-gate.ts`, never set by trusted code), which we already flagged as out-of-scope and needing a user-authorization channel.

So any deactivation gate is either **unsound** (forgeable) or **over-broad** (breaks the two flows above). The `--input-file` patch is the over-broad branch one layer up; `commandEndsPlanningPhase` is in the same class.

How would you like to distinguish completion from escape at the state layer? I don't think it's possible without a provenance-protected evidence channel (the model can't fabricate), which is a separate, larger change. Two options I see:
- Treat the standalone planning boundary as an **advisory rail** (enforced against ordinary implementation tool writes, which this PR does robustly), and track the provenance channel as a separate issue; or
- Land the provenance channel (user-authorized interview/consensus completion that gates *all* deactivation transports) as the prerequisite, then no transport patch is needed.

Happy to implement whichever you prefer — but I don't want to merge a deactivation gate that rejects legitimate CLI completion and session clear.

Yeachan-Heo:
Thanks — this confirms the blocker rather than clearing it.

The backend gate should not be a blanket “active standalone ralplan/deep-interview can never be cleared” rule, because the two cases you found are legitimate owner/runtime cleanup paths:

- trusted CLI completion writing `active:false, current_phase:"complete"`
- session/root cleanup via `state_clear`

The PR still needs to close the *model-writable transport* bypass without breaking those sanctioned paths. In practice, the fix should distinguish caller/intent, not just payload shape:

- keep legitimate completion/session cleanup paths allowed when they are invoked by trusted runtime code with the expected completion/cleanup context
- reject model/tool-originated `omx state write --input-file` / `--input` / substitution / MCP payloads that deactivate an active standalone planning phase without that trusted completion context
- cover the bypass at the resolved payload boundary so file-based payloads cannot slip past the hook parser
- add regression tests for both sides: malicious file payload rejected, legitimate CLI completion and session clear still pass

So the current PR remains held until that trust-boundary split is implemented and covered. Do not paper over it with a hook-only heuristic; the bypass is specifically that the hook cannot see the file payload.

—

Yeachan-Heo (changes requested):
REQUEST_CHANGES

Blocker: the new guard only covers Bash `omx state write/clear` text. MCP/structured state tools can still deactivate active standalone planning modes, so transport/path parity is not restored.

Exact evidence on head `30dd6520c7ca76cab05bfefb5741343958604906`:
- `src/scripts/codex-native-hook.ts` `buildDeepInterviewPreToolUseBoundaryOutput` / `buildRalplanPreToolUseBoundaryOutput` only special-case `toolName === "Bash"` plus `PLANNING_MODE_IMPLEMENTATION_TOOL_NAMES` (`Write`, `Edit`, `MultiEdit`, `NotebookEdit`, `apply_patch`). `mcp__omx_state__state_write` and `mcp__omx_state__state_clear` are not classified, so the planning boundary returns `null`.
- Local hook repro: with active session-scoped `deep-interview`, `dispatchCodexNativeHook({hook_event_name:"PreToolUse", tool_name:"mcp__omx_state__state_clear", tool_input:{mode:"deep-interview", ...}})` returned `outputJson: null`.
- Local hook repro: with active session-scoped `ralplan`, `dispatchCodexNativeHook({hook_event_name:"PreToolUse", tool_name:"mcp__omx_state__state_write", tool_input:{mode:"ralplan", active:false, current_phase:"complete", ...}})` returned `outputJson: null`.
- Backend repro: `executeStateOperation("state_clear", {mode:"deep-interview", session_id, workingDirectory})` deleted the active `deep-interview-state.json` (`before:true`, `after:false`).
- Backend repro: `executeStateOperation("state_write", {mode:"ralplan", active:false, current_phase:"complete", session_id, workingDirectory})` returned success and persisted `{active:false,current_phase:"complete"}`.

This violates the PR contract for parity across CLI state write/clear, `--input`/`--input-file`, MCP/state_write paths, and standalone deep-interview/ralplan deactivation/cleanup paths. It is also a coverage gap: the new tests exercise Bash `omx state ...` command strings and `--input-file`, but not MCP/structured state tools.

Positive evidence: GitHub reports head SHA matches, merge state CLEAN, and active CI lanes are green; local `npm run build` passed; targeted local tests passed (`keyword-detector`, `codex-native-hook`, `state/operations`, `modes/base-autopilot-gates`). Those do not cover the blocker above.

—

Yeachan-Heo:
Closing this after the latest maintainer review on head `30dd6520c7ca76cab05bfefb5741343958604906` requested changes. The remaining blocker is transport/path parity: MCP structured state tools can still deactivate active standalone planning modes while the new guard only covers Bash text paths.

A future PR can come back from a clean head once the model-writable transport bypass is closed without breaking trusted CLI completion or session cleanup flows.

—

Issue #1925, opened 2026-04-26
[Bug]


nyx-haile:
# Bug Report: Shared tmux server bloats to 4–11 GiB across long-running multi-project OMX usage, causing OOM

**Repo:** https://github.com/Yeachan-Heo/oh-my-codex
**Version observed:** `oh-my-codex@0.15.0`
**Environment:** Arch Linux, kernel 6.19.12, 16 GiB RAM + 10 GiB swap, tmux 3.6_a, Node 20+
**Severity:** High — reproducibly OOMs the user session.

## Summary

Running multiple `omx` sessions concurrently (different projects, different teams) on the **default tmux socket** causes the **shared tmux server** to grow to multi-gigabyte RSS over hours and eventually triggers Linux OOM, killing user services (`gnome-keyring-d`, `dbus-broker`, browser content processes, etc.). The growth is in tmux's anonymous heap and persists across OMX session destruction. We observed two OOM events, one with a tmux-server peak of ~17 GiB (RSS+swap):

| When | tmux RSS | tmux swap | Total system pressure | What got killed |
|---|---|---|---|---|
| 2026‑04‑24 09:34:33 EDT | 10.75 GiB | 6.07 GiB | RAM full + swap full | tmux server itself |
| 2026‑04‑25 23:42:48 EDT |  4.55 GiB | small     | ~25.7 GiB RAM+swap   | gnome-keyring, dbus-broker, LibreWolf content proc |

A clean restart of the tmux server immediately reclaims the memory, confirming the leak is in the tmux server process.

## What I think is happening (root cause)

OMX uses the **user's default tmux server** (no `tmux -L <socket>` isolation) as a shared bus across all projects, then drives that server very hard from many independent Node processes:

1. **Every project gets its own `notify-fallback-watcher.js` Node daemon.** It polls every 250 ms (adaptive up to 1 s when idle), max-lifetime 6 h, and is auto-respawned by `notify-hook` when it dies. Per-watcher tick rate is ~1.6 Hz sustained; we measured **53,745 `leader_nudge_tick` events in 9.21 h from a single project's watcher.**
2. **Each tick issues 1+ `tmux …` subprocess invocations** (`list-panes`, `display-message`, possibly `capture-pane` and `send-keys`).
3. With 5–6 projects open at once (we typically have `admin`, `forge`, `conch`, `strand`, plus notes worktrees) the **shared tmux server fields ~5–10 short-lived client connections per second for hours**.
[...]

Issue #1899, opened 2026-04-24
[Bug]


nyx-haile:
Detached tmux launch fails when tmux is installed but the tmux server/socket is no longer usable (likely after a restart or tmux-server loss). OMX still chooses the detached-tmux path, then fails on `tmux attach-session -t omx-...` instead of recovering or falling back. I saw this affect more than one OMX instance across repos on April 24, 2026.

A secondary symptom is that after the failure I was unable to scroll back to see history in the current session, which made recovery/debugging harder.

## Environment

- OS: Arch Linux (`Linux 6.19.12-arch1-1 x86_64`)
- Node.js version: `v22.22.2`
- `oh-my-codex` version: `0.14.4`
- Codex CLI version: `0.124.0`
- tmux version: `3.6a`

## Steps to reproduce

1. Launch OMX from a normal shell (not already inside tmux), so it chooses the detached-tmux launch path.
2. Lose the tmux server/session state while leaving OMX state behind (a machine restart seems like one likely trigger).
3. Launch OMX again in the same repo or another repo.
4. OMX attempts to attach to a session named like `omx-<repo>-<branch>-<session>`.
5. The launch fails with a tmux attach error instead of recovering or falling back.
6. After failure, scrollback/history may also be unavailable in the affected session.

Example error:

```text
Error Command Failed /usr/bin/tmux attach-session -t omx-forge-core-1776999496621-oay5dm
error connecting to /tmp/tmux-1000/default (Operation not permitted)
```

Related local checks after failure:

```text
tmux ls
error connecting to /tmp/tmux-1000/default (Operation not permitted)
```

## Expected behavior

If tmux is installed but its server/socket is stale, dead, or otherwise unusable, OMX should detect that before choosing detached tmux. It should either:

- recreate/repair the tmux session automatically, or
- fall back to direct launch cleanly,

and it should avoid leaving stale `.omx/state` / skill-active state behind.

It should also avoid breaking or obscuring scrollback/history access after the failure.

## Additional context

- This does **not** look like a bad session-name format; the `omx-<repo>-<branch>-<session>` naming appears intentional.
- The likely issue is that OMX treats tmux as “available” when `tmux -V` succeeds, but later `tmux attach-session` fails against the actual tmux socket/server state.
- After the failure, repo-local OMX state appeared inconsistent/stale.
- I observed the same class of failure in another OMX instance as well, so this does not seem repo-specific.

Yeachan-Heo:
[...]

Issue #215, opened 2026-02-23
[tmux] Scrollback gets obscured by live OMX output injection


minseol2:
## Summary
When using OMX inside tmux, scrolling up to review previous logs is difficult because output keeps getting injected/refreshed, causing earlier content to be obscured or pushed away.

## Environment
- Tool: oh-my-codex (OMX)
- Terminal: tmux
- Trigger context: sessions where `[OMX_TMUX_INJECT]` output injection is active

## Steps to Reproduce
1. Run several turns in an OMX session.
2. Enter tmux scrollback (copy-mode) and scroll up to inspect earlier output.
3. While scrolling, let new output/injection happen.
4. Observe that the previous logs get hidden, shifted, or effectively become unreadable.

## Expected Behavior
- While the user is in tmux scrollback/copy-mode, auto-injected output should be paused/reduced, or at least should not interfere with reading old logs.
- Previous output should remain stable and readable.

## Actual Behavior
- New output continues to render/inject while scrolling.
- Earlier logs are visually overwritten/shifted, making history inspection hard.

## Impact
- Hard to verify prior steps in long sessions.
- Reduced debugging and review productivity.

## Suggestions
- Detect tmux copy-mode and temporarily pause/minimize HUD/injection updates.
- Buffer output while in copy-mode, then flush after exit.
- Add a user setting such as `pause_in_copy_mode=true`.

Yeachan-Heo:
repo owner's gaebal-gajae (clawdbot) 🦞 — 이슈 확인했습니다. tmux scrollback 방해 버그, 명확한 재현 조건 있음. 작업 시작합니다.

Yeachan-Heo:
Fixed in PR #217 (merged to dev). The injection now checks `#{pane_in_mode}` before sending keys — if the pane is in copy-mode (scrollback), injection is skipped and retried on the next turn. 🦞

PR #3150, opened 2026-07-13
fix(tmux): preserve attached session scrollback

iamgalba:
## Summary

- let OMX-created tmux leader panes inherit the user's configured history limit while a client is attached
- apply the 500-line safety bound and clear stale leader history only when the final client detaches
- restore the inherited pane history limit when a client reattaches
- keep intentionally unattached Hermes bridge sessions bounded immediately

## Why

The detached-history safeguard added for #2597 prevents stale Codex panes from consuming multiple gigabytes of tmux memory, but applying the 500-line limit before the initial attach also truncates active interactive sessions. Users with a larger tmux `history-limit` cannot scroll back to the beginning of a long OMX conversation even while the session is still attached.

This keeps the memory protection at the lifecycle boundary where it is needed while preserving normal interactive scrollback:

1. Attached: inherit the session/global tmux history policy.
2. Final detach: cap the OMX leader pane at 500 lines and clear stale history.
3. Reattach: unset the pane override and inherit the user's tmux policy again.

HUD resize, client-attached reconciliation, mouse, notification, and attach callbacks are unchanged.

## Related issues

- #206 reports that scrolling and copying become unreliable in normal `omx --xhigh --madmax` sessions.
- #215 documents active tmux scrollback being obscured by live OMX output injection and establishes the expectation that users can inspect earlier output while a session is running.
- #1925 documents broader multi-project tmux growth reaching 4–11 GiB and causing OOM failures.
- #2597 is the direct detached-pane memory report behind the current safeguard, including pane histories reaching 8.8–67 GB.
- #1899 separately notes that unavailable scrollback made detached-launch failure recovery and diagnosis harder.

This PR addresses the lifecycle overlap among those reports: retain the user's scrollback policy while attached, then apply the bounded-history safeguard when OMX is actually detached.

## Validation

- `npm run build`
- targeted detached-session lifecycle tests: 8/8 passed
- `node dist/scripts/run-test-files.js dist/cli/__tests__/launch-fallback.test.js`: 24/24 passed
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`
- real tmux lifecycle probe: inherited `50000` -> detached `500` -> reattached `50000`

## Impact

Long-running attached OMX conversations retain the scrollback configured by the user. Fully detached or intentionally unattached OMX leader panes remain bounded against the stale-TUI memory growth reported in #2597.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -5121,14 +5121,20 @@ exit 0
     assert.equal(names.includes("register-resize-hook"), true);
     assert.equal(names.includes("reconcile-hud-resize"), true);
     assert.equal(DETACHED_TMUX_HISTORY_LIMIT, 500);
-    const historyHook = steps.find((step) => step.name === "register-detached-history-prune-hook");
-    assert.ok(historyHook);
-    assert.deepEqual(historyHook.args.slice(0, 3), ["set-hook", "-t", "omx-demo"]);
-    assert.match(historyHook.args[3] || "", /^client-detached\[[0-9]+\]$/);
+    assert.equal(names.includes("bound-unattached-history"), false);
+    const pruneHook = steps.find((step) => step.name === "register-detached-history-prune-hook");
+    assert.ok(pruneHook);
+    assert.deepEqual(pruneHook.args.slice(0, 3), ["set-hook", "-t", "omx-demo"]);
+    assert.match(pruneHook.args[3] || "", /^client-detached\[[0-9]+\]$/);
     assert.equal(
-      historyHook.args[4],
-      `if-shell -F '#{==:#{session_attached},0}' 'run-shell -b "tmux clear-history -t %leader >/dev/null 2>&1 || true"'`,
+      pruneHook.args[4],
+      `if-shell -F '#{==:#{session_attached},0}' 'run-shell -b "tmux set-option -pq -t %leader history-limit 500 >/dev/null 2>&1 || true; tmux clear-history -t %leader >/dev/null 2>&1 || true"'`,
     );
+    const restoreHook = steps.find((step) => step.name === "register-attached-history-restore-hook");
+    assert.ok(restoreHook);
+    assert.deepEqual(restoreHook.args.slice(0, 3), ["set-hook", "-t", "omx-demo"]);
+    assert.match(restoreHook.args[3] || "", /^client-attached\[[0-9]+\]$/);
+    assert.equal(restoreHook.args[4], "set-option -puq -t %leader history-limit");
   });
 
   it("detached history prune hook tolerates a dead leader pane", () => {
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES — exact-head adversarial review of `68be1705053826d937f4b90fbe537c6194536982` against current dev `e8e9467804a73cba0c982b3ed99a0ce8b6843d2c`.

The lifecycle design is not sound.

1. **BLOCKER — the claimed pane-local 500-line bound does not exist on supported tmux.** `src/cli/index.ts:1483-1499`, `1555-1563`, and `4437-4472` use `set-option -p ... history-limit`, but `history-limit` is a session option applied when a pane is created. On the repository's tmux 3.2a floor, a real isolated-server probe showed the existing leader remained `#{history_limit}=2000` after `set-option -pq ... history-limit 5`; after output, its history exceeded 5. The command only changed the session option for future panes. Therefore final detach and fresh Hermes paths clear once but do not keep the running leader bounded, and “reattach restoration” changes session policy rather than the existing pane's effective grid limit. This does not preserve the #2597 memory guarantee.

2. **BLOCKER — detach/reattach is racy.** `src/cli/index.ts:1555-1559` checks `#{session_attached} == 0`, then queues `run-shell -b`. A client can reattach and run the restore hook before that background worker executes; the stale worker then clears scrollback and reapplies the option while the user is attached. The new tests only assert generated command strings/fake-tmux logs and cannot detect this ordering.

3. **BLOCKER — Hermes reuse can mutate an attached interactive session.** `src/cli/index.ts:5397-5406` applies the history mutation before checking whether the reusable session is unattached. `OMX_HERMES_MCP_BRIDGE=1` only means the new caller will not attach; it does not prove another client is absent.

4. **BLOCKER — failed reuse attach leaves no detached recovery bound.** `src/cli/index.ts:5408-5421` unsets the option before `attach-session`; if attach fails while the session remains detached, no detach event follows to restore the safety state.

5. **Coverage gap.** Add real isolated-tmux tests for effective existing-pane history growth, two-client first/final detach, immediate detach→reattach ordering, attached-session Hermes reuse, fresh intentionally-unattached Hermes behavior, failed reuse attach recovery, and session/window/global precedence. Command-shape assertions are insufficient for this API.

Validation on the exact head:
- PASS: build + focused CLI lifecycle suites (`293/293`, `24/24`)
- PASS: `npm run lint`, `npm run check:no-unused`, `git diff --check`
- PASS: `npm pack --dry-run`
- FULL SUITE: 370/371 test files passed; unrelated `dist/mcp/__tests__/server-lifecycle.test.js` failed once and passed immediately in isolation (15/15)
- PASS: independent architecture review lanes both returned BLOCK

Closing because this is draft, behind dev, and the core tmux model is wrong. A replacement/reopened revision must rebase onto current `dev`, use a mechanism that actually bounds an already-created pane while detached, serialize/revalidate lifecycle transitions, and pass the real-tmux matrix above plus the focused/full/static/package gates.

— GJC exact-head adversarial review

PR #3169, opened 2026-07-14
fix(cli): accept BOM-prefixed state input files

Terwox:
## Summary

- accept one leading UTF-8 byte-order mark in JSON read through `omx state --input-file`
- keep inline `--input` parsing unchanged
- retain fail-closed behavior for malformed files, including multiple leading BOMs

## Root cause

Windows tools commonly emit UTF-8 text files with a leading BOM. Node preserves that marker as `U+FEFF` when the file is read as UTF-8, while `JSON.parse` rejects it before the opening JSON token. That made the documented file-based state recovery path fail for otherwise valid payloads.

## Impact

State read, write, and recovery commands can consume JSON payload files produced by Windows PowerShell and other BOM-emitting editors without weakening validation for inline JSON or malformed files.

## Validation

- `npm run build`
- `node --test dist/cli/__tests__/state.test.js` — 15/15 passed
- `npm run lint` — 740 files checked
- `npm run check:no-unused`
- `git diff --check`

### DIFF
--- a/src/cli/__tests__/state.test.ts
+++ b/src/cli/__tests__/state.test.ts
@@ -137,6 +137,46 @@ describe('stateCommand', () => {
     }
   });
 
+  it('reads UTF-8 BOM-prefixed structured input from --input-file', async () => {
+    const dir = await mkdtemp(join(tmpdir(), 'omx-state-input-file-bom-'));
+    const file = join(dir, 'payload.json');
+    await writeFile(file, `\uFEFF${JSON.stringify({ mode: 'ralph', all_sessions: true })}`, 'utf-8');
+    try {
+      let capturedOperation: string | undefined;
+      let capturedInput: Record<string, unknown> | undefined;
+      await stateCommand(['clear', '--input-file', file, '--json'], {
+        stdout: () => undefined,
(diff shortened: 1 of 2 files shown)


Yeachan-Heo:
MERGE_READY

Reviewed exact head `f596d3ec208f9766bc150082873702ad7c0665c6` against dev `e6b0509533039e4467f53699b7ad4e74492e7f59`.

The patch is correctly narrow: it removes exactly one leading U+FEFF only from `--input-file` content before strict `JSON.parse`; inline `--input` remains unchanged. Validation confirmed fail-closed behavior for BOM-after-whitespace, multiple BOMs, malformed/non-object/non-JSON input, and non-UTF-8 bytes, while BOM followed by legal JSON whitespace succeeds. Read/write/clear/get-status/list-active routing remains unchanged and recovery file input works.

Local gates passed: build, focused state tests (15/15), custom BOM/encoding/command matrix, no-unused, lint (740 files), diff check, and `npm pack --dry-run` including package verification. Hosted CI run 29351480505 is terminal success, including CI Status. PR head is unchanged, base dev is unchanged, and GitHub reports mergeable/clean.

—

PR #3167, opened 2026-07-14
fix(hud): deduplicate control-plane authority

Yeachan-Heo:
## Summary
- derive HUD authority from the canonical control-plane state root, with one admitted primary per domain
- add mandatory process-identity leases and ABA-safe lifecycle locks that fail closed on uncertain ownership
- route managed HUD producers through exact owner/session/leader/tmux-instance reconciliation while preserving unmanaged and legacy panes
- make Team HUD startup/shutdown/restore use the same canonical root, lock, alias, and exact-candidate contract
- gate authority-only fallback watchers and stop unsafe live PID-file reaping without birth/owner proof

Closes #3159.

The reported `bwrap` failure remains correlated context only; this change does not claim a causal fix for it.

## Verification
- `npm test` (376 compiled test files after rebasing onto current `dev`)
- `npm run check:no-unused`
- `npm run lint`
- `npm run smoke:packed-install`
- `git diff --check origin/dev...HEAD`
- final independent Architect review: CLEAR / APPROVE
- final Executor QA/red-team: CLEAR / APPROVE

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3173, opened 2026-07-15
fix(hud): verify reconciliation convergence

### LINKED ISSUE #3159: [Bug] Detached OMX sessions accumulate duplicate HUD panes and overwhelm notify authority
## Summary

On Ubuntu 24.04, long-lived detached OMX tmux sessions accumulated duplicate HUD panes. Each HUD parent continued running the notify-fallback authority path, producing a high-frequency `leader_nudge_tick` storm even though the authority state had locking, cooldown, and jitter enabled.

Terminating only the stale OMX tmux sessions stopped the watcher churn immediately and restored reliable sandbox command startup.

## Environment

- oh-my-codex: `v0.18.14`
- Codex CLI currently installed: `0.144.3`
- Node.js: `v22.22.0`
- OS: Ubuntu 24.04.3 LTS
- Kernel: `6.17.0-19-generic`
- Platform: Linux x64
- Surface: detached OMX tmux sessions plus a Codex App/native-hook session outside tmux

## Observed state

There were 20 long-lived HUD processes:

- 14 HUD processes with cwd `<home>`
- 6 HUD processes with cwd `<home>/joah`

Two individual detached sessions had accumulated many HUD panes:

- one session contained the leader pane, one normal HUD pane, and 13 additional `hud --watch --preset=focused` panes
- another session contained the leader pane, one normal HUD pane, and 4 additional focused HUD panes

All duplicate focused panes in a session used the same `OMX_SESSION_ID` and the same `OMX_TMUX_HUD_LEADER_PANE`.

Representative authority state:

```json
{
  "owner": "hud",
  "cooldown_ms": 5000,
  "skip_count": 1367733,
  "last_status": "skipped",
  "last_reason": "rate_limited"
}
```

Despite the 5-second cooldown, the notify log showed approximately 5-6 `leader_nudge_tick` events per second:

```json
{"type":"leader_nudge_tick","leader_only":true,"run_count":1,"stale_threshold_ms":180000,"precomputed_leader_stale":true,"reason":"leader_nudge_checked"}
```

The active one-shot watcher state showed a parent PID belonging to a different stale HUD than the PID recorded in the authority-owner state. This suggests multiple HUD parents were racing through the same cwd-scoped authority files.

## Impact

- `notify-fallback-YYYY-MM-DD.jsonl` repeatedly reached the 10 MB rotation threshold.
- Authority `skip_count` exceeded 1.36 million.
- Old detached Codex processes remained alive across CLI upgrades and auth rotation. One stale process loaded Codex `0.142.0` and emitted `401 token_invalidated` every 180 seconds while the current shared login cache and Codex `0.144.3` were valid.
- During the watcher storm, new sandboxed commands repeatedly failed at startup with:

```text
bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted
```

The bwrap failure is reported as a correlated secondary symptom, not as a proven direct causal mechanism.

## Recovery and verification

I terminated only the six stale `omx-*` tmux sessions, preserving repo files and durable Ultragoal artifacts.

After cleanup:

- active OMX HUD/watcher/old `model_instructions_file` processes: 0
- remaining `omx-*` tmux sessions: 0
- notify-fallback log size remained byte-for-byte unchanged over a 7-second observation
- a normal sandboxed `pwd` succeeded
- six concurrent sandboxed `pwd` commands all succeeded
- `codex login status` still reported a valid ChatGPT login

## Reproduction notes

The exact pane-duplication trigger is not yet isolated. The affected sessions had survived for weeks and across OMX/Codex upgrades, reconnects, and repeated HUD use.

A likely reproduction direction is:

1. Start a detached OMX tmux session.
2. Leave it alive across reconnects or repeated HUD initialization.
3. Invoke or reinitialize focused HUD display multiple times.
4. Run:
   ```sh
   tmux list-panes -a -F '#{session_name} #{pane_index} #{pane_pid} #{pane_start_command}'
   ```
5. Check whether additional `omx.js hud --watch --preset=focused` panes are appended instead of reusing or replacing the existing HUD pane.
6. Inspect `.omx/state/notify-fallback-authority-*.json` and the notify-fallback log cadence.

## Expected behavior

- At most one live HUD authority owner should exist per intended scope.
- Reinitializing HUD should reuse or replace its existing pane, not append another focused HUD pane.
- Detached sessions should not retain stale Codex/HUD/MCP process trees indefinitely after their owning lifecycle is abandoned.
- The authority lock and cooldown should remain effective even if duplicate HUD parents somehow exist.
- A stale detached Codex process should not keep polling with an invalidated in-memory token forever.

## Suspected areas

- detached-session HUD pane lifecycle and cleanup
- focused HUD initialization idempotency
- cwd-scoped notify authority ownership when multiple HUD parents exist
- stale detached session reaping after reconnect, upgrade, or auth rotation

## Related

- #2659
- #2465
- #981

Inference1:
## Summary

- fail closed when tmux scans, stale/orphan cleanup, lifecycle-lock setup, pane deletion, or resize-hook registration cannot be verified
- require post-create and final scans to prove exactly one owner-matching HUD pane instead of trusting an unobserved split pane id
- stop before creating a replacement when an existing malformed HUD pane cannot be killed
- add regression coverage for scan failures, duplicate races, pane-id reuse, hook failures, and reconciliation convergence

## Context

This is a focused supplement to #3167. It targets that PR's feature branch so the owner-authority fix and these convergence guarantees can land together without opening a competing implementation against `dev`.

## Validation

- `npm run build`
- HUD tests: 372 passed, 0 failed
- native-hook integration tests: 536 passed, 0 failed
- `npm run lint`
- `npm run check:no-unused`
- `npx tsc --noEmit`
- `git diff --check`

The full 377-file local run also exercised the broader suite. On macOS, six existing platform-sensitive files remain red because of GNU `stat -c` and `/var` versus `/private/var` path assumptions; none are touched by this diff. The focused suites above are green, and Linux CI remains the authoritative full-suite gate.

## Related

Supplements #3167.

Closes #3159 when merged through #3167.

### DIFF
--- a/src/hud/__tests__/reconcile.test.ts
+++ b/src/hud/__tests__/reconcile.test.ts
@@ -45,19 +45,67 @@ function testReconcileDomain(cwd: string, deps: ReconcileHudForPromptSubmitDeps)
   };
 }
 
-async function reconcileHudForPromptSubmit(cwd: string, deps: ReconcileHudForPromptSubmitDeps = {}) {
+type ReconcileTestDeps = ReconcileHudForPromptSubmitDeps & {
+  exposeCreatedPaneAfterCreate?: boolean;
+  exposeKilledPaneAfterKill?: boolean;
+};
+
+async function reconcileHudForPromptSubmit(cwd: string, deps: ReconcileTestDeps = {}) {
   const sessionId = deps.sessionId?.trim() || deps.env?.OMX_SESSION_ID?.trim() || '';
+  const leaderPaneId = deps.env?.TMUX_PANE?.trim() || '';
   const listPanes = deps.listCurrentWindowPanes;
+  const createPane = deps.createHudWatchPane;
+  const killPane = deps.killTmuxPane;
+  let createdPaneId: string | null = null;
+  const successfullyKilledPaneIds = new Set<string>();
+  const {
+    exposeCreatedPaneAfterCreate: _exposeCreatedPaneAfterCreate,
+    exposeKilledPaneAfterKill: _exposeKilledPaneAfterKill,
+    ...rawDeps
+  } = deps;
   return reconcileHudForPromptSubmitRaw(cwd, {
-    ...deps,
+    ...rawDeps,
     listCurrentWindowPanes: listPanes
-      ? (paneId) => listPanes(paneId).map((pane) => ({
-        ...pane,
-        ...(pane.startCommand.includes('OMX_SESSION_ID=') && !pane.paneInstanceId
-          ? { paneInstanceId: sessionId, sessionInstanceId: sessionId }
-          : {}),
-      }))
+      ? (paneId) => {
+        const panes = listPanes(paneId)
+          .filter((pane) => deps.exposeKilledPaneAfterKill === true || !successfullyKilledPaneIds.has(pane.paneId))
+          .map((pane) => ({
+            ...pane,
+            ...(pane.startCommand.includes('OMX_SESSION_ID=') && !pane.paneInstanceId
+              ? { paneInstanceId: sessionId, sessionInstanceId: sessionId }
+              : {}),
+          }));
+        if (
+          createdPaneId
+          && deps.exposeCreatedPaneAfterCreate !== false
+          && !panes.some((pane) => pane.paneId === createdPaneId)
+        ) {
+          panes.push({
+            paneId: createdPaneId,
+            currentCommand: 'node',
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks for the focused supplement. I’m closing this revision rather than carrying a parallel feature-branch patch. The current diff still changes required reconciliation semantics incorrectly: `registerHudResizeHook()` intentionally returns `false` when the required resize hook is installed but the optional `window-layout-changed` hook is unavailable on older tmux builds; treating that degraded-compatible state as total reconciliation failure breaks supported environments. The canonical #3167 repair lane is already active and will incorporate independently verified convergence coverage without inheriting this contract regression.

—

PR #3196, opened 2026-07-17
fix(ralplan): fail closed without documented leader proof

### LINKED ISSUE #3194: [Bug] #3181 hotfix still rejects fresh Codex 0.144.5 leaders with parent_not_active_leader
## Summary

PR #3187 closed #3181 by adding a `PreToolUse` leader-attestation bootstrap. On the merged dev commit, however, a fresh Codex 0.144.5 leader still cannot record the first adapted Ralplan role intent. The failure has changed from `missing_session` to:

```json
{"ok":false,"reason":"parent_not_active_leader"}
```

No Architect child is spawned. The same result was reproduced in a fresh non-interactive run, a fresh interactive run, and a resumed second turn.

## Environment

- OS: macOS, arm64
- Node.js: v26.5.0
- `oh-my-codex`: `v0.20.3-dev-3db1234755be` (merge commit `3db1234755bea28d1cb2e32d039e07a9ae1511c9`)
- Codex CLI: `0.144.5`
- Surface: outside tmux; exposed collaboration spawn has `task_name`, `message`, and `fork_turns`, without `agent_type`

## Steps to reproduce

1. Install the current dev build containing PR #3187.
2. Start a fresh outside-tmux Codex run in a disposable repository.
3. Before spawning any child, run:

   ```bash
   omx ralplan role-intent write \
     --role architect \
     --parent-thread "$CODEX_THREAD_ID" \
     --json
   ```

4. Observe exit code 1 and:

   ```json
   {"ok":false,"reason":"parent_not_active_leader"}
   ```

5. Inspect the sanitized tracker summary. The current session has no durable `leader_attested_at`, so no pending Architect role intent is created.

## Expected behavior

The first in-turn `PreToolUse` on a positively identified root leader should establish the durable leader attestation needed to record the Architect role intent. The existing child-exclusion and distinct Architect -> Critic ordering guarantees must remain fail-closed.

## Actual behavior

The root turn is not attested. The role-intent command therefore rejects the supplied parent as not being the active tracker leader, and adapted Ralplan cannot begin on the App-compatible surface.

## Evidence and likely contract mismatch

At merge commit `3db1234`:

- `hasUnambiguousNativeLeaderThreadBinding()` requires `payload.thread_id` or `payload.threadId` to equal `nativeSessionId`.
- The `#3181 Phase-1 (PreToolUse)` bootstrap calls `attestLeaderThread()` only when that predicate passes.
- The current Codex hooks documentation lists `session_id`, `transcript_path`, `cwd`, `hook_event_name`, and `model` as common fields. `PreToolUse` additionally lists `turn_id`, `tool_name`, `tool_use_id`, and `tool_input`; it does not document `thread_id` as a supported field.
- Codex documents that subagent hooks use the parent `session_id`, so `session_id` alone is not a safe current-thread identity.

The observed state is consistent with that contract mismatch: `PreToolUse` runs, but no leader attestation is published, and `recordNativeLeaderIntent()` returns `parent_not_active_leader`.

Official hook reference: https://learn.chatgpt.com/docs/hooks

## Suggested direction

1. Add a sanitized live Codex 0.144.5 hook-shape fixture and an end-to-end regression that starts without pre-created tracker state.
2. Base leader/child classification on a supported stable runtime identity and positively proven lifecycle ancestry, rather than requiring an undocumented `thread_id` carrier.
3. Preserve fail-closed behavior when root identity is unavailable or ambiguous.
4. Do not authorize from `session.json` alone and do not relax the all-session child/subagent exclusion.
5. If current supported hook fields cannot prove root identity safely, fail Ralplan preflight with an actionable unsupported-surface diagnostic rather than failing only at the first role-intent write.

## Regression coverage requested

- Fresh outside-tmux `omx exec`: first Architect role intent succeeds or preflight fails explicitly.
- Fresh interactive run: same behavior.
- Resumed second turn: no dependency on a prior completion-notify side effect.
- Native child / malformed-child payloads cannot self-attest as leader.
- Architect and Critic remain distinct, role-bound, and sequential.

## Related issues

- #3181 — original fresh-turn bootstrap issue, closed by PR #3187.
- openai/codex#32782 — separate native `spawn_agent.agent_type` availability issue. Fixing it would restore the preferred native-typed lane but would not repair this adapted fallback for surfaces that still lack `agent_type`.

## Logs / screenshots

No raw hook payload, transcript, filesystem path, session/thread UUID, username, email address, hostname, IP address, or authentication material is attached. The minimal sanitized command result above is sufficient to reproduce the failure.

Yeachan-Heo:
## Summary
- preserve typed native `agent_type` routing while making the Codex 0.144.5 adapted Ralplan path explicitly unsupported
- deny canonical adapted role-intent commands before pointer, tracker, recovery, or runtime work
- remove legacy adapted bind/recovery authority and reject `omx_adapted` consensus evidence
- neutralize routing-only Ralplan state on failed explicit preflight so it cannot drive HUD/runtime or Stop enforcement
- add sanitized packed coverage, exact CLI contracts, ADR, skills, generated guidance, and security regressions

## Verification
- build, typecheck, no-unused, lint, plugin mirror, and diff checks
- focused #3194, consensus, runtime, leader-contract, packed, recent-bug, and full affected hook suites
- direct CLI probe: preflight exits 1, emits exact JSON, and changes active Ralplan state to `active:false,current_phase:cancelled`
- Node 20 path-sensitive isolation: identical files pass through `/tmp/node20-runtime/node` but fail through copied `/tmp/node20` because ownership logic observes `process.execPath` basename `node20`; CI uses a `node` basename

Closes #3194


Yeachan-Heo:
MERGE_READY

Exact-head commit: `a894054dbfea07a4e7fcfedc3a4f74408ed4ac65`

- Architect hostile review: MERGE_READY, no findings.
- Executor QA/red-team: MERGE_READY, informational findings only.
- Hosted CI: all required lanes green, including Node 20 `cli-core-rest`, `hooks-notify-platform`, and `team-state-runtime`.
- Adapted bind/recovery authority is removed; `omx_adapted` consensus is rejected; typed native routing remains supported.
- Explicit unsupported preflight cancels routing-only Ralplan state before returning exact failure JSON and exit 1.

—

PR #3201, opened 2026-07-17
fix(config): reconcile duplicate project trust tables

### LINKED ISSUE #3199: [Bug] Project trust sync duplicates [projects."<cwd>"] when an un-fenced trust table exists, leaving config.toml unparseable
## Summary

`syncProjectScopeTrustStateFromRuntime` (`src/config/generator.ts`) strips its own marker-fenced block before re-appending it, which makes the sync idempotent — but only while the fence is intact. A `[projects."<cwd>"]` table that lives **outside** the fence is invisible to `stripMarkerBlock`, so the sync appends a *second*, fenced copy of the same table. TOML forbids declaring the same table twice, so the durable project `config.toml` becomes unparseable and codex refuses to start on every subsequent launch:

```text
Error loading configuration: Error parsing project config file /path/.codex/config.toml: TOML parse error at line 39, column 11
   |
39 | [projects."/path"]
   |           ^^^^^^^^
duplicate key
```

Because omx writes the config before codex reads it, this is not recoverable from the CLI — the user has to hand-edit `config.toml` to get codex back.

This is a follow-up to the trust-sync introduced for #2470, not a regression of it: the sync solves the original problem, it just isn't resilient to a project table it did not write.

- **Version:** `oh-my-codex` 0.20.2 (`codex-cli` 0.144.5, Node 24.14.0, linux arm64)
- **Affected:** project-scope launches (`.omx/setup-scope.json` → `scope=project`)

## Reproduction

Standalone, no omx setup required:

```js
// repro.mjs
import { syncProjectScopeTrustStateFromRuntime } from 'oh-my-codex/dist/config/generator.js';

const CWD = '/home/me/proj';
const runtimeConfig = `[projects."${CWD}"]\ntrust_level = "trusted"\n`;
const hooksPath = `${CWD}/.codex/hooks.json`;

// A durable project config.toml whose trust table is NOT inside the marker fence.
const projectConfig = `[features]
goals = true

[projects."${CWD}"]
trust_level = "trusted"
`;

const keys = (s) => (s.match(/^\[projects\./gm) || []).length;

let cur = projectConfig;
console.log('before:', keys(cur), 'projects table(s)');
cur = syncProjectScopeTrustStateFromRuntime(cur, runtimeConfig, hooksPath);
console.log('after :', keys(cur), 'projects table(s)');
console.log(cur);
```

Output:

```text
before: 1 projects table(s)
after : 2 projects table(s)

[features]
goals = true

[projects."/home/me/proj"]
trust_level = "trusted"

# OMX-synced Codex project trust state (from runtime CODEX_HOME)
[projects."/home/me/proj"]
trust_level = "trusted"

# End OMX-synced Codex project trust state
```

The result is not valid TOML.

## Expected vs actual

**Expected:** the sync converges on exactly one `[projects."<cwd>"]` table and the config stays parseable, regardless of whether a trust table already existed outside the fence.

**Actual:** the count goes `1 → 2` and stays at 2 on subsequent syncs — the un-fenced table is never stripped, while the fenced one is stripped and re-added each time. The config is unparseable from the first sync onward.

For contrast, a config whose trust table *is* inside the fence is correctly idempotent: running the real sync three times against ours left it at one table.

## Impact

Any project whose `config.toml` ever acquires an un-fenced `[projects]` table is one project-scope launch away from a codex that will not start, with an error that points at the config rather than at omx. It took us a while to connect the two.

I can't say with certainty how the un-fenced table first appeared in our repo — git history shows it correctly fenced at HEAD, and the un-fenced copy showed up as an uncommitted working-tree change. A hand-edit or an older layout seem the likely candidates, and `trust_level` is only ever written through the fenced sync in the shipped `dist`, so omx writing it directly doesn't look plausible. Either way the provenance seems less interesting than the fact that the state is reachable and unrecoverable-in-place, while `stripMarkerBlock` already knows exactly what it is looking for.

## Suggested fix

In `syncProjectScopeTrustStateFromRuntime`, before appending the fenced block, drop (or absorb into the fence) any `[projects.*]` table the sync is about to re-declare. That preserves the function's existing "the fence is the single source of truth" contract and makes it converge from the broken state instead of cementing it.

If silently rewriting user-authored tables is undesirable, failing loudly would still be a large improvement over emitting invalid TOML — the current failure surfaces as an opaque codex parse error at a line number the user never wrote.

Related: `repairProjectScopeTrustStateForLaunch` already exists to heal the 0.18.3 relaunch regression, but it early-returns unless a fenced block is found (`extractMarkerBlockContent(...)` → `if (!syncedTrustBlock) return projectConfig;`), so it does not cover the un-fenced case. I confirmed this by calling it against the repro config above: it returns the input unchanged.

## Workaround (for anyone hitting this)

Delete the `[projects."<cwd>"]` table that sits outside the `# OMX-synced Codex project trust state ...` / `# End OMX-synced Codex project trust state` markers, keeping the fenced one. Dropping it is safe even if no fenced block exists — omx re-syncs trust state from the runtime `CODEX_HOME` on the next launch, so at worst Codex asks to trust the workspace once more.

Yeachan-Heo:
## Summary
- reconcile equivalent external and managed project trust tables without deleting user-owned TOML
- preserve comments, payload gaps, BOM/EOL/trailing bytes, and unrelated project tables
- fail closed for malformed, ambiguous, unsupported, implicit nested, or comment-sensitive forms
- cover repeated sync, launch repair, hook-only state, and package parity

## Verification
- focused compiled config/CLI suites: 394/394 passed
- lint, typecheck, no-unused: passed
- packed tarball consumer replay: passed
- hostile exact-head Architect: CLEAR / APPROVE
- exact-head QA/red-team: passed

Fixes #3199

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -2756,6 +2756,10 @@ describe("project launch scope helpers", () => {
           'trusted_hash = "sha256:setup-owned"',
           "# End OMX-owned Codex hook trust state",
           "",
+          "# User-owned project trust source must remain external during launch repair.",
+          `[projects."${wd}"] # retained external ownership`,
+          'trust_level = "trusted"',
+          "",
           "# OMX-synced Codex project trust state (from runtime CODEX_HOME)",
           `[projects."${wd}"]`,
           'trust_level = "trusted"',
@@ -2793,6 +2797,13 @@ describe("project launch scope helpers", () => {
       );
       assert.ok(runtimeConfig.includes(`[projects."${wd}"]`));
       assert.ok(repairedProjectConfig.includes(`[projects."${wd}"]`));
+      assert.match(repairedProjectConfig, /User-owned project trust source must remain external/);
+      assert.match(repairedProjectConfig, new RegExp(`^${escapeRegExp(`[projects."${wd}"]`)} # retained external ownership$`, "m"));
+      assert.equal(
+        countMatches(repairedProjectConfig, new RegExp(`^${escapeRegExp(`[projects."${wd}"]`)}(?:\\s|$)`, "gm")),
+        1,
+        "launch repair must remove only the marker-owned duplicate before the runtime mirror is written",
+      );
     } finally {
       await rm(wd, { recursive: true, force: true });
     }
--- a/src/config/__tests__/generator-idempotent.test.ts
+++ b/src/config/__tests__/generator-idempotent.test.ts
@@ -15,9 +15,11 @@ import {
   hasExactOmxSeededBehavioralDefaultsPair,
   mergeConfig,
   repairConfigIfNeeded,
+  repairProjectScopeTrustStateForLaunch,
   stripExistingOmxBlocks,
   stripManagedCodexHookTrustState,
   stripOmxFeatureFlags,
+  syncProjectScopeTrustStateFromRuntime,
   upsertManagedCodexHookTrustState,
   stripOmxSeededBehavioralDefaults,
 } from "../generator.js";
@@ -32,6 +34,10 @@ import { OMX_FIRST_PARTY_MCP_SERVER_NAMES } from "../omx-first-party-mcp.js";
 /** Count occurrences of a pattern in text */
 function count(text: string, pattern: RegExp): number {
   return (text.match(pattern) ?? []).length;
+}
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 }
 
 async function writeSetupGeneratedHookTrustFixture(wd: string): Promise<{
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
MERGE_READY

Exact delivery identity:
- head: `b826d1054afa08166efc4bd78f0feae4e96f821e`
- base: `e4c0be64085ee340078a920ac600f1147c616d12`
- hosted CI run: https://github.com/Yeachan-Heo/oh-my-codex/actions/runs/29579826514 (green)
- hostile Architect: `CLEAR / APPROVE` (`23-Issue3199RebasedArchitect`)
- exact-head QA/red-team: passed (`24-Issue3199RebasedQA`; `/tmp/issue-3199-rebased-head-qa.json`)
- cleanup sweep: `CLEAN` (`11-Issue3199Cleaner`)
- Gajae receipt: `.gjc/_session-019f6ebd-b9c4-7000-9ca5-a465ed8554c3/ultragoal/ledger.jsonl`, event `c2e4d997-19e9-4bb7-bcae-1a423c356264`

The receipt remains pending merge SHA, post-merge `dev` build, terminal comment URL, and completion timestamp; those fields will be finalized against the same stable ledger identity after merge.

—

PR #3205, opened 2026-07-18
Allow rejected Codex roots to exit safely

WangErgouaaaa:
## Summary

- allow an unmatched root Stop to exit only when that payload session has no scoped OMX state
- preserve fail-closed behavior for unusable pointers and scoped Team/Ralph state without mutating the authoritative owner
- harden the Team status fixture against collisions with real host tmux pane IDs and document the approved repair

## Root cause

A rejected secondary root can remain conversationally usable after SessionStart loses ownership, but every later Stop is blocked as `session_scope_unmatched`. Restored VS Code terminals can amplify this by launching several Codex roots in one checkout.

## Test Plan

- [x] `npm test` — all 383 test files completed; catalog check passed
- [x] native hook suite — 555/555
- [x] Team CLI suite — 76/76
- [x] installed-package black-box — stateless unmatched Stop returns null; scoped unmatched Stop remains blocked
- [x] authoritative live pointer remained byte-identical across install and acceptance

### DIFF
--- a/docs/superpowers/plans/2026-07-18-unmatched-stop-safe-exit.md
+++ b/docs/superpowers/plans/2026-07-18-unmatched-stop-safe-exit.md
@@ -1,0 +1,498 @@
+# Unmatched Codex Stop Safe Exit Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks for the contribution. This repository does not accept pull requests targeting `main`; active development targets `dev`.

This PR also mixes the portable OMX change with host-specific launcher design and local filesystem instructions, so it cannot be retargeted as-is. Please open a new, minimal PR against `dev` containing only the repository-owned source and regression tests. Exclude host-local shell configuration, installation paths, and machine-specific operational plans.

Closing this PR rather than rewriting contributor history.

—

WangErgouaaaa:
Replaced by #3206. The new PR targets `dev` and contains only the portable native-hook source change plus its regression test; host-local configuration, design documents, operational plans, and unrelated test changes are excluded.

Issue #3138, opened 2026-07-12
[Bug] Ralplan terminal state splits between native session and OMX_SESSION_ID when owner alias is absent


arikon:
## Summary

Ralplan can create two state scopes for one user-facing run:

- native Codex session ID from `.omx/state/session.json`;
- a distinct runtime owner ID from `OMX_SESSION_ID`.

Completing the OMX-owner state does not complete the native state read by the Stop hook, which then loops with `continue_from_artifact`. This reproduces the failure class fixed by #2959/#2960, but here `session.json` lacks `owner_omx_session_id`, so the owner-alias canonicalization has no metadata to use.

## Environment

- OS: macOS arm64
- Node.js version: `v24.18.0`
- `oh-my-codex` version: `v0.20.2`
- Codex CLI version: `0.144.0-alpha.4` (Codex App bundled CLI)
- Surface: Codex App/native hooks attached to OMX tmux runtime

## Steps to reproduce

1. Start a Codex native-hook session attached to an OMX runtime where:

   ```text
   OMX_SESSION_ID=omx-1783868526603-kjypjl
   ```

   but `.omx/state/session.json` contains only the native ID and no owner alias:

   ```json
   {
     "session_id": "019f565a-1a79-7023-8ccf-cf06088fc060",
     "native_session_id": "019f565a-1a79-7023-8ccf-cf06088fc060"
   }
   ```

2. Invoke `$ralplan`. Observe that native activation creates:

   ```text
   .omx/state/sessions/019f565a-1a79-7023-8ccf-cf06088fc060/ralplan-state.json
   ```

3. Continue the workflow through sequential native Architect and Critic approval. OMX runtime activity creates/uses a second scope:

   ```text
   .omx/state/sessions/omx-1783868526603-kjypjl/ralplan-state.json
   ```

4. Complete Ralplan with the supported terminal command and valid tracker-backed consensus evidence:

   ```sh
[...]

PR #3206, opened 2026-07-18
fix(hooks): allow stateless unmatched root Stops to exit

WangErgouaaaa:
## Summary

- return no Stop output when an unmatched root session has no session-scoped OMX state
- keep unusable pointers and unmatched sessions with scoped Team or Ralph state fail-closed
- verify the authoritative pointer remains byte-identical and no native Stop state is created

## Test Plan

- [x] `npm run build`
- [x] issue #3138 focused regression — 1/1
- [x] scoped Team/Ralph mismatch regressions — 2/2
- [x] native-hook suite — 557/558; the sole unrelated Conductor PATH parsing failure reproduces unchanged on a clean `origin/dev` snapshot at `0e67d48d`

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -4334,13 +4334,19 @@ PY`,
       assert.equal(foreignStop.outputJson?.stopReason, "session_pointer_unusable");
       assert.equal(existsSync(join(foreignCwd, ".omx", "state", "native-stop-state.json")), false);
 
+      const conflictingPointerPath = join(conflictingCwd, ".omx", "state", "session.json");
+      const conflictingPointerBeforeStop = await readFile(conflictingPointerPath, "utf-8");
       const unmatchedStop = await dispatchCodexNativeHook({
         hook_event_name: "Stop",
         cwd: conflictingCwd,
         session_id: "native-unmatched-stop-3138",
       }, { cwd: conflictingCwd });
-      assert.equal(unmatchedStop.outputJson?.decision, "block");
-      assert.equal(unmatchedStop.outputJson?.stopReason, "session_scope_unmatched");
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES

The proposed unmatched-Stop exception is not fail-closed under hostile session input or concurrent state creation.

Concrete blockers at exact head `ab4718bac810d06c140dfc4af2c3700934779f07`:

1. `stopPayloadSessionId` is used directly in `join(stateDir, "sessions", stopPayloadSessionId)` without a validated single-segment session-ID contract. Traversal/absolute/hostile identifiers can redirect the authority probe away from the intended scoped session path. A negative result can therefore authorize `outputJson = null` without proving absence of state for a valid scoped identity.
2. The absence check is a non-atomic `existsSync` snapshot. The function subsequently awaits native-subagent classification before dispatching the Stop result. Session-scoped Team/Ralph/native state can be created after the check and before the null return, so the code can allow Stop despite state existing at decision time.
3. The changed test covers only one ordinary identifier and a static absent directory. It does not cover traversal/absolute/empty/Unicode-hostile IDs, symlinked session paths, alternate/nested selected-root fixtures, or deterministic creation between the existence check and final dispatch.
4. Hosted status for this exact head contains only PR metadata checks (`PR Size Label`, `Draft Check`, and skipped `PR Base Guidance`). The full product CI lanes and protected aggregate are absent, so metadata green cannot authorize merge.

The bounded hostile review plan also failed its final independent Architect/Critic gate on these same executable proof boundaries: exact zero-mutation observation, hostile identity oracle, and check-to-dispatch safety remain unproven. Please validate/canonicalize the session identifier before filesystem authority checks and make the absence/dispatch decision race-safe, with focused adversarial regression coverage.

—

PR #3237, opened 2026-07-20
fix(ultragoal): bound paused Stop guidance

Yeachan-Heo:
## Summary
- recognize paused Ultragoal review and user-decision states
- emit one pause-specific Stop message without invalid completion reconciliation
- allow the duplicate Stop replay to finish

## Verification
- `npm run build`
- `npx biome lint src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts`
- focused paused Stop regression
- focused Ultragoal Stop regressions

PR #3235, opened 2026-07-20
Prevent live session pointers from native-start replacement

WangErgouaaaa:
## Summary
- Reject native `SessionStart` reconciliation when a live pointer already belongs to a different native session.
- Preserve same-native metadata refresh and stale/dead pointer recovery behavior.
- Update session and native-hook regressions, including PATH isolation for conductor direct-write tests.

## Why
A nested or ephemeral native Codex session could publish over a live parent pointer. Stop authorization then sees stale or unrelated pointer evidence and fails closed.

## Test Plan
- [x] `npm run build`
- [x] `npm run check:no-unused`
- [x] `node dist/scripts/run-test-files.js dist/hooks/__tests__/session.test.js`
- [x] `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js`
- [x] `git diff --check`

## Not Tested
- Installed OMX live smoke after packaging.

PR #3145, opened 2026-07-13
fix(team): make PowerShell psmux pane creation safe

Yeachan-Heo:
Closes #3123

## Summary
- emit native PowerShell `$env:` assignments plus `&` for Team and standalone HUD startup, preserving paths and values containing spaces or embedded single quotes
- keep generated HUD ownership metadata readable through the canonical HUD reader while rejecting malformed, mixed, duplicate, quoted-string, and comment lookalikes fail-closed
- snapshot native-Windows target and server-global pane IDs before creation, accept exactly one canonical split result, and require consecutive exact global `<pane_id> 0` observations before use
- restrict rollback to provenance-approved operation-created panes; leader, cross-window/pre-existing, duplicate, ambiguous, and malformed IDs never gain destructive authority
- preserve existing POSIX/MSYS command shapes, encoded native worker startup, direct native HUD resize, mismatched-owner/unreadable-owner shutdown behavior, and intentional pre-creation HUD cleanup

`src/team/__tests__/runtime.test.ts` is the one documented exception to the original four-file plan: its native-Windows shutdown mock needed stateful global snapshot/liveness responses so the broader regression gate exercised the new contract instead of failing from stale fixture topology.

## Verification
Contract tests: Linux fixtures.

- `npm run build` — pass
- `node --test dist/hud/__tests__/tmux.test.js dist/team/__tests__/tmux-session.test.js` — 280/280 pass
- focused native shutdown regression — 1/1 pass
- `node dist/scripts/run-test-files.js dist/hud/__tests__ dist/team/__tests__` — 49 test files pass
- `npm run lint` — pass
- `npm run check:no-unused` — pass
- `git diff --check` — pass
- current `origin/dev` baseline comparison for the six previously failing startup/shutdown cases — 6/6 pass on baseline and 6/6 pass on the branch after the issue-scoped fixes

Independent exact-head gates for `87279997415ed7c9c89f76c38284408c932c3c74`:
- Architect: CLEAR / APPROVE, no blockers
- Executor QA/red-team: APPROVE, no blockers

## Scope boundaries
- #3121 retains broader send/notify/PID/kill/teardown exact-live ownership
- #3122 retains pane-owner option fallback
- #3124 retains Git Bash data-plane / PowerShell control-plane bridging
- no release, publish, or merge action is included

Real Windows/psmux runtime: not run. Linux mocked `win32` fixtures prove command/parser/lifecycle contracts only.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3238, opened 2026-07-20
fix: bound native Stop hook pointer loops

masterFoad:
## Summary
- No-op Stop when the selected session pointer belongs to a foreign cwd.
- Bound other unusable session-pointer Stop failures to one diagnostic block per replay chain.
- Add focused regressions and update native hook docs.

## Testing
- `npm run build`
- `node --test --test-concurrency=1 --test-name-pattern='owner-env terminal|bounds Stop replays|bounds Stop replays when session.json is malformed|no-ops Stop when session.json points to another worktree|session-scoped team id' dist/scripts/__tests__/codex-native-hook.test.js`
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`
- Direct CLI smoke for a foreign-cwd Stop payload returned `{}`


Yeachan-Heo (approved):
## Red-team review — PR #3238 (`fix: bound native Stop hook pointer loops`)

**Verdict: MERGE_READY**

Reviewed the full diff and the surrounding native Stop / session-pointer authorization state machine adversarially against head `8a3b2b1763683cef076abb3ce47a0b2ecc5bc114`. The change is scoped to the *unauthorized* Stop failure branch in `dispatchCodexNativeHook` (`src/scripts/codex-native-hook.ts:20457-20477`) and is a genuine loop-bounding fix, not an authorization weakening.

### Findings

**1. foreign-cwd no-op on first Stop is correct, not a bypass.** The failure branch is reached only when `allowImplicitSessionSideEffects === false`. A `foreign-cwd` pointer (`isSessionStateAuthoritativeForCwd` → `sameFilePath` mismatch) has authority over *another* workspace and cannot authorize this cwd's Stop at all; blocking on it was a false block that looped forever (the pointer status never changes across replays). A legitimately active session in *this* cwd is authoritative-for-cwd → `usable`, so it routes to `buildStopHookOutput` (persistence enforced there), never the failure branch. No active-workflow persistence is reachable to bypass.

**2. No attacker/stale-pointer persistence bypass.** Ralph/team/autopilot/Ultragoal continuation Stop-blocks all live on the `usable`/`absent`/`authorized` path. Forging `stop_hook_active` only lets the forger's *own* unauthorized Stop terminate; it cannot reach or mutate an authorized session's persistence. The unmatched (`session_scope_unmatched`) case still leaves the authoritative pointer untouched.

**3. `stop_hook_active` replay suppression is safely scoped.** Each hook invocation reads its own payload; the flag is per-Codex-stop-attempt and resets on genuine progress. The failure branch emits only one kind of block, so the simpler raw check (vs. the signature-scoped `maybeReturnRepeatableStopOutput` on the usable path) cannot suppress an unrelated later Stop — a fresh Stop carries `stop_hook_active=false` and re-emits the diagnostic. No cross-session state is mutated.

**4. One truthful diagnostic preserved.** malformed / identity-indeterminate / `session_scope_unmatched` each block once on the first Stop (`stopReason` asserted in tests) then no-op on the Codex-marked replay. `foreign-cwd` intentionally emits no block (it is another workspace's pointer) — consistent with its exclusion from the diagnostic set.

**5. Native-stop-state durability/non-mutation is correct.** The failure branch never persists a signature; regression tests assert `native-stop-state.json` is not created and a pre-existing file is byte-identical before/after.

**6. Safe interaction with adjacent work.** #3237 (paused Stop guidance) and #3235 (live pointer replacement rejection) both operate on the usable path — #3235 is the complementary mitigation preventing a live pointer from being flipped to `foreign-cwd`. Exact pane/cwd authority (#3145) and Windows/path normalization hold: `sameFilePath` canonicalizes via `realpathSync.native`, and the symlink-alias authority test confirms the real cwd is never misclassified as foreign.

### Verification
- `codex-native-hook.test.js`: 573/573 pass (includes the new foreign-cwd no-op, malformed/identity-indeterminate replay-bound, and unmatched-replay cases).
- `session.test.js` + `prompt-session-provenance.test.js`: 69/69 pass (pointer classification, symlink authority, live-replacement rejection).
- Clean `tsc` build; docs contract row updated accurately.
- Hosted CI green for exact head `8a3b2b1`: required check `CI Status` pass; Typecheck / Lint / Docs / Ralph Persistence Gate / Coverage Gate / all Node test lanes pass.

Reviewed serially; no subagents/teams spawned (host cgroup pressure honored). Contract change treated as adversarial and independently verified against source and tests rather than reporter claims.

—

PR #3258, opened 2026-07-22
fix(tmux): preserve separator argv boundaries

### LINKED ISSUE #3256: [Bug] Dev detached tmux launches fail during pane-id because guarded commands escape the separator
## Summary

Detached tmux launches on the current `dev` branch fail immediately after creating the leader pane:

```text
Error: detached launch safety failure during pane-id
[omx] detached launch safety report: {"phase":"pane-id","transitions":["D0","D1"],"rollback":{"attempted":["rollback","parent-env","runtime-codex-home","rollback","session:kill-session"],"failures":[{"step":"session:kill-session","status":"failed","code":"unknown"}]}}
```

The failure appears to be a regression introduced by commit [`d6105f51`](https://github.com/Yeachan-Heo/oh-my-codex/commit/d6105f51891c3e2a70abb0cfbefdbe663a93ffd9), `fix(tmux): bind exact pane authority at mutation sinks`.

`runDetachedLeaderMutation()` constructs a tmux command sequence using an escaped separator:

```ts
const success =
  `${args.map(quoteShellArg).join(" ")} \\; display-message -p ${quoteShellArg(receipt)}`;
```

Because this string is passed directly to tmux through `execFileSync`, rather than first being parsed by a shell, tmux receives `\;` literally. With tmux 3.7b, it becomes another argument to `set-option` instead of separating the commands:

```text
command set-option: too many arguments (need at most 2)
```

Using an unescaped `;` in the same command succeeds and returns the expected receipt.

## Environment

- OS: macOS, Darwin 25.5.0, arm64
- tmux: 3.7b
- Node.js: v24.14.0
- Bun: 1.3.14
- `oh-my-codex`: `0.20.3`, installed from `github:Yeachan-Heo/oh-my-codex#dev`
- Exact dev revision: `e4c67b0f73a7bd13321f64b50f464177e0dc2afb`
- Codex CLI: `0.144.6`
- `omx doctor`: 20 passed, 0 warnings, 0 failed

The same failure was originally observed with an npm-installed build of the same dev revision, then reproduced after installing the revision through Bun. The package manager does not affect the failure.

## Steps to reproduce

1. Install the current dev branch:

   ```shell
   bun install -g 'github:Yeachan-Heo/oh-my-codex#dev'
   ```

2. From an interactive terminal where tmux is available, force a detached launch:

   ```shell
   omx --madmax --xhigh --tmux
   ```

3. Observe:

   ```text
   [omx] madmax isolated state: /Users/example/.omx-runs/run-... (source: ...)
   Error: detached launch safety failure during pane-id
   [omx] detached launch safety report: {"phase":"pane-id","transitions":["D0","D1"],"rollback":{"attempted":["rollback","parent-env","runtime-codex-home","rollback","session:kill-session"],"failures":[{"step":"session:kill-session","status":"failed","code":"unknown"}]}}
   ```

`--xhigh` and Codex itself are not involved yet; the failure happens during OMX's detached tmux bootstrap.

## Expected behavior

OMX should:

1. create the detached leader session;
2. tag the session with `@omx_instance_id`;
3. set the history limits;
4. create and finalize the HUD pane;
5. release the detached-leader barrier;
6. attach to the resulting session.

## Actual behavior

The leader session and pane are created successfully, but the first guarded `tag-session` mutation fails.

The detached leader then remains alive until its release timeout. A retained release report from the failed run contains:

```json
{
  "version": 1,
  "kind": "failed",
  "finalized": true,
  "error": "detached leader release timed out"
}
```

After the timeout, the pane exits and tmux removes the session, which is why a later `omx cleanup` reports no orphaned MCP processes or stale temporary directories.

## Isolated tmux reproduction

The tmux parsing problem can be reproduced independently of Codex.

Given a valid leader-pane authority condition, running the command sequence generated by `runDetachedLeaderMutation()` produces:

```text
broken_status=1
broken_output=command set-option: too many arguments (need at most 2)
```

The generated success command has this form:

```text
'set-option' '-t' 'omx-issue-repro' '@omx_instance_id' 'owner' \; display-message -p 'receipt'
```

Changing only the separator to an unescaped semicolon:

```text
'set-option' '-t' 'omx-issue-repro' '@omx_instance_id' 'owner' ; display-message -p 'receipt'
```

produces:

```text
fixed_status=0
fixed_output=receipt
```

This was tested against a separate disposable tmux server created with `tmux -L`, so no existing tmux session or configuration was involved.

## Suspected source path

The first failing call is the `tag-session` branch:

https://github.com/Yeachan-Heo/oh-my-codex/blob/dev/src/cli/index.ts#L6081-L6084

It calls:

https://github.com/Yeachan-Heo/oh-my-codex/blob/dev/src/cli/index.ts#L1543-L1550

The same `\\; display-message` construction currently appears in three places:

- `runDetachedLeaderMutation()`
- `runDetachedHudMutation()`
- `buildDeferredDetachedHudGuard()`

At minimum, the direct `runDetachedLeaderMutation()` command requires the correct separator for the actual tmux parsing layer. The two nested/deferred cases should also be audited for their respective parsing depth.

## Why the safety report looks like this

The report matches this failure path:

1. `new-session` succeeds.
2. `captureDetachedLeaderAuthority()` succeeds.
3. The next bootstrap step is `tag-session`.
4. `runDetachedLeaderMutation()` invokes the malformed tmux command.
5. The bootstrap catch maps every non-`new-session` failure to `pane-id`.
6. Rollback tries to call `kill-session` through the same guarded mutation.
7. Since the owner tag was never committed—and the guarded command construction is also affected—the rollback mutation returns no receipt.
8. That plain `Error` has no recognized system error code, producing `"code":"unknown"`.

The displayed `pane-id` phase is therefore misleading: pane capture completed successfully, and the actual failure is the guarded tag-session mutation.

## Regression/test gap

The fake-tmux wrapper in `launch-fallback.test.ts` recognizes `if-shell`, extracts the receipt text from the success command, and returns that receipt without asking a real tmux parser to execute the command string.

Consequently, the test accepts `\;` even though real tmux rejects the generated command.

A regression test should either:

- run a small isolated real-tmux smoke test when tmux is available; or
- use a parser-realistic fixture that fails when an escaped separator is passed at the wrong parsing layer.

It would also help if `renderDetachedLaunchSafetyReport()` included the underlying `DetachedLaunchSafetyError.cause`. In this case, the actual tmux message (`command set-option: too many arguments`) is suppressed, leaving only the secondary phase and rollback symptoms.

Yeachan-Heo:
Fixes #3256.

## Summary
- Correct detached tmux receipt command-list separators.
- Preserve literal pane formats and real TAB delimiters through the deferred outer-parser boundary.
- Add isolated private-server stored-hook provenance and lifecycle cleanup coverage.

## Verification
- `npm test`
- Focused private tmux stored-hook regression
- Full `launch-fallback` suite: 32/32
- Runtime relaunch regression

—

### DIFF
--- a/src/cli/__tests__/launch-fallback.test.ts
+++ b/src/cli/__tests__/launch-fallback.test.ts
@@ -1,4 +1,4 @@
-import { describe, it } from 'node:test';
+import { describe, it, type TestContext } from 'node:test';
 import assert from 'node:assert/strict';
 import { existsSync } from 'node:fs';
 import { chmod, mkdir, mkdtemp, readFile, readdir, rename, rm, writeFile } from 'node:fs/promises';
@@ -6,7 +6,7 @@ import { execFileSync, spawn, spawnSync } from 'node:child_process';
 import { tmpdir } from 'node:os';
 import { basename, dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { HUD_TMUX_HEIGHT_LINES } from '../../hud/constants.js';
+import { HUD_RESIZE_RECONCILE_DELAY_SECONDS, HUD_TMUX_HEIGHT_LINES } from '../../hud/constants.js';
 import { DETACHED_TMUX_HISTORY_LIMIT } from '../index.js';
 import {
   closeLaunchSessionBindingOnce,
@@ -15,9 +15,9 @@ import {
   writeSessionStart,
   type LaunchSessionBinding,
 } from '../../hooks/session.js';
+import { isRealTmuxAvailable, withTempTmuxSession } from '../../team/__tests__/tmux-test-fixture.js';
 
 const CLI_SPAWN_TIMEOUT_MS = 60_000;
-
 function buildRunOmxEnv(envOverrides: Record<string, string>): NodeJS.ProcessEnv {
   const env: NodeJS.ProcessEnv = { ...process.env };
   for (const key of Object.keys(env)) {
@@ -75,7 +75,6 @@ function shouldSkipForSpawnPermissions(_err: string): false {
   // not a pass. Node's test runner has no portable dynamic skip here.
   return false;
 }
-
 
 async function createGitRepo(wd: string): Promise<string> {
   const repo = join(wd, 'repo');
@@ -273,6 +272,19 @@ async function waitForPath(path: string, expectedLines: number = 1): Promise<voi
   throw new Error(`timed out waiting for ${path}`);
 }
 
+function parseShimTmuxArgv(contents: string): string[][] {
+  return contents
+    .split('tmux argv:\n')
+    .slice(1)
+    .map((record) => record.split('\nend tmux argv')[0]!.split('\n').filter(Boolean));
+}
+
+async function assertShimLogQuiescent(path: string, quietPeriodMs: number, message: string): Promise<void> {
+  const before = await readFile(path, 'utf-8').catch(() => '');
+  await new Promise((resolve) => setTimeout(resolve, quietPeriodMs));
+  assert.equal(await readFile(path, 'utf-8').catch(() => ''), before, message);
+}
+
 async function stopHeldOmx(child: ReturnType<typeof spawn>, releasePath: string): Promise<void> {
   await rm(releasePath, { force: true });
   await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


Yeachan-Heo:
Exact-head review: **CLEAR / APPROVE** for `fb4db8a4ea91ad83a44adccb730e6295f0588057`.

The stored-hook execution is isolated from launch-time reconciliation: launch records settle and the shim is quiescent before baseline clearing; the subsequent ordinary-PTY resize produces exactly the two post-baseline nested `list-panes` calls with literal `#{pane_id}`, `#{pane_dead}`, `#{pane_pid}`, and real TAB delimiters. Authority-denial, foreign-session isolation, receipt semantics, and exact cleanup expectations remain covered.

—

Issue #3270, opened 2026-07-23
[Bug] Ralplan handoff always fails when the host consensus receipt verifier is unavailable


Ox24:
## Summary

With native role routing available, Autopilot can complete its Ralplan lifecycle with sequential Architect and Critic approvals, but it still cannot hand off to Ultragoal. The run terminalizes with `documented_host_consensus_receipt_unavailable` because the installed consensus resolver requires an official host-issued receipt while no documented verifier is available.

This is independent of flattened collaboration-tool compatibility: native role-specific spawns complete successfully, and the relevant flattened-name regression tests pass. The blocker occurs only at the consensus-to-execution transition.

## Environment

- OS: macOS 26.5.2 (arm64)
- Node.js version: v26.0.0
- `oh-my-codex` version: v0.20.3-dev-bb124b29b7ee (plugin mode)
- Codex CLI version: 0.145.0

## Steps to reproduce

1. Start a clean Codex session with OMX plugin mode and native `agent_type` role routing available.
2. Run `$autopilot` for any concrete task that requires planning and complete the deep-interview gate.
3. In Ralplan, run the native `architect` review and wait for an `APPROVE` verdict.
4. Only after Architect completes, run the native `critic` review and wait for an `APPROVE` verdict.
5. Attempt the normal Ralplan-to-Ultragoal handoff, or inspect the terminal Ralplan state with `omx state read --mode ralplan --json`.

## Expected behavior

After the required sequential reviews approve, OMX should have a documented host surface that issues and verifies the required consensus receipt so the normal execution handoff can proceed.

If that verifier is not implemented or available on the current surface, OMX should fail during preflight before running the expensive review lifecycle, or use another explicitly documented safe behavior. A default Autopilot run should not reach an approval state that can never advance.

## Actual behavior

Architect and Critic both approve, but the consensus gate remains `complete:false`. The run terminalizes as failed with `documented_host_consensus_receipt_unavailable`.

The installed resolver currently treats native reviews and tracking as lifecycle evidence only and returns the unavailable-host-receipt blocker for every local evidence source. There is no documented user-accessible action that can satisfy the gate, and user-authored receipt-shaped fields are correctly rejected.

## Logs / screenshots

Sanitized terminal state:

```json
{
  "active": false,
  "current_phase": "failed",
  "run_outcome": "failed",
  "error": "documented_host_consensus_receipt_unavailable",
  "ralplan_architect_review": {
    "agent_role": "architect",
[...]

PR #3273, opened 2026-07-23
fix(hooks): drain oversized native hook stdin

chief-impact7:
## Summary

Codex writes each native hook event as a complete stdin stream. The current 1 MiB guard destroys stdin as soon as the retained parsing limit is reached, which can make the parent writer fail with `Broken pipe (os error 32)`.

## Changes

- keep the existing 1 MiB parsing cap
- discard bytes beyond the cap while draining the stream to EOF
- add a child-process regression test that fails with `write EPIPE` before the fix
- document the bounded-retention/full-drain contract

## Validation

- [x] `npm run build`
- [x] `npx biome lint src/scripts/codex-native-hook.ts src/scripts/__tests__/codex-native-hook.test.ts`
- [x] new oversized parent-writer regression test
- [x] oversized native-hook tests: 6 passed on the original base before rebasing to `dev`
- [ ] full native-hook file: unrelated existing `allows Ralplan Markdown draft-only...` failure on macOS because an absolute `/var/folders/.../.omx/drafts` path is rejected

## Checklist

- [x] Native hook documentation updated
- [x] Backward compatibility preserved: oversized payload handling remains fail-closed

## Related

No existing issue found for the exact `failed to write hook stdin: Broken pipe` error.

### DIFF
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -29,6 +29,10 @@ Project-scoped resume and search discovery include generated runtime homes under
 
 `omx doctor` can confirm that these files exist and are shaped correctly. It does not prove that the same shell/profile can complete an authenticated Codex request; use `codex login status` plus a real `omx exec --skip-git-repo-check -C . "Reply with exactly OMX-EXEC-OK"` smoke test for that boundary.
 
+The native hook CLI retains at most 1 MiB of stdin for parsing. It still drains
+the complete stream when that limit is exceeded so Codex can finish writing the
+event payload without receiving `EPIPE`.
+
 ## Ownership split
 
 - **Plugin-scoped Codex hooks**: `plugins/oh-my-codex/hooks/hooks.json` for plugin installs on Codex versions with `[features].plugin_hooks`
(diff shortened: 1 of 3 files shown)


Yeachan-Heo (approved):
MERGE_READY

Reviewed exact head `59bf043b1710726ec45b80ea16c53ff02c3fdaf6` against base `bb124b29b7eec69d2f33b9f8520ee6b59ebc3deb`.

The three-file change is narrow and matches the public native-hook contract. Replacing early `process.stdin.destroy()` with continued iteration preserves the retained 1 MiB parsing bound while applying stream backpressure and draining to EOF, so the parent writer can complete without `EPIPE`. Oversized input remains unparsed; UTF-8 truncation is confined to the retained diagnostic/event-name prefix; oversized PreToolUse and generic events remain fail-closed; oversized Stop keeps the existing active-workflow block and inactive no-op behavior. Reader lifetime now correctly follows writer EOF, with no extra retained payload growth. The regression test exercises a 2 MiB parent-to-child pipe and checks both writer completion and child exit; its one-shot promises are safe against duplicate resolution and do not introduce a platform-specific shell dependency.

Novelty/scope check found the original 1 MiB guard in #2615/#2618 and no duplicate drain fix on current dev. The active #3270 Autopilot consensus-preflight lane concerns lifecycle preflight/terminalization, not stdin intake; there is no semantic overlap beyond sharing the large native-hook module.

Exact-head CI run 30005021662 concluded success, including typecheck, no-unused, lint, build, the hooks/notify/platform test lane, Node 20 grouped lanes, Node 22 smoke, coverage, docs, and the macOS Ralplan preflight gate. That exact evidence supersedes the contributor's reported local macOS draft-path failure: it is not reproduced as a regression on this head.

No blocking correctness, race, hang, portability, scope, or contract issue found.

—

PR #3276, opened 2026-07-23
fix(auth): validate metadata before credential switch

### LINKED ISSUE #3275: [Bug] auth use can fail after replacing live credentials when slots metadata is invalid
## Summary

`omx auth use <slot>` can exit with an error **after it has already replaced the live Codex `auth.json`**.

`useSlot()` writes the selected credential file first and only then parses `~/.omx/auth/slots.json`. If that metadata file is malformed, or contains a legacy/invalid slot name, metadata parsing throws after the credential switch has committed. The CLI exits 1 even though the active account changed.

This makes a failed, security-sensitive command leave a surprising side effect and breaks failure atomicity for a deterministic preflight error.

## Environment

- OS: macOS 26.6 (darwin arm64)
- Node.js: v20.11.1
- oh-my-codex: current `dev` at `bb124b29`
- Codex CLI: 0.145.0

## Steps to reproduce

```bash
root="$(mktemp -d)"
mkdir -p "$root/home/.codex" "$root/home/.omx/auth"

printf '%s\n' '{"access_token":"old"}' > "$root/home/.codex/auth.json"
printf '%s\n' '{"access_token":"new"}' > "$root/home/.omx/auth/work.json"
printf '%s\n' \
  '{"version":1,"slots":[{"slot":"../bad","createdAt":"x","updatedAt":"x"}]}' \
  > "$root/home/.omx/auth/slots.json"

HOME="$root/home" \
CODEX_HOME="$root/home/.codex" \
OMX_AUTO_UPDATE=0 \
node dist/cli/omx.js auth use work

echo "exit=$?"
cat "$root/home/.codex/auth.json"
```

## Expected behavior

The command exits nonzero and preserves the original live credentials because the existing metadata is invalid.

## Actual behavior

The command exits 1 with:

```text
Error: invalid auth slot name: use 1-64 letters, numbers, '.', '_' or '-' and start with a letter or number
```

but the live file already contains:

```json
{"access_token":"new"}
```

Malformed JSON in `slots.json` produces the same partial commit.

## Root cause

In `src/auth/storage.ts`, `useSlot()` calls `atomicWriteFile(liveAuthPath, ...)` before `readAuthMetadata(home)`. The latter can deterministically reject while parsing or validating metadata.

## Suggested fix

Read and validate the existing metadata before replacing live `auth.json`, then reuse that parsed value for the metadata update. Add regression tests for malformed JSON and invalid slot metadata asserting that the original live auth remains unchanged.

lux-02:
## Summary

`omx auth use <slot>` could report failure after it had already replaced the live Codex credentials. `useSlot()` parsed and validated `slots.json` only after committing the live `auth.json` replacement, so malformed or legacy-invalid metadata produced a partial commit.

This change performs that deterministic metadata preflight before mutating live credentials.

## Changes

- Read and validate auth slot metadata before replacing live `auth.json`.
- Reuse the validated metadata for the existing `currentSlot`/`lastUsedAt` update.
- Add regression coverage for malformed JSON and invalid slot-name metadata, asserting that failed switches preserve the original live credentials.

## Validation

- [x] `npm run build`
- [x] `node --test dist/auth/__tests__/storage.test.js` — 8/8 pass
- [x] `npm run lint`
- [ ] `npm test` — attempted in both sandboxed and elevated local runs; the elevated run reached unrelated macOS-only failures in `launch-fallback.test.js` because BSD `script` rejects the test fixture's `-c` flag. The focused auth suite and all static gates above are green; full Linux coverage is left to PR CI.
- [ ] `omx doctor` — not applicable; setup/config behavior is unchanged

## Checklist

- [x] Docs updated when needed — not needed; this preserves the documented CLI behavior
- [x] Backward-compatibility impact considered — successful slot switching is unchanged

## Related

Closes #3275

### DIFF
--- a/src/auth/__tests__/storage.test.ts
+++ b/src/auth/__tests__/storage.test.ts
@@ -5,7 +5,13 @@ import { mkdir, mkdtemp, readFile, rm, stat, symlink, writeFile } from "node:fs/
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { atomicWriteFile, addSlotFromAuthFile, listSlots, readAuthMetadata, useSlot } from "../storage.js";
-import { resolveLiveAuthPath, resolveOmxAuthDir, resolveSlotPath, validateSlotName } from "../paths.js";
+import {
+  resolveAuthMetadataPath,
+  resolveLiveAuthPath,
+  resolveOmxAuthDir,
+  resolveSlotPath,
+  validateSlotName,
+} from "../paths.js";
 
 async function tempHome(): Promise<string> {
   return await mkdtemp(join(tmpdir(), "omx-auth-storage-"));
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
MERGE_READY

Reviewed exact head `7e356da57aa24f10442cba93412dba10db526c5f` against exact base `fe3c1738b60e015767b9aa03db1b4a7f7e9f85fd`.

The patch correctly moves metadata parsing ahead of the live credential replacement, so deterministic malformed-JSON and invalid legacy-slot metadata failures occur before live auth mutation. Independent reproductions confirmed that both failures preserve the original live auth bytes and mode, preserve metadata and the selected slot, and do not expose credential sentinels. The successful path still writes the selected credentials with mode 0600 and preserves current-slot, last-used, quota, exhaustion-clear, and sibling-slot bookkeeping behavior.

Ordering and contract review found no blocker: selected-slot readability remains first; metadata is validated before `ensurePrivateDir` and live-target checks; live symlinks remain rejected; `atomicWriteFile` retains temp-file/rename semantics and owner-only mode; metadata is written only after the live replacement. This fixes deterministic preflight atomicity, but intentionally does not make the later live-auth plus metadata writes a transactional two-resource operation—an I/O failure after the live rename can still leave metadata stale. The moved metadata read modestly widens existing race windows, but introduces no new path traversal, symlink-following, mode, or actionable TOCTOU regression under the current single-writer contract.

Caller review covered direct CLI use and initial/later hotswap rotations. Focused storage, hotswap, and compiled CLI suites passed (23 tests), including token-redaction coverage. Static gates passed: build, lint, no-unused typecheck, and `tsc --noEmit`. `git diff --check` passed. The change matches issue #3275's failure-atomicity scope without adding a competing implementation.

Exact-head CI run 30009651852 is terminal green (`success`) with no pending or failed jobs, and PR head/base remain the reviewed SHAs.

—

PR #3279, opened 2026-07-23
fix(native-assets): validate cached binaries before reuse

### LINKED ISSUE #3278: [Bug] native binary cache returns corrupted or partial entries without validation
## Summary

`hydrateNativeBinary()` and the `omx sparkshell` / `omx api` resolver fast paths treat anything that merely exists at a native cache path as a valid binary.

A zero-byte file, a truncated file, or even a directory is returned without validation. Because the release manifest is never requested on a cache hit, a valid release asset cannot repair the cache. The bad path is reused until the user manually finds and removes it.

The cache writer can also create this state itself: it copies the extracted binary directly to the final cache path rather than publishing a completed temporary file with an atomic rename.

## Environment

- OS: macOS 26.6 (darwin arm64)
- Node.js: v26.0.0
- oh-my-codex: current `dev` at `12da62941ff6`
- Reproduced with an isolated `OMX_NATIVE_CACHE_DIR` and localhost manifest/archive fixture

## Reproduction

For each scenario, place the entry at the exact path returned by `resolveCachedNativeBinaryPath()`, then call `hydrateNativeBinary()` while a valid manifest and archive are available:

1. zero-byte regular file
2. non-empty truncated file
3. directory occupying the binary path

Observed across three repeated runs:

```text
control-fresh-download: downloaded valid binary (server hits: 2)
empty-cache-file: bad cache returned unchanged (server hits: 0)
truncated-cache-file: bad cache returned unchanged (server hits: 0)
directory-at-cache-path: directory returned as the binary (server hits: 0)
```

The same cache-first shortcut is present in `resolveSparkShellBinaryPathWithHydration()` and `resolveApiBinaryPathWithHydration()`, so normal CLI execution can bypass hydration before it has a chance to recover.

## Expected behavior

- Cache entries should be accepted only when they are regular, non-empty files whose persisted checksum matches.
- Unverified legacy entries should be refreshed when the release asset is available, while remaining usable as an offline fallback when the manifest/archive cannot be reached.
- A hydrated binary and its checksum should be fully written to same-directory temporary files before publication, so interruption cannot expose a partial final binary.

## Actual behavior

- Existing cache entries are checked with `existsSync()` only.
- New downloads validate the archive size and SHA-256, but the extracted cached binary has no persisted integrity record.
- The extracted binary is copied directly to the final cache path.

Relevant code on the affected base:

- `src/cli/native-assets.ts`: cache fast path and direct final-path `copyFile()`
- `src/cli/sparkshell.ts`: cache fast path before packaged/repo-local/hydration fallback
- `src/cli/api.ts`: equivalent cache fast path

## User impact

A process interruption, disk-full condition, or damaged cache can leave `omx sparkshell` / `omx api` permanently unable to start. Subsequent launches keep selecting the bad cache entry and never attempt the valid release download. The error path does not tell the user which cache entry to delete.

## Suggested fix

- Persist a SHA-256 sidecar for hydrated cache binaries.
- Verify regular-file status, non-zero size, and sidecar digest before reuse.
- Rehydrate unverified/mismatched entries when assets are available.
- Preserve a structurally valid legacy cache only as an offline fallback.
- Copy and chmod into unique same-directory temporary files, then rename the binary and checksum into place.
- Apply the same verified-cache predicate to the sparkshell and api resolver fast paths.
- Add regressions for empty, truncated, directory, verified-cache, legacy-offline, and resolver-bypass cases.

## Duplicate check

No existing issue or PR covers native hydrated-binary cache integrity or interrupted final-path publication. The closest items address different causes/components:

- #528: npm self-update integrity
- #812: GLIBC compatibility
- #2066 / #2068: explore checkout detection

lux-02:
## Summary

Native cache fast paths accepted any existing path, so empty/truncated files or directories could bypass hydration permanently. Hydration also copied extracted binaries directly to the final cache path, allowing interrupted writes to expose partial cache entries.

This change gives hydrated binaries a verifiable cache record and publishes them through same-directory temporary files.

## Changes

- Persist a SHA-256 sidecar for each successfully hydrated native binary.
- Require regular-file, non-empty, checksum-matching cache entries in the sparkshell and api resolver fast paths.
- Refresh unverified legacy or checksum-mismatched entries when a release asset is available.
- Preserve structurally valid legacy entries when auto-fetch is disabled or the manifest/archive is unavailable, retaining offline compatibility.
- Copy, chmod, and checksum unique same-directory temporary files before renaming them into the cache.
- Add regressions for verified cache hits, empty/truncated/directory recovery, legacy offline fallback, and resolver cache bypasses.

## Validation

- [x] `npm run build`
- [x] Focused compiled suites:
  - `node --test dist/cli/__tests__/native-assets.test.js dist/cli/__tests__/sparkshell-cli.test.js dist/cli/__tests__/api.test.js` — 47/47 pass
- [x] Original isolated reproduction, repeated 3 times — control passed and 0/3 bug scenarios remained on every run
- [x] `npm run lint`
- [x] `npx tsc --noEmit`
- [x] `git diff --check`
- [ ] `npm test` — attempted locally. The changed native suites passed when reached, but the broader macOS run reported unrelated existing environment-sensitive failures, including BSD `script -c` incompatibility and `/var` vs `/private/var` path assertions, plus active-session auth fixture interference. The run was stopped after those unrelated failures were established; Linux PR CI remains authoritative.
- [ ] `omx doctor` — not applicable; setup/config behavior is unchanged

## Compatibility

Sidecarless cache entries are refreshed once when release assets are reachable. If auto-fetch is disabled or the manifest/archive cannot be reached, a structurally valid legacy entry remains available as the fallback. Successful verified cache hits remain network-free.

## Checklist

- [x] Docs updated when needed — not needed; this preserves the existing CLI contract and changes only internal cache validation/publication

## Related

Closes #3278

### DIFF
--- a/src/cli/__tests__/api.test.ts
+++ b/src/cli/__tests__/api.test.ts
@@ -1,5 +1,6 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
+import { createHash } from 'node:crypto';
 import { chmod, mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import { tmpdir } from 'node:os';
@@ -13,6 +14,7 @@ import {
   resolveApiBinaryPathWithHydration,
   runApiBinary,
 } from '../api.js';
+import { resolveCachedNativeBinaryChecksumPath } from '../native-assets.js';
 
 function runOmx(
   cwd: string,
@@ -101,8 +103,13 @@ describe('resolveApiBinaryPath', () => {
       const cachedDir = join(cacheDir, '0.8.15', 'linux-x64-musl', 'omx-api');
       const cachedBinary = join(cachedDir, 'omx-api');
       await mkdir(cachedDir, { recursive: true });
-      await writeFile(cachedBinary, '#!/bin/sh\n');
+      const cachedContent = '#!/bin/sh\n';
+      await writeFile(cachedBinary, cachedContent);
       await chmod(cachedBinary, 0o755);
+      await writeFile(
+        resolveCachedNativeBinaryChecksumPath(cachedBinary),
+        `${createHash('sha256').update(cachedContent).digest('hex')}\n`,
+      );
 
       assert.equal(
         await resolveApiBinaryPathWithHydration({
@@ -118,6 +125,8 @@ describe('resolveApiBinaryPath', () => {
       await rm(cwd, { recursive: true, force: true });
     }
   });
+
+
 });
 
 describe('runApiBinary', () => {
--- a/src/cli/__tests__/native-assets.test.ts
+++ b/src/cli/__tests__/native-assets.test.ts
@@ -1,7 +1,7 @@
 import { createServer } from 'node:http';
 import { describe, it } from 'node:test';
 import assert from 'node:assert/strict';
-import { chmod, mkdtemp, mkdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { chmod, mkdtemp, mkdir, readFile, readdir, rm, writeFile } from 'node:fs/promises';
 import { createHash } from 'node:crypto';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
@@ -11,6 +11,7 @@ import {
   inferNativeAssetLibc,
   isRepositoryCheckout,
   resolveCachedNativeBinaryCandidatePaths,
+  resolveCachedNativeBinaryChecksumPath,
   resolveCachedNativeBinaryPath,
   type NativeReleaseManifest,
   resolveNativeReleaseAssetCandidates,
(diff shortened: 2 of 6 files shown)


Yeachan-Heo (changes requested):
REQUEST_CHANGES on exact head `1dfaf1c165e447201acbbc750c329953ed464cff`.

Blocking findings:

1. **Legacy non-empty corruption is still executable offline** (`src/cli/native-assets.ts:401-419`, especially 407 and 413). Any regular non-zero-byte cache entry without a sidecar is classified as `legacy`, and `hydrateNativeBinary()` returns it when auto-fetch is disabled, the manifest is unavailable, or no matching asset exists. That includes the original truncated-file class whenever truncation leaves at least one byte. With no packaged/repo-local binary, the resolver returns that path for execution. The new test explicitly blesses arbitrary non-empty bytes as the compatibility contract, but this does not meet #3278's requirement to stop reusing truncated cache artifacts. **Required:** either make offline legacy reuse opt-in with an explicit unsafe compatibility contract, or structurally/authentically validate legacy entries so corrupted non-empty files cannot be selected; add an end-to-end resolver test proving a non-empty truncated legacy artifact is not executed.

2. **The verifier follows symlinks for both authority files** (`src/cli/native-assets.ts:209-225`). `stat()` and `readFile()` follow links, so a symlinked binary plus symlinked `.sha256` is accepted as `verified`; a symlinked legacy binary is also accepted as a regular non-empty file. Publication also uses `stat()` while clearing directory entries (`src/cli/native-assets.ts:368-375`), so it does not establish a no-symlink invariant. This makes the sidecar self-consistency check attacker-controlled rather than authoritative for a cache-owned regular file. **Required:** use `lstat`/no-follow checks for the binary and sidecar, reject symlinks and non-regular files, re-check the published pair safely, and add binary-symlink, sidecar-symlink, and symlink/directory replacement tests on supported platforms.

3. **Archive/manifest paths are not confined to the hydration staging root** (`src/cli/native-assets.ts:325-348, 421-453`). `asset.archive` can contain `../` and escape `tempRoot`; `asset.binary_path` is joined directly and the exact-candidate fast path can resolve outside `extractDir`. Extraction itself delegates to `tar`/`unzip` without validating entry paths or link entries, and `findExtractedBinaryPath()` treats non-directory entries, including symlinks, as binary candidates. **Required:** reject absolute/traversing manifest paths, preflight archive entries against traversal and unsafe link types, require the selected extracted binary to be a regular no-follow file contained under `extractDir`, and test traversal, absolute paths, duplicate/ambiguous binary selection, and symlink entries.

Publication also needs adversarial coverage for the two-rename window: crash after binary rename, crash after sidecar rename attempt, concurrent hydrators, replacement of an existing valid pair, Windows/POSIX rename behavior, modes, and temp cleanup. The current tests exercise sequential happy-path repair only and do not demonstrate transaction-safe pair authority.

Verification performed locally at this exact head:
- focused native-assets / sparkshell / api tests: pass (47 tests total)
- build: pass
- Biome lint: pass
- no-unused TypeScript check: pass
- package bin contract and sparkshell packaging tests: pass
- `npm pack --dry-run --json --ignore-scripts`: pass

GitHub CI run 30018745142 is still active, so MERGE_READY is unavailable independently of the blockers above.

—

Issue #3284, opened 2026-07-23
[Bug] Persisted subagent reopen misclassifies the root session as a resumable subagent


arikon:
## Summary

Persisted subagent reopen can classify the current root Codex session itself as a saved subagent. On the next `SessionStart source: resume`, OMX instructs the root agent to call `resume_agent()` with its own root session ID, and the native tool can return `{"status":"running"}`.

The durable tracker shows the underlying identity inversion: a turn ID is stored as `leader_thread_id`, while the actual user/root session ID is stored as `kind:"subagent"`.

This was reproduced in two independent Codex App user/root sessions six days apart. It is related to native subagent provenance, but it is not the same direction as #2734: #2734 misclassified a fresh child as leader during Ralplan consensus; this bug misclassifies the root session as a resumable child during persisted reopen.

## Environment

- OS: macOS 26.5 arm64
- Node.js version: v22.23.1 (current reproduction environment)
- `oh-my-codex` version: v0.20.3 for the current reproduction; the earlier reproduction was on the v0.20.x line
- Codex CLI version: 0.144.5 and 0.145.0-alpha.27
- Surface: Codex App / VS Code-originated native-hook session, outside tmux

## Steps to reproduce

1. Start a normal root Codex session whose rollout metadata contains:

   ```json
   {
     "thread_source": "user",
     "session_id": "<root-session-id>"
   }
   ```

2. Use native subagents so OMX persists `subagent-tracking.json`.
3. Let the root session stop and reopen/resume it.
4. Observe the injected `Persisted subagent reopen` instructions.
5. Inspect the saved IDs and the tracker.
6. Call the requested `resume_agent()` operations.

## Expected behavior

- The current root/user session ID must never appear in `savedSubagents`.
- The durable tracker must store the actual root session/thread identity as leader and only true native descendants as `kind:"subagent"`.
- `resume_agent(<current-root-id>)` should be rejected or filtered before the tool call.
- Reopen should resume only distinct child handles that belong to the current root session.

## Actual behavior

First independent reproduction:

```json
// rollout session_meta
{
  "session_id": "019f6e0c-...-42a4",
  "thread_source": "user",
  "cli_version": "0.144.5"
}
```

The reopened root received its own ID in the saved-subagent list and executed:

```text
resume_agent("019f6e0c-...-42a4")
=> {"status":"running"}
```

The tracker stored:

```json
{
  "session_id": "019f6e0c-...-42a4",
[...]

PR #3296, opened 2026-07-25
fix(tmux): guard deferred HUD resize sinks in command-list context (#3292)

### LINKED ISSUE #3292: [Bug] Detached tmux HUD expands to 7–23 rows because guarded resize hook runs invalid command
## Summary

When OMX launches Codex in a newly created detached tmux session, the bottom HUD pane can occupy a large portion of the terminal instead of its configured 2 rows.

This is particularly visible with:

```sh
omx --tmux --madmax --high
```

or:

```sh
omx --tmux --madmax --xhigh
```

The reasoning flags are not themselves responsible. They only make this detached launch path easy to encounter.

Screenshot showing the enlarged OMX HUD pane

## Environment

```text
oh-my-codex package version: 0.20.3
source commit: 99d87f36
source describe: v0.20.3-73-g99d87f36
tmux: 3.7b
Codex CLI: 0.145.0
OS: macOS arm64
observed client size: 178x35
```

## Steps to reproduce

1. Open a terminal that is not already inside tmux.
2. Use a terminal substantially taller than 24 rows, for example `178x35`.
3. Run:

   ```sh
   omx --tmux --madmax --high
   ```

4. From another terminal, inspect OMX's panes:

   ```sh
   tmux list-panes -a \
     -F 'session=#{session_name} pane=#{pane_id} top=#{pane_top} height=#{pane_height} bottom=#{pane_bottom} window=#{window_width}x#{window_height} command=#{pane_start_command}'
   ```

5. Locate the pane whose start command contains `hud --watch` or `OMX_DETACHED_HUD_OPERATION`.

## Actual behavior

The HUD is larger than its configured 2 rows.

Two live examples from separate OMX sessions:

```text
client: 178x35
window: 178x34
leader: 25 rows
HUD:     8 rows
```

```text
client: 178x35
window: 178x34
leader: 26 rows
HUD:     7 rows
```

On an isolated `204x68` client, the same raw tmux layout mechanism expanded a 2-row HUD to 23 rows.

## Expected behavior

The OMX HUD should remain at `height=2` after initial client attachment and subsequent terminal resizing.

## Root cause

There are two parts to the failure.

### 1. Detached tmux geometry expands the HUD before correction

Detached bootstrap creates an unattached session without explicit dimensions:

```ts
// src/cli/index.ts:5067-5090
[
  "new-session",
  "-d",
  // no -x/-y
]
```

OMX then creates the HUD at 2 rows before attaching the real client:

```ts
// src/cli/index.ts:5092-5105
[
  "split-window",
  "-v",
  "-l",
  String(HUD_TMUX_HEIGHT_LINES),
]
```

The configured height is 2 in `src/hud/constants.ts:1`.

A minimal isolated measurement produced:

```text
before attach:
  leader=21
  HUD=2
  window=80x24

after attaching a 178x35 client:
  leader=26
  HUD=7
  window=178x34
```

### 2. The guarded resize hook executes an invalid tmux command

OMX installs attach/resize reconciliation hooks, but the detached guard matches a shell command that includes the `tmux` executable:

```ts
// src/cli/index.ts:1583-1585
/tmux resize-pane -t %[0-9]+ -y [1-9][0-9]*/g
```

It then places the complete match in the successful branch of a nested tmux `if-shell`:

```ts
// src/cli/index.ts:1574-1577
const success = `${match} ; display-message -p ...`;
const hudGuard = `if-shell ... ${quoteShellArg(success)} ...`;
```

That branch is a **tmux command list**, not a shell command. The resulting stored hook contains a branch equivalent to:

```text
tmux resize-pane -t %1 -y 2
```

At that command boundary, tmux interprets `tmux` as the name of an internal tmux command and returns:

```text
unknown command: tmux
```

Minimal demonstration:

```sh
socket=omx-command-context-repro

tmux -L "$socket" -f /dev/null \
  new-session -d -x 178 -y 34 -s repro 'sleep 300'

hud="$(
  tmux -L "$socket" split-window \
    -v -l 7 -d -P -F '#{pane_id}' -t repro 'sleep 300'
)"

tmux -L "$socket" if-shell -F -t "$hud" '1' \
  "tmux resize-pane -t $hud -y 2" \
  ""
```

Result:

```text
unknown command: tmux
```

The pane remains 7 rows. Removing the executable prefix works:

```sh
tmux -L "$socket" if-shell -F -t "$hud" '1' \
  "resize-pane -t $hud -y 2" \
  ""
```

The pane then becomes 2 rows.

The production hook hides the failure with:

```sh
>/dev/null 2>&1 || true
```

The one-shot `client-attached` hook subsequently unregisters, while the remaining `client-resized` hook continues to fail the same way. This leaves the enlarged HUD persistent.

Relevant current code:

- [`buildDeferredDetachedHudGuard()`](https://github.com/yeachan-heo/oh-my-codex/blob/99d87f36/src/cli/index.ts#L1568-L1590)
- [Detached session creation and HUD split](https://github.com/yeachan-heo/oh-my-codex/blob/99d87f36/src/cli/index.ts#L5067-L5105)
- [Attach/resize reconciliation registration](https://github.com/yeachan-heo/oh-my-codex/blob/99d87f36/src/cli/index.ts#L5171-L5214)
- [Guard application to all deferred HUD mutations](https://github.com/yeachan-heo/oh-my-codex/blob/99d87f36/src/cli/index.ts#L6162-L6175)

## Why `--high` / `--xhigh` appear related

They do not modify tmux geometry. They are translated only into:

```text
model_reasoning_effort="high"
model_reasoning_effort="xhigh"
```

`--madmax` is correlated because it commonly enters the isolated/detached launch flow. The underlying defect applies to detached tmux launches generally; explicit `--tmux` makes the reproduction deterministic.

## Test gap

The unit test verifies that the generated hook text contains `resize-pane`, but it does not execute the command in its final nested tmux command context:

- `src/team/__tests__/tmux-session.test.ts:466-490`

The real-tmux launcher test starts with the HUD already at 2 rows, removes the `client-attached` hook, and verifies snapshot transport rather than changing the HUD to a non-target height and asserting that the guarded mutation restores it:

- `src/cli/__tests__/launch-fallback.test.ts:935-1040`

Consequently, an invalid resize command can pass while the HUD remains 2 rows in the test's initial detached geometry.

## Suggested fix direction

When converting the shell resize command into a nested tmux command branch, remove the executable prefix:

```text
tmux resize-pane -t %1 -y 2
```

should become:

```text
resize-pane -t %1 -y 2
```

Alternatively, retain an explicit shell boundary with `run-shell`.

Seeding `new-session` with validated current terminal dimensions via `-x/-y` would reduce the initial layout jump, but the resize hooks still need to function for later client resizes.

## Acceptance criteria

- A detached session created at `80x24`, then attached to `178x35`, leaves the HUD at 2 rows.
- A test deliberately changes the HUD to 7 rows, triggers the stored resize hook, and observes 2 rows afterward.
- The hook does not emit `unknown command: tmux`.
- `--high` and `--xhigh` behave identically with respect to layout.
- Existing authority checks still prevent mutation of unrelated panes.

Yeachan-Heo:
Fixes #3292

## Summary

The detached tmux HUD stayed enlarged (7–23 rows) after a client attached, instead of returning to `HUD_TMUX_HEIGHT_LINES` (2). The stored `client-resized` reconciliation hook could never resize it.

The issue identified one root cause. Implementation found **two** defects in `buildDeferredDetachedHudGuard` (`src/cli/index.ts:1568`), and the second one *masked* the first — which is why the reported `unknown command: tmux` was only intermittently visible.

### Defect 1 — command-context error (the reported root cause)

The resize sink was spliced into a nested tmux `if-shell` success branch **including its `tmux ` executable prefix**. That branch is a tmux **command list**, not a shell, so tmux parsed `tmux` as a command name:

```
unknown command: tmux
```

The regex now captures the bare command and embeds only the capture:

```diff
-    /tmux resize-pane -t %[0-9]+ -y [1-9][0-9]*/g,
+    /tmux (resize-pane -t %[0-9]+ -y [1-9][0-9]*)/g,
```

This matches the sibling invariant already used by `runDetachedLeaderMutation` and `runDetachedHudMutation` (`src/cli/index.ts:1543-1566`), which embed bare tmux commands.

### Defect 2 — format-escaping error (found during implementation)

`run-shell` applies **exactly one outer tmux format pass** to the stored hook payload. Only the *original* command's `#{pane_*}` formats were escaped; the leader and HUD authority conditions **injected by `guardedResize`** were not. tmux therefore pre-expanded them against the hook's own pane and collapsed them to constants. The shim capture shows what actually reached tmux:

```
["if-shell","-F","-t","%0","1","if-shell -F -t '%1' '0' 'resize-pane -t %1 -y 2 ; ...'"]
```

Leader condition → literal `1`, HUD condition → literal `0`. The resize was denied **before the sink was ever parsed**. The injected conditions are now escaped so they survive the outer pass and are evaluated at execution time against the guarded panes.

### Why both are required

Real tmux 3.2a, detached 80x24, HUD enlarged to 7, natural first larger-client attach:

| sink | injected conditions | HUD after | `unknown command: tmux` |
|---|---|---|---|
| prefixed | unescaped *(= `dev` today)* | 7 ❌ | false *(masked)* |
| prefixed | escaped | 7 ❌ | **true** |
| bare | unescaped | 7 ❌ | false |
| **bare** | **escaped** *(this PR)* | **2** ✅ | false |

Defect 2 masks defect 1: with the conditions collapsed to `0`, the inner command list is never selected, so tmux never reaches the invalid token. Fixing only the prefix leaves the HUD enlarged; fixing only the escaping surfaces the reported error.

### Bonus: the fail-closed sentinel was bypassable

`if (guardedCommand === command)` compared the post-regex output against the **pre-escaping** original. A payload carrying `#{pane_id}`/`#{pane_dead}`/`#{pane_pid}` but **no** resize sink mutated through escaping alone and silently skipped the throw. Replaced with an explicit replacement counter, so the guard genuinely fails closed.

## Scope

One product function. Preserved exactly: the `/g` flag (both the immediate and the `sleep N`-delayed sink), the original command's `##{pane_*}` escaping, the receipt `; display-message -p` suffix, the outer `tmux if-shell` prefix (that layer *is* a shell context inside `run-shell -b`), and both authority-condition builders.

Only the four synchronous callers' conditions stay **unescaped** — correctly, since `execTmuxFileSync` passes them as direct argv with no format pass. Escaping there would have broken them.

**Not done** (explicit non-goals): `new-session -x/-y` seeding, `run-shell` re-wrap, authority/hook-registration refactor, reasoning-flag semantics changes, and issues #3284/#3289/#3290. One new `export` (`guardDetachedHudDeferredMutation`) as the direct test seam; both authority types stay module-private.

## Commits

| commit | purpose |
|---|---|
| `465502cd` | tests + the one visibility export — **product behavior unchanged**, so the discriminators are observably red |
| `6831e91f` | the product fix |
| `36a7f809` | fixture contract coverage added after the cleanup review |

`465502cd` is retained deliberately as the immutable pre-fix evidence commit.

## Verification

Verified on **tmux 3.2a** (reporter used 3.7b; the nested command-list semantics are version-independent).

**Pre-fix discriminators**, captured at `465502cd`:

- fail-closed unit test → `AssertionError: Missing expected exception`
- bare-sink companion → fails
- injected-condition escaping → fails
- real-tmux restore → **`7 !== 2`**

**Post-fix**: `index` 327/327 · `launch-fallback` 35/35 (3 consecutive runs) · `tmux-test-fixture` 7/7 · focused 6-file run all green · `lint` · `check:no-unused` · `verify:plugin-bundle` · `verify:native-agents` · `npm pack --dry-run` (zero forbidden paths).

**Adversarial suite, 22/22**: repeated resizes at three distinct geometries all restore to 2; idempotent re-fire; owner drift, `pane_pid` drift, and pane recycling all correctly **deny**; unrelated session byte-identical; 12 fail-closed bypass attempts all throw or guard correctly.

### Notes for the reviewer

- **AC3 is a post-fix guarantee, not a pre-fix discriminator.** Because defect 2 masks defect 1, the `unknown command: tmux` oracle passes *vacuously* on the unfixed tree — the error is unreachable, not absent-by-correctness. Stated plainly rather than presented as a matching matrix.
- `npm run smoke:packed-install` fails **identically on the pristine base** `99d87f36` (this workstation has codex-cli 0.145.0 vs the script's 0.142.5 boundary) — a pre-existing environment limitation, not a regression.
- The base `99d87f36` fails its own full `npm test` with 19 unrelated failures (MCP/Stop-hook/UserPromptSubmit suites), so full-suite red is a pre-existing baseline condition.

Not merging — stopping here for independent exact-head review.

---

🤖 generated by **gaebal-gajae (clawdbot)**

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -4698,6 +4698,26 @@ exit 0
       assert.equal(countMatches(hookCommand, /tmux if-shell -F/g), 2);
       assert.equal(countMatches(hookCommand, /run-shell -b/g), 1);
     });
+
+    it("escapes injected authority conditions so they survive the outer run-shell format pass", () => {
+      const args = buildRegisterResizeHookArgs("omx-detached:0", "omx_resize_detached", "%77", 2, 7700, "instance-1");
+      const hookCommand = guardDetachedHudDeferredMutation(leader, hud, args).at(-1)!;
+
+      // run-shell applies exactly one outer tmux format pass to the stored payload.
+      // Unescaped, tmux pre-expands these conditions against the hook's own pane and
+      // collapses them to constants, denying the resize before the sink is parsed.
+      for (const condition of [
+        "##{==:##{pane_id},%11}",
+        "##{==:##{pane_pid},1100}",
+        "##{==:##{@omx_instance_id},instance-1}",
+        "##{==:##{pane_id},%77}",
+        "##{m:*OMX_DETACHED_HUD_OPERATION=operation-1*,##{pane_start_command}}",
+      ]) {
+        assert.ok(hookCommand.includes(condition), `stored payload must escape ${condition}`);
+      }
+      assert.doesNotMatch(hookCommand, /(^|[^#])#\{==:#\{pane_id\},%77\}/);
+      assert.doesNotMatch(hookCommand, /(^|[^#])#\{&&:/);
+    });
   });
 
   it("registerDetachedHudLayoutReconcileHook reads TMUX from the detached leader pane before registering", () => {
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -98,6 +98,7 @@ import {
   CODEX_SQLITE_HOME_ENV,
   DETACHED_TMUX_HISTORY_LIMIT,
   isExistingTmuxWindowTooCrampedForLaunchHud,
+  guardDetachedHudDeferredMutation,
 } from "../index.js";
 import { buildResumeArgsWithPreservedFlags, stripHotswapArg } from "../../auth/hotswap.js";
 import { mergeConfig, repairConfigIfNeeded } from "../../config/generator.js";
@@ -112,6 +113,7 @@ import {
 } from "../../config/models.js";
 import type { ProcessEntry } from "../cleanup.js";
 import { splitWorkerLaunchArgs } from "../../team/model-contract.js";
+import { buildRegisterResizeHookArgs } from "../../team/tmux-session.js";
 
 
 const testDir = dirname(fileURLToPath(import.meta.url));
@@ -4654,6 +4656,48 @@ exit 0
     assert.equal(env.OMX_TMUX_HUD_OWNER, "1");
     assert.equal(env.OMX_ROOT, "/repo");
     assert.equal(env.OMX_ENTRY_PATH, "/repo/dist/cli/omx.js");
+  });
+
+  describe("detached HUD deferred mutation guard", () => {
+    const leader = {
+      paneId: "%11",
+      panePid: 1100,
+      sessionName: "omx-detached",
+      sessionId: "$1",
+      sessionCreated: "1700000000",
+      windowId: "@1",
+      windowIndex: "0",
+      ownerId: "instance-1",
+    };
+    const hud = {
+      paneId: "%77",
+      panePid: 7700,
+      sessionId: "$1",
+      windowId: "@1",
+      operationMarker: "operation-1",
+    };
+
+    it("throws fail-closed when a deferred HUD payload has format escapes but no resize sink", () => {
+      const sinkFree = ["run-shell", "-b",
+        "tmux list-panes -a -F '#{pane_id}\t#{pane_dead}\t#{pane_pid}' >/dev/null 2>&1 || true"];
+      assert.throws(
+        () => guardDetachedHudDeferredMutation(leader, hud, sinkFree),
+        /detached deferred HUD mutation lacks a recognized resize sink/,
+      );
+    });
+
+    it("guards immediate and delayed resize sinks as bare tmux command-list commands", () => {
+      const args = buildRegisterResizeHookArgs("omx-detached:0", "omx_resize_detached", "%77", 2, 7700, "instance-1");
+      const guarded = guardDetachedHudDeferredMutation(leader, hud, args);
+      const hookCommand = guarded.at(-1)!;
+
+      assert.match(hookCommand, /resize-pane -t %77 -y 2 ; display-message/);
+      assert.doesNotMatch(hookCommand, /tmux resize-pane -t %77 -y 2 ; display-message/);
+      assert.match(hookCommand, /tmux if-shell -F/);
+      assert.match(hookCommand, /##\{pane_id\}/);
+      assert.equal(countMatches(hookCommand, /tmux if-shell -F/g), 2);
+      assert.equal(countMatches(hookCommand, /run-shell -b/g), 1);
+    });
   });
 
   it("registerDetachedHudLayoutReconcileHook reads TMUX from the detached leader pane before registering", () => {
--- a/src/cli/__tests__/launch-fallback.test.ts
+++ b/src/cli/__tests__/launch-fallback.test.ts
@@ -285,6 +285,26 @@ async function assertShimLogQuiescent(path: string, quietPeriodMs: number, messa
   assert.equal(await readFile(path, 'utf-8').catch(() => ''), before, message);
 }
 
+async function waitForServerLogSubstring(
+  readServerLog: () => Promise<string>,
+  offset: number,
+  needle: string,
+  timeoutMs = 3_000,
+  intervalMs = 25,
+): Promise<string> {
+  const deadline = Date.now() + timeoutMs;
+  let lastWindow = '';
+  while (Date.now() <= deadline) {
+    const log = await readServerLog();
+    lastWindow = log.slice(offset);
+    if (lastWindow.includes(needle)) return lastWindow;
+    await new Promise((resolve) => setTimeout(resolve, intervalMs));
+  }
+  throw new Error(
+    `timed out after ${timeoutMs}ms waiting for ${JSON.stringify(needle)} in private tmux server log; tail=${JSON.stringify(lastWindow.slice(-2_000))}`,
+  );
+}
+
 async function stopHeldOmx(child: ReturnType<typeof spawn>, releasePath: string): Promise<void> {
   await rm(releasePath, { force: true });
   await new Promise<void>((resolve, reject) => {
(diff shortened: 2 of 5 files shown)


Yeachan-Heo:
## Verification evidence

Exact head: `36a7f809f49b848b3c7caa5fd9e9f7d692bdc008` · base `dev` @ `99d87f36` · **tmux 3.2a** (reporter used 3.7b; nested command-list parsing semantics are version-independent).

### Pre-fix discriminators, captured at the immutable `465502cd`

That commit contains the tests plus the one visibility export, with **product behavior unchanged**, so the failures below are observed on genuinely unfixed code:

| criterion | result at `465502cd` |
|---|---|
| AC1/AC2 real-tmux restore | ✖ `AssertionError: the first larger client attach must restore the enlarged detached HUD — 7 !== 2` |
| AC6 fail-closed throw | ✖ `AssertionError: Missing expected exception` |
| bare-sink companion | ✖ fails |
| injected-condition escaping | ✖ fails |
| AC5 owner denial | ✔ passes *(preservation check)* |
| AC3 oracle | ✔ passes **vacuously** — see note |

### Post-fix

| gate | result |
|---|---|
| `dist/cli/__tests__/index.test.js` | 327/327 |
| `dist/cli/__tests__/launch-fallback.test.js` | 35/35 — **3 consecutive runs** |
| `dist/team/__tests__/tmux-test-fixture.test.js` | 7/7 |
| focused 6-file compiled run | all green |
| `npm run lint` | clean |
| `npm run check:no-unused` | clean |
| `npm run verify:plugin-bundle` | 29 skill dirs verified |
| `npm run verify:native-agents` | 22 agents, 37 prompt assets |
| `npm pack --dry-run --json` | 3397 entries, required artifacts present, **zero** forbidden paths |

The three new real-tmux cases, timed and green each run:

```
✔ restores an enlarged detached HUD on the first larger client attach   (7090ms)
✔ records the guarded resize command without an unknown-command parser error (7130ms)
✔ denies detached HUD mutation after the pane owner changes             (7069ms)
```

They are deliberately **independent** — each owns its fixture, work directory and shim — so an early height failure cannot mask the later oracle and denial evidence.

### Adversarial suite — 22/22

**Guard surface (12):** sink-free payload bearing all three pane formats *(the actual bypass)*, sink-free without formats, bare-only sink, `-y 0`, `-y 007`, negative height, `%abc` target, `set-hook` without `run-shell -b`, both `run-shell` branch shapes, and the passthrough shape — all correctly throw or guard. Both the immediate and the delayed sink are wrapped (2 `tmux if-shell`, 1 `run-shell -b`, 2 bare inner sinks, **0** prefixed). No injected authority condition is ever triple-escaped; the original command is escaped exactly once.

**Real private tmux (10):**

```
passed | C1a/b/c repeated resize at 121x41, 122x42, 130x45   -> 2 every time
passed | C2  idempotent re-fire when already at 2            -> 2
passed | C3  owner drift denies                              -> stays 7
passed | C3b ownership restored resizes again                -> 2
passed | C4  pane_pid drift after respawn denies             -> stays 7
passed | C5  recycled pane id not mutated by stale hook      -> stays 7
passed | C6  unrelated session byte-identical
passed | C7  no injection sentinel created
```

Private server confirmed terminated after the run.

### Scope discipline

Five files. One product hunk (`@@ -1571,26 +1571,34 @@`). **Zero** diff lines touch reasoning-flag or geometry code — AC4 holds by construction, and all reasoning pins stay green. The pre-existing provider fake at `launch-fallback.test.ts:1825` is **byte-unchanged** and appears in the diff only as context. No added test line applies an execution surface to captured tmux payload text.

### Known limitations, stated rather than papered over

1. **AC3 is a post-fix guarantee, not a pre-fix discriminator.** Because defect 2 masks defect 1, `unknown command: tmux` is *unreachable* on the unfixed tree — so that case passes vacuously pre-fix. Its non-vacuity is protected by a preceding assertion requiring the server-log window to actually cover the `resize-pane` command.
2. **`npm run smoke:packed-install` fails — identically on the pristine base** `99d87f36`. This workstation has codex-cli 0.145.0 against the script's 0.142.5 boundary. Pre-existing environment limitation; reported rather than worked around.
3. **The base fails its own full `npm test`** with 19 unrelated failures (MCP server-lifecycle, native Stop-hook, UserPromptSubmit suites). Full-suite red is a pre-existing baseline condition. Two branch full-suite runs each failed exactly one *different* unrelated test, both passing in isolation — load-induced flakiness under concurrent `npm pack` and real-tmux work, not a regression.

Not merging. Stopping here for independent exact-head adversarial review.

---

🤖 generated by **gaebal-gajae (clawdbot)**

Yeachan-Heo:
## MERGE_READY

Independent exact-head red-team review. Nothing here was taken on trust from the verification comment or from the 23/23 hosted checks.

### Head binding

| check | value |
|---|---|
| local HEAD | `36a7f809f49b848b3c7caa5fd9e9f7d692bdc008` |
| `refs/pull/3296/head` | `36a7f809f49b848b3c7caa5fd9e9f7d692bdc008` |
| REST `head.sha` | `36a7f809f49b848b3c7caa5fd9e9f7d692bdc008` |
| base / `merge-base` | `99d87f36e1ab14fcfc76c54ca0ff809bb965f6cf` / **identical** |

Merge-base equals the base tip, so the diff reviewed is the real merge delta — no stale-rebase illusion. 5 files, +397/−48, exactly one product hunk.

### The three defects are real, and the fix is right

**1 — nested `tmux` inside a tmux command list.** The old `guardedResize` re-emitted the entire match, so the `if-shell` *success body* was `tmux resize-pane …`. That body is parsed as a **command list**, where `tmux` is not a command → `unknown command: tmux`. The new capture group substitutes the bare sink. The outer wrapper correctly *keeps* its `tmux ` prefix because it is emitted into `run-shell` **shell** context. That shell-vs-command-list asymmetry is the whole bug, and the patch encodes it in the right direction on both sides.

**2 — authority conditions collapsed before evaluation.** `run-shell` applies one outer format pass to the stored payload. Unescaped, `#{==:#{pane_id},%77}` is pre-expanded against the *hook's own* pane and frozen into a constant, denying every resize. `outerFormatEscaped` doubles `#{`→`##{` on injected conditions only, consistent with the pre-existing `##{pane_id}` convention already applied to the command body. I checked the escape arity directly: the condition builders emit only single `#{`, so `replaceAll` cannot triple-escape.

**3 — the fail-closed guard was aliased open.** This is the one I'd call security-relevant. `guardedCommand === command` compared the **post-escape** output against the **pre-escape** input. A sink-free payload containing `#{pane_id}` differs by escaping alone, so it passed the inequality check and the guard silently returned an **unguarded** payload into a stored hook. `guardedSinkCount === 0` is the correct discriminator. Good catch, and the right fix rather than a patch over the symptom.

### Attack surface

- **Injection into the conditions:** none reachable. Every authority field is regex-validated at capture (`paneId` `%\d+`, `panePid` safe positive int, `sessionId` `\$\d+`, `windowId` `@\d+`, `sessionCreated`/`windowIndex` digit-only), and `ownerId` / `operationMarker` are internally generated. No attacker-controlled text reaches the condition string.
- **Regex tightness:** `/tmux (resize-pane -t %[0-9]+ -y [1-9][0-9]*)/g` correctly rejects `-y 0`, `-y 007`, negatives and `%abc`.
- **Counter scope:** `guardedSinkCount` is function-local, reset per call — no cross-call leakage.
- **Call sites:** both are inside `guardDetachedHudDeferredMutation` (`set-hook` and `run-shell` branches) and consume the return purely as a payload string. Nothing depended on the old identity semantics.

### The false-green question — answered independently

I did **not** accept the pre-fix table. I checked out `465502cd` into a **detached temp worktree** and ran it there.

First, the commit itself: its only `src/cli/index.ts` change is `function` → `export function`. Product behaviour is genuinely unchanged, so failures observed there are on truly unfixed code.

Result at `465502cd`:

```
✖ throws fail-closed when a deferred HUD payload has format escapes but no resize sink
    AssertionError: Missing expected exception.
✖ guards immediate and delayed resize sinks as bare tmux command-list commands
    AssertionError: input expected NOT to match /tmux resize-pane -t %77 -y 2 ; display-message/
```

**One correction to the evidence comment.** The table lists "injected-condition escaping ✖ fails" at `465502cd`, but that test was added in the *fix* commit `6831e91f`, not in the discriminator commit — it did not exist at `465502cd` to fail. I therefore replayed its five assertions manually against the unfixed build:

```
FAIL ##{==:##{pane_id},%11}
FAIL ##{==:##{pane_pid},1100}
FAIL ##{==:##{@omx_instance_id},instance-1}
FAIL ##{==:##{pane_id},%77}
FAIL ##{m:*OMX_DETACHED_HUD_OPERATION=operation-1*,##{pane_start_command}}
unescaped #{&&: present: true      → would be RED
```

So the *claim* holds — the discriminator is genuine — but the *attribution* was off by one commit. Worth fixing in future evidence tables; it is a bookkeeping error, not a false green.

Your **AC3 vacuity disclosure is accurate and correctly reasoned**: defect 2 masks defect 1, so `unknown command: tmux` is unreachable pre-fix and that oracle passes vacuously. Guarding its non-vacuity with a preceding assertion that the server-log window actually covers the `resize-pane` command is the right mitigation. Stating this rather than burying it is the reason this review went quickly.

### Local verification at exact head

| suite | result |
|---|---|
| `dist/cli/__tests__/index.test.js` | **327/327** |
| `dist/cli/__tests__/launch-fallback.test.js` | **35/35** (all 3 real-tmux cases green) |
| `dist/team/__tests__/tmux-test-fixture.test.js` | **7/7** |
| `dist/cli/__tests__/windows-popup-loop-contract.test.js` | **4/4** |
| `npm run lint` | clean |
| `npm run check:no-unused` | clean |

The contract suite matters here specifically because it *source-greps* `cli/index.ts` for `function guardDetachedHudDeferredMutation` and for the call-site line, and the new `export` keyword touches that matched text. I executed it rather than reasoning that the regex still matches. It does. tmux 3.2a.

**On the claimed pre-existing baseline failures:** I ran `dist/mcp/__tests__/server-lifecycle.test.js` — one of the suites named as failing — at this exact head. **15/15 green.** That suite is not red here at all, which supports the load-induced-flakiness reading over a structural baseline problem, and either way it cannot be a blocker for this PR.

### Non-blocking observation

`buildNestedTmuxShellCommand` emits a **quoted absolute path** instead of a literal `tmux` when `process.platform === 'win32'` and tmux resolves absolutely. That sink shape would not match the new regex, so `guardedSinkCount` stays 0 and the guard throws. This **fails closed**, not open, and the detached-HUD guard path is gated behind `!nativeWindows` at the `buildDetachedSessionFinalizeSteps` call site — so it is unreachable today. Not a blocker. Flagging it because it is the concrete instance of a broader coupling: the guard depends on the exact textual shape emitted by `buildHudResizeCommand`, so a future change to that emitter converts a guarded path into a hard throw. Fail-closed is the right default, but a tracking issue for that coupling would be worth opening separately.

### Verdict

Three genuine defects, one of them a real fail-closed bypass. Minimal product surface — a single hunk. Discriminators independently confirmed red on unfixed code. All focused suites green at exact head. The one inaccuracy I found is an attribution error in the evidence table, not a correctness or coverage gap.

No product code was mutated, and nothing was committed, pushed, or merged from this review lane. The pre-fix replay used a detached temporary worktree; this worktree stayed at `36a7f80` throughout.

**MERGE_READY.**

—

Issue #3294, opened 2026-07-24
[Bug] Ultragoal native aggregate goal uses ambiguous cwd-relative state paths in nested projects


arikon:
## Summary

When `omx ultragoal` is started from a project subtree instead of the repository root, the native aggregate Codex goal still points to:

```text
.omx/ultragoal/goals.json
.omx/ultragoal/ledger.jsonl
```

Those paths are relative, but the goal text does not record the working directory or the selected Ultragoal state root. A later OMX/Codex step can resolve them from the repository root and read a stale or unrelated root-level Ultragoal, or fail to find the active project-local state.

This is adjacent to #3139, but distinct: #3139 covers CLI/hook authority diverging between nested `.omx` trees. This report is specifically about ambiguous paths embedded in the native aggregate Codex goal objective.

## Environment

- `oh-my-codex`: `0.20.3`
- Installed package metadata: `Yeachan-Heo/oh-my-codex`
- Codex goal mode: aggregate
- Launch directory: a project subtree below the repository root

The installed package currently defines the aggregate objective as:

```text
Complete the durable ultragoal plan in .omx/ultragoal/goals.json, including later accepted/appended stories, under the original brief constraints; use .omx/ultragoal/ledger.jsonl as the audit trail.
```

## Minimal reproduction

Assume this repository layout:

```text
/repo/
├── .omx/ultragoal/                         # old or unrelated root-level state
└── subproject/
    └── .omx/ultragoal/                     # active state for this run
        ├── goals.json
        └── ledger.jsonl
```

1. Start the workflow from the project subtree:

   ```sh
   cd /repo/subproject
   omx ultragoal ...
   ```

2. Confirm that the active durable state is:

   ```text
   /repo/subproject/.omx/ultragoal/goals.json
   /repo/subproject/.omx/ultragoal/ledger.jsonl
   ```

3. Inspect the native aggregate Codex goal. It contains only:

   ```text
   .omx/ultragoal/goals.json
   .omx/ultragoal/ledger.jsonl
   ```

4. Later, resolve those references from the repository root `/repo`, as can happen after a session/root handoff or when an agent treats the repo root as the path base.

5. The same goal text now selects:

   ```text
   /repo/.omx/ultragoal/goals.json
   /repo/.omx/ultragoal/ledger.jsonl
   ```

   instead of the active state under `/repo/subproject`.

## Expected behavior

[...]

Issue #3293, opened 2026-07-24
bug(hooks): deep-interview self-lock recurs with session-local OMX shim and undocumented handoff scope


engels74:
## Summary

The deep-interview PreToolUse self-lock reported in #3239 is partially recurring on current `dev`.

PR #3240 successfully fixed the attached-tmux `omx question` path: structured questions now work. However, after completing the interview, Autopilot can still become permanently trapped in `deep-interview` because:

1. the documented deep-interview → Ralplan state handoff is rejected;
2. `omx cancel` is rejected because the real attached-tmux session resolves `omx` through OMX's session-local runtime shim rather than directly to `dist/cli/omx.js`;
3. `omx cancel --force` is also rejected inside the active deep-interview;
4. running `omx cancel --force` outside Codex reaches the CLI but fails on contradictory session-owner metadata;
5. the active guard then blocks Ralplan artifacts, delegated planner writes, and several recovery or diagnostic commands.

This recreates the core #3239 failure mode: the workflow cannot advance and cannot use its documented cancellation path.

Related:

- #3239
- PR #3240
- #3263 / PR #3264
- #3280 / PR #3290

The external-terminal owner-metadata failure overlaps #3280. However, #3280 and PR #3290 do not appear to cover the session-local runtime shim mismatch or the undocumented deep-interview handoff scope described below.

## Environment

- oh-my-codex: `0.20.3`
- Local source branch: `dev`
- Local source HEAD: `99d87f36e1ab14fcfc76c54ca0ff809bb965f6cf`
- Git describe: `v0.20.3-73-g99d87f36`
- Node.js: `v24.14.0`
- OS: macOS `26.5.2`, arm64
- Surface: attached tmux
- Workflow: `$oh-my-codex:autopilot`
- Session: `omx-1784879150409-cnbkxt`
- Active phase when reproduced: `deep-interview`

The installed `dist/scripts/codex-native-hook.js` contains the same relevant cancellation and handoff logic as the checked-out TypeScript source.

`codex --version` could not be collected from the affected session because that read-only version probe was itself rejected by the active deep-interview boundary.

### 1. Start Autopilot in attached tmux

Launch Codex through OMX in attached tmux. This creates and prepends the session-local OMX command shim.

Activate Autopilot:

```text
$oh-my-codex:autopilot <task>
```

Autopilot initializes with:

```json
{
  "active": true,
  "mode": "autopilot",
  "current_phase": "deep-interview"
}
```

[...]

PR #3290, opened 2026-07-24
fix(cancel): reconcile resumed session ownership

Yeachan-Heo:
## Summary

- keep exact canonical `omx cancel [--force]` available when only inherited `NODE_EXTRA_CA_CERTS` is present
- reconcile one stale top-level resumed skill owner inside the existing exact-session cancellation transaction
- preserve nested/live/ambiguous/cross-session denials, Team+Ralph peer state, mode-only behavior, path/symlink safety, and truthful rollback diagnostics
- add packed outside-tmux recovery coverage and preserve the merged #3285 native cache/release contract

Fixes #3280.

## Verification

- `npm test` — all 400 compiled test files passed
- focused native-hook/session-owner/cancellation regressions passed
- `npm run lint`
- `npm run check:no-unused`
- `npm run smoke:packed-install`
- `npm run sync:plugin:check`
- `npm run verify:plugin-bundle`
- #3285 native asset/release preservation suites passed
- Node 20/22/24 focused lanes passed before exact-current-dev reconciliation; reconciled full gates passed on `99d87f36e1ab14fcfc76c54ca0ff809bb965f6cf`

## Scope

This is the narrow #3280 cancellation deadlock fix. It does not widen #3272 continuation transports, infer root SessionStart authority, add generic repair APIs, or merge the PR.

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

Issue #3272, opened 2026-07-23
[Bug] Ultragoal recovery self-blocks after stale session pointer


bdrum:
## Summary

  Ultragoal/Conductor can enter a self-blocking recovery state after a stale `.omx/state/session.json` pointer.
  Durable Ultragoal state remains valid (`G001 in_progress`), but the writable state scope initially fails
  because `session.json` points to a dead process. After restoring Ultragoal state with explicit `session_id`,
  the active Main-root Conductor guard blocks the very transports required to continue: `create_goal`, native
  subagent spawn, `omx` state/ledger commands, and even read-only Bash commands such as `find`.

  This looks related in class to #3270: fail-closed workflow gates are correct, but the current surface reaches a
  state where no documented safe recovery/continuation path remains.

  ## Environment

  - OS: Linux x64
  - Node.js version: v22.15.0
  - `oh-my-codex` version: v0.20.3 observed in plugin/runtime paths
  - Codex CLI version: 0.145.0 observed in affected local runs

  ## Steps to reproduce

  1. Start or resume an Ultragoal run so durable `.omx/ultragoal/goals.json` has an active goal, e.g. `G001
  status=in_progress`.
  2. Let the original owner process/session die so `.omx/state/session.json` still exists but points to a stale/
  dead PID.
  3. Attempt to restore/continue Ultragoal state:

     ```sh
     omx state write --input '{"mode":"ultragoal","active":true,"current_phase":"reviewing"}' --json

  4. Observe the unusable session pointer error.
  5. Retry with explicit session_id, which succeeds:

     omx state write --input '{"session_id":"omx-1784800905160-
     lcv7m3","mode":"ultragoal","active":true,"current_phase":"reviewing"}' --json

  6. Continue the required Ultragoal handoff/review flow:
      - call create_goal when Codex goal is absent;
      - spawn/reuse a delegated subagent;
      - run read-only verification commands;
      - write state/ledger metadata through omx.

  ## Expected behavior

  After explicit session recovery succeeds, Conductor should allow the documented safe continuation transports
  needed by Ultragoal:

  - create_goal / get_goal / update_goal when the Ultragoal handoff requires Codex goal reconciliation;
  - native subagent spawn/reuse transports, including flattened Codex hook names if Codex emits them;
  - bounded read-only Bash commands;
  - validated omx state / Ultragoal ledger metadata writes bound to the recovered session.

[...]

PR #3298, opened 2026-07-25
fix(state): recover exact stale session binding and revalidate durable state commits (#3272)

Yeachan-Heo:
## Summary

Fixes the narrowed reproduction in #3272: a stale-dead selected `.omx/state/session.json` pointer defeated an exact current `OMX_SESSION_ID`, so the identical state write only succeeded after manually replacing the pointer.

`resolveWritableStateScope` now binds the environment session as the writable scope when — and only when — one authoritative pointer snapshot is classified `stale-dead`, that snapshot holds selected-root authority (recorded cwd containment plus canonical `state_root` equality), the pointer bytes are stable across a re-read, and `OMX_SESSION_ID` passes `normalizeSessionId`. Scope resolution never writes `session.json`; SessionStart remains the only pointer writer.

## What changed

- **Stale-dead reconciliation** (`src/mcp/state-paths.ts`): one post-selection decision snapshot plus one raw byte-stability re-read drive both the authority and liveness decisions. Malformed, foreign-cwd, identity-indeterminate, and stale-dead-without-binding pointers all remain fail-closed on `unusableSession`. The authority-snapshot catch was narrowed so only a JSON `SyntaxError` yields `null` and filesystem errors propagate; the stability re-read now swallows only `ENOENT`.
- **Fail-closed Ultragoal ordering** (`src/ultragoal/artifacts.ts`): no durable story/goal transition before writable lifecycle authority resolves.
- **W4** (`src/state/operations.ts`): `completeRalplanSession` derives every target from the pinned `baseStateDir` and the caller's `capturedScope` with no internal re-resolution, revalidates before each write and the unlink, and requires `capturedScope` whenever `beforeCommit` is supplied. A failed root-skill unlink no longer reports success.
- **W5** (`src/ultragoal/artifacts.ts`): `assertUltragoalWritableLifecycleAuthority` returns a discriminated `resolved` / `no-pointer-compat` authority token; `withUltragoalMutationLock` requires exact pre-lock/post-lock equality and fails closed on any drift, including both `resolved`↔`no-pointer-compat` directions. The `unusableSession` message text and the single documented `unboundEnvironment` compatibility exception are unchanged.
- **Write-site revalidation**: a typed `WritableCommitSite` inventory plus `createWritableCommitRevalidator`, whose one-based `commitOrdinal` is closure-local to a single top-level operation. This covers the previously unguarded pre-assert workflow-transition writes reached through `completeSourceModeState`, and the canonical session skill-state unlink in `syncCanonicalSkillStateForMode`.
- **Pinned run-state** (`src/runtime/run-state.ts`): accepts a pinned target path and skips scope re-resolution when supplied; behavior is unchanged without it.
- **Docs** (`docs/troubleshooting.md`): corrected stale-pointer recovery guidance and a plainly worded known-limitation note.

## Scope of the guarantee

Scope revalidation is a **point-in-time check** performed immediately before each commit. It reduces but does **not** eliminate the window between validation and the filesystem write, and a completion touching several files is **not** atomic — a concurrent SessionStart publication can still interleave. Nothing here claims TOCTOU closure, serialization, continuous authority, or atomic multi-file completion.

Serializing writable commits with SessionStart pointer publication is deliberately deferred to a named follow-up lane, `state: serialize writable commits with SessionStart pointer publication`, which also owns `neutralizeOwnedRoutingRalplan` and the remaining uninstrumented pre-existing durable writers (notably the Ralph artifact writes in `src/ralph/persistence.ts`, untouched here).

## Known coverage gap (disclosed, not hidden)

The five `completeRalplanSession` commit sites are not test-reachable: `buildRalplanConsensusGateFromSources` (`src/ralplan/consensus-gate.ts`, untouched by this PR) returns `complete: false` with `documented_host_consensus_receipt_unavailable` for every input, so `validateRalplanTerminalConsensus` throws before the first guarded commit. This was verified with a direct probe that observed zero site events, and independently re-confirmed with forged approving review data. Making that gate injectable or fabricating a host receipt would manufacture authorization evidence and is out of scope.

## Verification

- `bun run build` — pass
- `bun run lint` — pass (783 files, no findings)
- `bun x tsc -p tsconfig.no-unused.json` — pass
- `npm run test:node` — pass (400 test files)
- Focused compiled suites: `state-paths` 60/60, `operations` 104/104, `ultragoal/artifacts` 61/61, `skill-active` 12/12

Adversarial coverage includes the authority-bypass matrix (sibling `state_root`, parent/child cwd, symlink canonicalization), pointer immutability by bytes and mtime, byte-instability races against live/malformed/deleted replacements, closure-local ordinal integrity under concurrency, exact-pair takeover drift detection that fails loudly if a write is inserted or removed, W4 guard enforcement, and W5 token drift across all five transitions with `goals.json` and `ledger.jsonl` proven byte-identical after every rejection.

This branch is rebased onto current `origin/dev` (`f9536389`, terminal green CI) with zero file overlap and no conflicts; it is 8 ahead / 0 behind. A small number of process-spawn tests (`server-lifecycle`, `detached-leader-teardown`, `codex-plugin-layout`) are order- and environment-sensitive on the development host and also fail intermittently on untouched `origin/dev`; one such failure was traced to gitignored `.omx/` runtime state being copied into a test fixture, not to this change. Analysis is recorded in the lane evidence.

## Out of scope

#3293, #3290, #3294, #3270; broad native-assignment / routing / owner-reconciliation / Conductor redesign; cancellation/resume; CLI cancel; native-hook transport; `all_sessions` changes; the SessionStart publication-lock transaction. The patches shared in the issue thread were treated as advisory evidence only and were not ported.

Refs #3272

Mutation verification: neutralizing the non-object pointer guard drives the state-paths suite red, and neutralizing the `skill-active.session-unlink` revalidation drives the operations suite red; both files were then restored byte-exactly (`sha256sum -c` clean) and both suites re-run green.


Yeachan-Heo:
## MERGE_READY

Independent exact-head red-team review, review-only lane. No product mutation, no commit/push, no merge from this lane.

### Head / base binding

| Check | Value |
|---|---|
| local worktree HEAD | `ea02cdf14060a28cac0087aaba654b456405283f` |
| REST `pulls/3298.head.sha` | `ea02cdf14060a28cac0087aaba654b456405283f` |
| `ls-remote refs/pull/3298/head` | `ea02cdf14060a28cac0087aaba654b456405283f` |
| `origin/dev` | `f95363894a45a929f239be64c703b9354a2df21e` |
| `merge-base(origin/dev, HEAD)` | `f9536389` — equal to `origin/dev`, zero base drift |
| diff vs current dev | 17 files, +1532 / -112 |

All three head sources agree; the branch is strictly ahead of current dev with no rebase debt.

### Attack results on the four named surfaces

**1. Stale session pointer recovery — holds.**
`resolveWritableStateScope` (`src/mcp/state-paths.ts:625-709`) takes exactly one authoritative snapshot and derives both authority and liveness from those same immutable bytes, so authority and liveness cannot be evaluated against different pointer generations. Recovery is conjunctive and narrow: normalized `OMX_SESSION_ID`, a snapshot that passed the selected-root authority test (nonempty recorded cwd, canonical `state_root == baseStateDir`, observed cwd contained by recorded cwd), `classifySessionStateLiveness === 'stale-dead'`, a normalizable `session_id` in the pointer, and byte-identical re-read. Attempted bypasses that all correctly fail closed: `identity-indeterminate` liveness (probe throw, unreadable `/proc`, cmdline-hash mismatch, non-integer `pid_start_ticks`) never reaches recovery; sibling `state_root`; parent/child cwd escape; recovery is unreachable when the pointer is `usable` since `metadata` is then truthy. Recovery is read-only — no branch writes `session.json`, so SessionStart remains the sole pointer writer. Second-order check: `discoverSessionAuthorityBaseStateDir` still gates ancestor selection on `isSessionStateUsable`, so a stale-dead pointer cannot promote a foreign ancestor root into `baseStateDir` and then be "recovered" into it.

**2. Durable writable-scope revalidation — sound, and correctly scoped in its claim.**
`createWritableCommitRevalidator` closes over the pinned `capturedScope` and `baseStateDir`; `assertWritableStateScopeUnchanged` re-resolves and compares all four of `source`, `sessionId`, `stateDir`, and the recomputed base dir. The important structural fix is that call sites no longer recompute their target after authorization: `startMode`, `updateModeStateInternal`, `state_write`, and `state_clear` now build paths as `join(scope.stateDir, getStateFilename(mode))` instead of round-tripping through `getStatePath(...)`, and `syncRunStateFromModeState` accepts a pinned `targetPath` that skips re-resolution entirely. That closes the real bug class — an authorized write landing in a re-resolved directory — rather than only detecting it. `commitOrdinal` is closure-local per top-level operation, so concurrent operations cannot interleave ordinals. The in-code doc comment states plainly that this is a point-in-time check that narrows but does not eliminate the write window and that multi-file completion is not atomic; the PR body repeats this. That is the correct claim, and I did not find it overstated anywhere in the diff.

**3. Skill-state unlink revalidation — holds, and fixes a real silent-failure.**
`syncCanonicalSkillStateForMode` (`src/state/skill-active.ts:571-579`) now revalidates before `unlink` and the previous `catch(() => {})` is narrowed to swallow only `ENOENT`; every other errno propagates. Same treatment on the `ralplan.root-skill-unlink` site in `completeRalplanSession`, where a failed unlink previously still returned success. The `state-clear.primary` unlink is likewise guarded. The `beforeCommit` is threaded, not reconstructed, through `writeSkillActiveStateCopiesForStateDir` → `writeSkillActiveStateCopiesToPaths`, so the pinned scope is preserved across the whole sync.

**4. Fail-closed non-object pointer handling — holds.**
`readAuthoritativeSessionSnapshotFromBaseStateDir` rejects `null`, arrays, and scalars before any property access, so a JSON-valid non-object pointer resolves to the stable `unusableSession` error instead of a `TypeError`. Error narrowing is right in both directions: `ENOENT` → `null`, `SyntaxError` → `null`, everything else propagates rather than being laundered into a generic "unusable" classification. I specifically checked that this does not accidentally widen the recovery path — a non-object pointer yields `snapshot === null`, which makes the `snapshot &&` guard in the recovery branch false, so it still throws.

### Cross-checks against the PR's own claims (not taken on trust)

- **Disclosed `completeRalplanSession` coverage gap — verified truthful.** `buildRalplanConsensusGateFromSources` (`src/ralplan/consensus-gate.ts:76-101`) unconditionally returns `complete: false` with `documented_host_consensus_receipt_unavailable` for every input, so `validateRalplanTerminalConsensus` throws before the first guarded commit. Confirmed that file is untouched by this PR (`git diff origin/dev...HEAD -- src/ralplan/consensus-gate.ts` is empty). The gap is real, correctly attributed, and disclosing it beats fabricating a host receipt.
- **Error-message rename is complete.** `unboundEnvironment` → `sessionBindingMismatch` on the live-pointer mismatch path only; the `unboundEnvironment` string itself is retained for the genuinely-unbound case and is still the single documented compatibility exception in `assertUltragoalWritableLifecycleAuthority`. `src/scripts/codex-native-hook.ts` accepts both, and the three assertion-only test updates match.
- **`readUltragoalPlanSnapshot` split is behaviour-correct.** `omx ultragoal status` is the only consumer changed, and it is exactly the read-only surface that should not perform a durable legacy migration. The migration moved under `withUltragoalMutationLock` with a re-read (`readUltragoalPlanUnderLock`), so a concurrent mutator cannot be clobbered.
- **Codex bot review is not usable evidence** — it only reported usage limits. Nothing in this verdict rests on it.

### Local verification I ran (smallest sufficient set)

- `npx tsc -p tsconfig.no-unused.json` — clean
- `npm run build` — clean
- `state-paths` **60/60**, `operations` **104/104**, `ultragoal/artifacts` **61/61**
- Independent mutation probes against build output (restored byte-exact, `sha256sum -c` clean, worktree clean):
  - neutralize the non-object pointer guard → `state-paths` red with `Cannot read properties of null (reading 'cwd')` — the guard's own failure mode, exactly as designed
  - neutralize the byte-stability re-read comparison → **T11** red
  - neutralize the `skill-active.session-unlink` revalidation → `operations` red
  These confirm the new tests are load-bearing rather than vacuous. Hosted CI is green 23/23 on this exact head.

### Non-blocking notes (do not gate merge)

1. `assertWritableStateScopeUnchanged` uses `writableScopeRevalidationHookForTests(event!)` with a non-null assertion while `event` is declared optional. Test-hook path only, but the signature and the assertion disagree; making `event` required at the internal call site would remove the lie.
2. The `WritableCommitSite` union is a hand-maintained inventory. Nothing structurally prevents a future durable write from being added without a site entry and therefore without revalidation. `src/ralph/persistence.ts` is already a known uninstrumented writer and is correctly deferred to the named follow-up lane — worth keeping that lane visible so the inventory does not silently rot.
3. Four test-only hook setters now live in production `state-paths.ts`. Consistent with existing `ForTests` precedent in this repo, so not a new sin.

### Verdict

**MERGE_READY.** The change is fail-closed at every branch I attacked, pins write targets instead of re-deriving them post-authorization, narrows two error-swallowing catches that previously hid real failures, and its limitations section is accurate rather than promotional — it claims a point-in-time check and delivers exactly that, with no TOCTOU-closure or atomicity claim to disprove. The one genuine coverage gap is disclosed with a correct root cause I independently confirmed.

Merge authority is not exercised from this lane.

—

Issue #3327, opened 2026-07-27
test: canonicalize macOS /var temporary fixture roots across suites


arikon:
## Summary

On macOS, eight test files fail when the inherited `TMPDIR` is the standard `/var/folders/...` spelling, but all 33 failures pass when the same tests run with `TMPDIR=/private/tmp`.

This is test-fixture portability drift, not a request to weaken production path guards. macOS exposes `/var` as a symlink to `/private/var`; the affected tests retain lexical `mkdtemp(tmpdir())` paths while production correctly returns canonical paths or rejects symlink-based authority roots.

## Reproduction

From `dev` at `7f8b17d0111cfe5794567cea3c806bdbfb21f955`:

```sh
npm run build
node dist/scripts/run-test-files.js \
  dist/cli/__tests__/auth.test.js \
  dist/cli/__tests__/mcp-parity.test.js \
  dist/cli/__tests__/session-scoped-runtime.test.js \
  dist/cli/__tests__/uninstall.test.js \
  dist/cli/__tests__/update.test.js \
  dist/hooks/extensibility/__tests__/dispatcher.test.js \
  dist/team/__tests__/scaling.test.js \
  dist/ultragoal/__tests__/artifacts.test.js
```

Default macOS `TMPDIR` results:

- `auth`: 9/10, 1 failure
- `mcp-parity`: 7/8, 1 failure
- `session-scoped-runtime`: 39/41, 2 failures
- `uninstall`: 57/79, 22 failures
- `update`: 76/79, 3 failures
- `dispatcher`: 15/16, 1 failure
- `scaling`: 117/119, 2 failures
- `ultragoal/artifacts`: 79/80, 1 failure

Total: 33 failures.

With `TMPDIR=/private/tmp`, all eight files pass: 10/10, 8/8, 41/41, 79/79, 79/79, 16/16, 119/119, and 80/80.

## Root cause

The tests construct security-sensitive roots and exact path expectations from the lexical result of `mkdtemp(join(tmpdir(), ...))`. On macOS that is commonly `/var/folders/...`, while `realpath()` and process cwd resolution produce `/private/var/folders/...`.

Examples include:

- exact expected CLI/session paths built from the non-canonical fixture root;
- state pointers containing a lexical `state_root` that production correctly rejects as symlinked authority;
- allowlist roots passed through `OMX_MCP_WORKDIR_ROOTS` that production correctly rejects as symlink roots;
[...]

PR #3326, opened 2026-07-27
fail closed on unauthoritative plan bootstrap (#3309)

achieve0410:
## Summary

- reject new Ultragoal plan bootstrap when an explicit `OMX_SESSION_ID` is malformed or is not bound to authoritative `session.json`
- leave no `.omx/ultragoal` directory or durable plan artifacts after a fresh-bootstrap rejection
- preserve every existing plan artifact byte-for-byte when an unbound `--force` recreation is rejected
- preserve root-mode bootstrap and the existing-plan unbound compatibility path

## Scope

This is a fresh reconstruction of the same narrow bootstrap-authority slice from current `dev`, as requested when #3315 was closed:
https://github.com/Yeachan-Heo/oh-my-codex/pull/3315#issuecomment-5095600155

Base SHA: `7f8b17d0111cfe5794567cea3c806bdbfb21f955`

The patch is limited to `createUltragoalPlan()` authority enforcement and its focused regression tests. Existing-plan mutators retain their documented unbound compatibility behavior; this does not broaden lifecycle policy or attempt terminal-state convergence.

The stable patch ID matches the final focused patch from #3315: `faa9210cc36fd72838ec8f7db986c7dec1291e19`.

## Test intent

The tests lock these contracts:

1. explicit unbound or malformed session -> bootstrap rejects before publication
2. rejection is atomic -> no new `.omx/ultragoal` directory or durable artifacts
3. `--force` does not bypass authority -> existing artifacts remain byte-for-byte unchanged
4. valid authoritative session binding succeeds regardless of state-directory scope
5. root-mode bootstrap and existing-plan unbound mutation compatibility remain unchanged

## Verification

- 173/173 related tests pass
  - CLI Ultragoal
  - Ultragoal artifacts
  - state paths
- build passes
- lint passes across 788 files
- TypeScript `--noEmit` passes
- no-unused check passes
- `git diff --check` passes

## Local full-suite note

The complete 405-file local run finished with four unrelated failing files:

- `launch-fallback`: four macOS failures because BSD `script` rejects GNU-only `script -c`
- `uninstall`: two Windows-path durability fixtures fail under local POSIX path semantics
- `codex-native-hook`: one unchanged Conductor/PATH assertion is affected by the surrounding active OMX App workspace
- `smoke-packed-install`: one timing-sensitive assertion reaches the 5s global deadline before the PATH-candidate budget

No local full-suite green claim is made.

## Hosted exact-head CI

[...]

Issue #3324, opened 2026-07-27
bug(ultragoal): exact OMX_SESSION_ID is ignored for identity-indeterminate pointers


arikon:
## Summary

`omx ultragoal` durable mutations reject an authoritative Codex App session when the selected `session.json` PID probe is `identity-indeterminate`, even if `OMX_SESSION_ID` exactly matches `session.json.session_id`.

The emitted recovery instruction is therefore ineffective:

```text
Cannot resolve writable state scope: session.json is present but unusable. Bind the exact current session (set OMX_SESSION_ID to the current session id)
```

## Reproduction

On macOS in Codex App, with an authoritative pointer whose PID cannot be probed (`process.kill(pid, 0)` returns `EPERM`):

```bash
OMX_SESSION_ID=019fa21d-204a-71a2-a897-25cede3d9823 \
  omx ultragoal complete-goals --retry-failed --json
```

Actual result:

```text
[ultragoal] Refusing a durable ultragoal mutation before writable lifecycle authority is restored: Cannot resolve writable state scope: session.json is present but unusable. Bind the exact current session (set OMX_SESSION_ID to the current session id); see docs/troubleshooting.md (stale session pointer recovery).
```

The same exact session works through the explicit `session_id` input supported by `omx state write`.

## Root cause

`resolveWritableStateScope()` only accepts `OMX_SESSION_ID` in the non-usable-pointer branch when liveness is `stale-dead` (`src/mcp/state-paths.ts:646-680`). An `identity-indeterminate` pointer discards usable metadata and unconditionally throws, without checking whether the environment binding exactly equals the authoritative pointer's canonical `session_id`.

`assertUltragoalWritableLifecycleAuthority()` delegates to that resolver and then tells the user to set the already-set exact ID (`src/ultragoal/artifacts.ts:871-884`).

## Expected behavior

When all of the following are true, writable scope should resolve to the selected session directory:

- `session.json` has full cwd/state-root authority;
- liveness is `identity-indeterminate` rather than definitely stale;
- `OMX_SESSION_ID` exactly equals `session.json.session_id`;
- the pointer bytes are stable across the existing recovery reread.

Mismatched environment IDs, malformed/foreign pointers, and pointer replacement races must remain fail-closed.

## Proposed minimal fix

[...]

PR #3318, opened 2026-07-27
fix(hooks): stop misclassifying omx/gjc read-only discovery as writes (#3313, #3314)

Yeachan-Heo:
## Summary

Both issues share one root cause in `src/scripts/codex-native-hook.ts`: `commandHasDeepInterviewWriteIntent()` treats every recognized `omx`/`gjc` shell invocation shape as write intent via the generic `extractConductorBashMutations()` scan, without consulting the existing `--help`/`-h`/`--version`/`read`/`status` allowlist that `omxCliInvocationHasMutationIntent()` already implements correctly. `isAllowedDeepInterviewBashWrite`/`isAllowedRalplanBashWrite` never routed enough command shapes through that allowlist, so several genuinely read-only or documented-lifecycle commands were denied.

Fixes (all in `src/scripts/codex-native-hook.ts`):

- **`omx state read --mode <mode> --json` was denied.** `isAllowedOmxReadOnlyCommand()`'s `state read|status` branch only tolerated a bare `--json`. Added `stateReadTrailingArgsAreSafe()` to accept `--mode <value>` / `--mode=<value>` alongside `--json`.
- **`omx cancel` was unconditionally denied in standalone (non-Autopilot) deep-interview.** `isAllowedDeepInterviewCommandSpecificBash()` now routes it through the same `directOmxCancelCommandHasTrustedExecutionContext()` trust check ralplan already uses, instead of `return false` unconditionally. An unproven execution context (PATH smuggling, untrusted shim, etc.) still denies.
- **`omx sparkshell -- <read-only argv>` had no recognition path.** Sparkshell's direct-argv dispatch mode (`omx sparkshell [--json] [--budget N] -- <argv>`) executes the wrapped argv without shell reinterpretation, so each wrapped token is already literal. Added `extractOmxSparkshellDirectArgvCommand()` to re-quote the wrapped literal argv into a synthetic command string and re-run it through the *same* write-intent classifiers used for a directly-typed command (`isAllowedOmxSparkshellWrappedReadOnlyCommand()`), instead of trusting a bespoke allowlist. `--shell` / `--tmux-pane` / any unrecognized sparkshell flag is refused (not unwrapped), so only bounded literal-argv dispatch is reclassified — a wrapped mutation command (e.g. `omx sparkshell -- bash -c 'echo x > f'`) still denies.
[...]

PR #3331, opened 2026-07-28
test: portable PTY script(1) argv helper for Darwin/Linux (#3328)

### LINKED ISSUE #3328: test: make PTY script(1) fixture portable to macOS BSD
## Summary

Four `launch-fallback` tests fail on macOS because their PTY fixture hardcodes the GNU/util-linux `script(1)` argv shape:

```text
script -q -e -c <command> /dev/null
```

macOS ships BSD `script(1)`, where `-c` is not an option and the command follows the output file.

## Reproduction

```console
npm run build
node --test --test-name-pattern='restores an enlarged detached HUD|records the guarded resize command|denies detached HUD mutation|returns the attached client' dist/cli/__tests__/launch-fallback.test.js
```

Result on macOS: 0 passed, 4 failed. Each failure contains:

```text
script: illegal option -- c
usage: script [-aeFkpqr] [-t time] [file [command ...]]
```

## Root cause

The GNU argv is duplicated in:

- `src/team/__tests__/tmux-test-fixture.ts`
- `src/cli/__tests__/launch-fallback.test.ts`

This is a test-fixture portability bug. Production launch behavior is not involved and must remain unchanged.

## Proposed fix

Use one test-only platform-aware argv helper:

- macOS BSD: `script -q /dev/null /bin/sh -c <command>`
- GNU/Linux: `script -q -e -c <command> /dev/null`

The BSD form preserves the child exit status on macOS, including nonzero exits. Add explicit Darwin and Linux argv contract tests, then use the helper at both PTY call sites.

## Acceptance checks

- all four affected macOS tests pass
- the complete `launch-fallback` file passes
- Darwin and Linux argv contracts pass
- build, lint, and no-unused checks pass

Yeachan-Heo:
## Summary
Closes #3328.

Consolidates the two duplicated GNU/util-linux `script(1)` argv call sites
in `src/team/__tests__/tmux-test-fixture.ts` and
`src/cli/__tests__/launch-fallback.test.ts` into one test-only,
platform-aware `buildPtyScriptCommand()` helper exported from
`tmux-test-fixture.ts` (already the shared import edge between both files).

- **Linux/util-linux contract:** `script -q -e -c <command> /dev/null`
- **Darwin/BSD contract:** `script -q /dev/null /bin/sh -c <command>`

Verified against Apple's `shell_cmds` `script.1` source
(`apple-oss-distributions/shell_cmds`) rather than assuming: BSD `script`
has no `-c` flag (the command is a trailing positional argv after the log
file), and `-e` is documented as "Accepted for compatibility with
util-linux script. The child command exit status is always the exit
status of script." — i.e. BSD `script` unconditionally propagates the
child's exit status even without `-e`, so `-e` is correctly omitted from
the Darwin argv (not simply "unavailable").

The wrapped command stays exactly one argv element passed to
`/bin/sh -c` on both platforms — no shell string concatenation is
introduced, preserving existing argv tokenization/injection safety.

Also adds `isRealScriptAvailable()` (mirroring the existing
`isRealTmuxAvailable()`), and a `skipUnlessScriptAvailable()` test guard
(mirroring `skipUnlessPrivateRealTmux()`), so the PTY e2e test in
`launch-fallback.test.ts` skips deterministically when `script(1)` is
missing/unusable, and fails loudly only when `CI` is set (same pattern
already used for tmux availability).

**Test-only scope — no production launch behavior is changed.**

## Overlap check
- #3318 touches `src/scripts/**` only — no overlap.
- #3326 touches `src/cli/__tests__/ultragoal.test.ts` and
  `src/ultragoal/**` only — no overlap.
- #3324 and #3327 do not exist as PRs in this repository.

## Testing
- `node --test dist/team/__tests__/tmux-test-fixture.test.js` — 12/12 pass,
  including 5 new `buildPtyScriptCommand` Darwin/Linux contract tests
  (host-independent, exercised via explicit `platform` argument).
- `node --test --test-name-pattern="returns the attached client to the shell" dist/cli/__tests__/launch-fallback.test.js`
  — the 4 focused regression assertions (zero-status return + session
  cleanup, nonzero-status return + session cleanup) pass.
- `node --test dist/cli/__tests__/launch-fallback.test.js` — full file,
  35/35 pass.
- `npm run build` — clean.
- `npm run check:no-unused` — clean.
- `npx biome lint` on changed files — clean (`OK`).
- `git diff --stat` — 3 files changed, 99 insertions(+), 6 deletions(-),
  test files only.

### DIFF
--- a/src/cli/__tests__/launch-fallback.test.ts
+++ b/src/cli/__tests__/launch-fallback.test.ts
@@ -15,7 +15,7 @@ import {
   writeSessionStart,
   type LaunchSessionBinding,
 } from '../../hooks/session.js';
-import { isRealTmuxAvailable, withTempTmuxSession } from '../../team/__tests__/tmux-test-fixture.js';
+import { buildPtyScriptCommand, isRealScriptAvailable, isRealTmuxAvailable, withTempTmuxSession } from '../../team/__tests__/tmux-test-fixture.js';
 
 const CLI_SPAWN_TIMEOUT_MS = 60_000;
 function buildRunOmxEnv(envOverrides: Record<string, string>): NodeJS.ProcessEnv {
@@ -354,6 +354,13 @@ function skipUnlessPrivateRealTmux(t: TestContext): boolean {
   if (isRealTmuxAvailable()) return true;
   assert.equal(process.env.CI, undefined, 'CI must provide tmux for the private-server detached launch regression');
   t.skip('tmux is not installed');
+  return false;
+}
+function skipUnlessScriptAvailable(t: TestContext): boolean {
+  if (process.platform === 'win32') { t.skip('PTY launch harness requires posix script(1)'); return false; }
+  if (isRealScriptAvailable()) return true;
+  assert.equal(process.env.CI, undefined, 'CI must provide script(1) for the PTY launch regression');
+  t.skip('script(1) is not installed');
   return false;
 }
 
@@ -2685,7 +2692,7 @@ exit 0
 
   it('returns the attached client to the shell with the child status after a normal detached child exit', async (t) => {
     if (!skipUnlessPrivateRealTmux(t)) return;
-    if (process.platform === 'win32') { t.skip('PTY launch harness requires posix script(1)'); return; }
+    if (!skipUnlessScriptAvailable(t)) return;
     const wd = await mkdtemp(join(tmpdir(), 'omx-launch-detached-e2e-'));
     try {
       await withTempTmuxSession(async (fixture) => {
@@ -2712,9 +2719,12 @@ exit 0
             'OMX_HOOK_DERIVED_SIGNALS=0',
             'TERM=xterm-256color',
           ].map((kv) => `export ${kv}`).join('; ');
+          const ptyCommand = buildPtyScriptCommand(
+            `${envPrefix}; cd ${JSON.stringify(wd)} && exec ${JSON.stringify(process.execPath)} ${JSON.stringify(omxBin)} --tmux ${JSON.stringify('e2e prompt')}`,
+          );
           const result = spawnSync(
-            'script',
-            ['-q', '-e', '-c', `${envPrefix}; cd ${JSON.stringify(wd)} && exec ${JSON.stringify(process.execPath)} ${JSON.stringify(omxBin)} --tmux ${JSON.stringify('e2e prompt')}`, '/dev/null'],
+            ptyCommand.executable,
+            ptyCommand.args,
             {
               encoding: 'utf-8',
               timeout: 120_000,
--- a/src/team/__tests__/tmux-test-fixture.test.ts
+++ b/src/team/__tests__/tmux-test-fixture.test.ts
@@ -1,7 +1,10 @@
 import { execFileSync } from 'node:child_process';
 import assert from 'node:assert/strict';
+import { mkdtemp, rm, symlink } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
 import { describe, it, type TestContext } from 'node:test';
-import { isRealTmuxAvailable, tmuxSessionExists, withTempTmuxSession } from './tmux-test-fixture.js';
+import { buildPtyScriptCommand, isRealTmuxAvailable, tmuxSessionExists, withTempTmuxSession } from './tmux-test-fixture.js';
 
 function skipUnlessTmux(t: TestContext): void {
   if (!isRealTmuxAvailable()) {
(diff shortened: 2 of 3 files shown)


Yeachan-Heo:
## MERGE_READY

**Exact head reviewed:** `2f58ce61cfa9fef85acf92d6fe9403f05fd63f23`
**Current dev at review time:** `8dc95eaa3c939018e3283e195ea0bdb22321e90b` (advanced past `7f8b17d0` via #3326; re-evaluated, no impact)

### Verdict rationale

Correctly scoped test-fixture portability fix. Production launch code is untouched; the change is confined to the two PTY call sites named in #3328 (`src/team/__tests__/tmux-test-fixture.ts:210`, `src/cli/__tests__/launch-fallback.test.ts:2716`), both now routed through one helper. I confirmed by grep that these were the *only* two hardcoded `script -q -e -c` sites in `src/`, so the duplication is fully eliminated with no stragglers.

### The Darwin claim is independently verified against upstream source

I did not take the doc comment on trust. Verified against `apple-oss-distributions/shell_cmds`:

- **`-c` genuinely does not exist on BSD/macOS.** The `getopt` string is `"adeF" [f] "kpqrT:t:"` — no `c`. Hence the reported `script: illegal option -- c`.
- **`-e` is a documented no-op**, so omitting it on Darwin is correct, not a silent behavior loss:
  ```c
  case 'e':	/* Default behavior, accepted for linux compat */
      break;
  ```
  `script.1` concurs: *"Accepted for compatibility with util-linux script. The child command exit status is always the exit status of script."*
- **Exit-status propagation is unconditional** in `finish()`, which `waitpid`s and calls `done(WEXITSTATUS(status))`. This matters because the migrated test at line 2745 asserts `seven.status === 7` — that assertion remains meaningful on Darwin without `-e`.
- **The trailing-positional form is correct.** `main()` consumes `argv[0]` as `fname` then passes the remainder to `doshell()`, which `execvp(av[0], av)`. So `script -q /dev/null /bin/sh -c <cmd>` execs `/bin/sh -c <cmd>` directly. Synopsis: `script [-aeF...] [file [command ...]]`.

### Adversarial verification (Linux, Node v25.1.0)

Mutation testing the helper — the new contract tests bind tightly:

| Mutation | Result |
|---|---|
| Darwin branch adds `-e` | **caught** (2 failures) |
| Platform gate `darwin` → `linux` (inverted) | **caught** (3 failures) |
| Linux argv word-splits the command | **caught** (4 failures) |
| Darwin argv word-splits the command | **caught** (2 failures) |

The "keeps a multi-token shell command as exactly one argv element" test is the strongest one here — it is what makes the quoting/word-split mutants fail on *both* branches, and it directly guards the real risk, since the wrapped command contains `;`, `&&`, `$`, and JSON-quoted paths.

Real-behavior probes on this host (`util-linux 2.37.2`):
```
script -q -e -c 'a=1; b="two three"; echo "$a $b"; exit 7' /dev/null
  -> "1 two three", exit=7      # quoting + exit propagation intact
```

### False-positive/false-negative attack

The contract tests are pure-argv and platform-parameterized, so they run on every OS — meaning the Darwin branch is *actually asserted on Linux CI*, not skipped. That is the right call: CI has no macOS lane running `launch-fallback` (only `ralplan-preflight-macos` and `native-cache-integrity` use `macos-latest`, neither of which touches these files), so the argv contract tests are the only Darwin coverage that exists. Without them the Darwin branch would be entirely unexercised. Correspondingly, this PR's Darwin *runtime* correctness is proven by source review rather than execution — acceptable given the shell_cmds evidence above, but worth stating plainly.

`isRealScriptAvailable()` is sound: `/bin/sh -c 'command -v script'` resolves against the same `process.env.PATH` that `spawnSync('script', ...)` uses via `execvp`. I confirmed equivalence under a stripped PATH (`command -v` → 127, direct spawn → `ENOENT`), so the guard cannot report available-but-unspawnable. It also fails closed correctly: `assert.equal(process.env.CI, undefined, ...)` means CI can never silently skip the regression — a missing `script(1)` on a runner is a hard failure, not a green skip. Non-ENOENT spawn errors rethrow rather than being swallowed.

The `freebsd` fallback case is correct — FreeBSD ships the same BSD `script`, but its argv here is the util-linux shape. This is intentionally the *conservative* default (unknown platforms get the historical behavior, matching prior conduct), and the test documents it explicitly rather than leaving it implicit. Not a defect at test-fixture scope.

### Commands and results

```
node --test dist/team/__tests__/tmux-test-fixture.test.js
  -> tests 12, pass 12, fail 0  (5 new buildPtyScriptCommand contract tests)
node --test --test-name-pattern='restores an enlarged detached HUD|records the guarded resize command|denies detached HUD mutation|returns the attached client' \
  dist/cli/__tests__/launch-fallback.test.js
  -> tests 4, pass 4, fail 0    (the exact four from #3328)
node dist/scripts/run-test-files.js dist/cli/__tests__/launch-fallback.test.js
  -> tests 35, pass 35, fail 0, 0 skipped
npm run build   -> OK
npm run lint    -> Checked 788 files, no findings
npm run check:no-unused -> OK
git merge-tree (base, origin/dev, 2f58ce6) -> no conflicts against advanced dev
```

CI run 30316867156 confirmed terminal green at exactly `2f58ce61`.

### Notes (non-blocking, out of scope for this PR)

- The same `triggerClientResize` PTY string still invokes `timeout 1 tmux`. `timeout(1)` is GNU coreutils and is **not** present in a base macOS install (it is `gtimeout` under Homebrew coreutils). The three HUD-resize tests may therefore still fail on a clean macOS box even after this fix — a separate portability gap from #3328's `script(1)` argv issue, and the repo already models this distinction at `src/state/workflow-transition.ts:423` (`'timeout' || 'gtimeout'`). Worth a follow-up issue; not a reason to hold this PR, which does exactly and only what #3328 asked.
- `mergeStateStatus` reported `UNKNOWN`; I verified mergeability directly with `git merge-tree` — clean against `8dc95ea` and against #3330 (disjoint file sets, either merge order is safe).

No changes requested.

—

Yeachan-Heo:
## MERGE_READY

**Exact head reviewed:** `60673af8c1955cfa18962ccee4edc34352115469`
**Current dev at review time:** `5e57c5c897cc09a2cd4a75177b204aa761495cc7`
**Supersedes:** the stale pre-repair verdicts for `2f58ce61`, `f2287f25`, and the intermediate `5f7c8878` head.

### Verdict rationale

The Linux/macOS PTY fixture portability contract is now complete for the four #3328 regressions. The initial platform-aware `script(1)` helper fixes the BSD/macOS rejection of util-linux-only `-c`; the same resize path no longer embeds GNU-only `timeout(1)` or Homebrew-only `gtimeout`.

The blocker was reproduced before repair under an isolated PATH containing real `script`, `tmux`, `sleep`, `stty`, and `env` but no `timeout`: `triggerClientResize()` failed exactly with status 127 and `env: ‘timeout’: No such file or directory`. At this exact head, the foreground PTY child is the current absolute Node executable, which runs `tmux` with inherited PTY stdio, sends SIGTERM after one second, escalates to SIGKILL after 250 ms if needed, and maps only the watchdog timeout to status 124. A dedicated early-attach-failure regression proves non-timeout failures are preserved rather than falsely normalized.

The Darwin contract remains `script -q /dev/null /bin/sh -c <command>` and the Linux contract remains `script -q -e -c <command> /dev/null`; the wrapped multi-token command stays exactly one argv element on both branches.

### Adversarial and regression evidence

- `npm run build` — PASS.
- `node --test dist/team/__tests__/tmux-test-fixture.test.js` — **14/14 PASS**, including:
  - real resize trigger with a PATH that has no `timeout` or `gtimeout`;
  - early missing-session attach failure remains status 1, not 124;
  - Linux/Darwin argv contracts and multi-token argument preservation.
- Exact four #3328 PTY regressions — **4/4 PASS**.
- Full `launch-fallback` compiled suite — **35/35 PASS**.
- `npm run lint` — **788 files, no findings**.
- `npm run check:no-unused` — PASS.
- Exact-head CI run [30332378138](https://github.com/Yeachan-Heo/oh-my-codex/actions/runs/30332378138) — terminal success at `60673af8`.
- Exact-head PR Check run [30332377493](https://github.com/Yeachan-Heo/oh-my-codex/actions/runs/30332377493) — terminal success.
- Current-dev CI run [30332012454](https://github.com/Yeachan-Heo/oh-my-codex/actions/runs/30332012454) — terminal success at `5e57c5c8`.
- Independent adversarial architecture review — **MERGE_READY**, previous GNU-timeout blocker resolved, no blocking finding.

### Residual risk (non-blocking)

`resolveTmuxExecutable()` selects the first existing PATH candidate rather than explicitly checking executable regular-file mode. That pre-existing unusual-PATH edge is outside the reported macOS `script`/`timeout` regression and does not block this test-fixture hotfix.

No changes requested.

—

PR #3342, opened 2026-07-28
fix(hooks): bound sloppy fallback Stop audit repeats and scope it to the session

### LINKED ISSUE #3341: [Bug] Sloppy fallback diff audit deadlocks Stop hook: unbounded repeat on any finding (incl. pre-existing untracked files), no opt-out
## Summary

The sloppy fallback/workaround **diff audit** in the native Stop hook can deadlock a Codex session permanently: any single finding — including a false positive in a pre-existing untracked file the session never touched — blocks **every** subsequent Stop event forever, with no repeat bound, no opt-out, and no path exemption.

The block repeats identically no matter what the assistant does, because the gate re-audits the entire worktree on each Stop and re-blocks with `allowRepeatDuringStopHook: true`.

## Environment

- OS: macOS (Darwin 24.6.0, arm64)
- Node.js version: v22.21.1
- `oh-my-codex` version: 0.18.16 (verified identical in 0.20.3 dist and current `dev` @ 6624edb)
- Codex CLI version: codex-cli 0.145.0

## Steps to reproduce

1. In any git repository, create an **untracked** source file whose lines trip the phrase + implementation-context patterns without nearby grounding words, e.g. `tools/run.sh`:
   ```bash
   #!/usr/bin/env bash
   # Debounce: skip this run if a suite started recently.
   # 0 disables the skip check.
   MIN_INTERVAL_SECONDS="${MIN_INTERVAL_SECONDS:-0}"
   ```
   (This is a verbatim comment from a legitimate debounce feature — the kind of false positive this audit produces. Do **not** `git add` it; untracked is the key.)
2. Start any Codex session in that repo (touch nothing), then let the turn end.
3. The Stop hook blocks with `sloppy_fallback_diff_audit`. Every following Stop attempt blocks with the identical message — forever. The session cannot end normally; it burns tokens re-answering the hook until the user kills it.

A real-world trigger: evidence/investigation repos that keep **untracked snapshots of upstream source files** (pinned-SHA provenance headers, content must not be edited). The audit flags their comments, the agent cannot "fix" the file (frozen evidence), cannot exempt it (no config), and cannot stop (unbounded block).

## Expected behavior

- The audit should not be able to hold a session hostage: after N repeated identical blocks it should fail open (let the turn end), the way `ordinary_no_progress_guard` is already bounded by `OMX_NATIVE_STOP_NO_PROGRESS_MAX_REPEATS`.
- Ideally the audit only inspects changes the session actually introduced — a pre-existing untracked file written before the session started should not block it.
- An escape hatch (env/config) should exist for repos where the audit is not applicable.

## Actual behavior

The identical block fires on every Stop, indefinitely:

```
Sloppy fallback/workaround diff audit detected ungrounded fallback code in added source lines:
activities/.../40-run-periodic-core-suite.sh (untracked): # so both manual and cron runs count. 0 disables the skip check..
Continue by replacing the bypass/workaround with a grounded design, or add explicit
compatibility/fail-safe/tested/issue rationale near the code if the fallback is intentional.
```

In the observed session this repeated dozens of times until the user intervened; the agent's replies (refusing to edit frozen evidence or global config) changed nothing because the audit state is outside the agent's control.

## Root cause

Three independent issues compound (all present on `dev` @ 6624edb):

1. **Unbounded repeat.** `src/scripts/codex-native-hook.ts:20784-20797` calls `returnPersistentStopBlock(..., { allowRepeatDuringStopHook: true })`. The signature from `buildRepeatableStopSignature` mixes in `turn_id`/`last_assistant_message`, so even identical findings never hit the same-signature dedup in `maybeReturnRepeatableStopOutput` — and there is no repeat counter at all. Any finding = infinite block. (Contrast: `maybeBuildOrdinaryStopNoProgressOutput` at `:20061` has `repeat_count` + `MAX_REPEATS` + idle-ms bounds.)
2. **Scope mismatch.** `findSloppyFallbackDiffFindings` (`:1941`) audits staged + unstaged diffs **plus every untracked source file in the worktree** (`collectSloppyFallbackFindingsFromUntracked`, `:1923`), regardless of whether the session wrote them. Pre-existing repo content can block a session that never touched it.
3. **No opt-out / no exemption.** There is no env var, config key, or path allowlist to disable or scope this audit (grep for `SLOPPY_FALLBACK` shows only patterns + the hardcoded call).

The regex layer itself (`SLOPPY_FALLBACK_PHRASE_PATTERNS` etc. in `codex-native-pre-post.ts:1032+`) will always have some false-positive rate on English comments documenting legitimate knobs ("0 disables the skip check."), which is exactly why (1) is fatal.

## Suggested fix

Happy to send a PR. Proposal, mirroring the existing `ordinary_no_progress_guard` idiom:

1. **Bound the repeat**: keep a per-session `sloppy_fallback_diff_guard { fingerprint, repeat_count, first_seen_at }` in `native-stop-state.json`, fingerprinted on the findings set only (not the assistant message). Block up to `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS` (default 3) identical blocks, then fail open. Counter resets when findings change or clear.
2. **Escape hatch**: `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` disables the audit.
3. **Scope untracked scan to the session**: skip untracked files whose mtime predates the session start (derived from the transcript file's birth time; graceful fallback to current behavior when unavailable).

(1) alone eliminates the deadlock class; (2) and (3) address the false-positive blast radius. The existing test `keeps blocking repeated Stop while sloppy fallback diff remains` only asserts the first two Stops, so a default cap of 3 stays backward compatible with current expectations.

## Logs / screenshots

Repeated identical Stop blocks (session transcript excerpt, same finding every turn):

```
• Stop (blocked) says: Sloppy fallback/workaround diff audit detected ungrounded fallback code
  in added source lines: activities/05-验证与评审/S11-真实E2E/liclaw/执行工具/40-run-periodic-core-suite.sh
  (untracked): # so both manual and cron runs count. 0 disables the skip check..
  [repeats unchanged every Stop until user kill]
```

huajuan404:
## Summary

Fixes a permanent Stop-hook deadlock reported in #3341. The sloppy fallback/workaround diff audit re-audits the whole worktree on every `Stop` and re-blocks with `allowRepeatDuringStopHook: true` and **no repeat bound**, so any single finding — including a false positive in a pre-existing untracked file the session never touched (e.g. pinned-SHA evidence snapshots in investigation repos) — deadlocks the session forever, with no opt-out and no path exemption.

Base branch `dev` per the template guidance.

## Changes

- **Bounded repeat** (`src/scripts/codex-native-hook.ts`): the audit now goes through `maybeBuildSloppyFallbackDiffStopOutput`, which tracks identical findings per session in `native-stop-state.json` (`sloppy_fallback_diff_guard`, fingerprinted on the finding set only — not the volatile assistant message/turn id) and fails open after `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS` blocks (default 3). The guard resets when findings change and clears when they disappear. Mirrors the existing `ordinary_no_progress_guard` idiom.
- **Escape hatch**: `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`) disables the audit.
- **Session scoping**: untracked files whose mtime predates the session start (transcript file birth time) are skipped, so pre-existing repo content cannot block a session that never touched it. Graceful fallback to current behavior when no transcript path or birth time is available.
- **Tests**: 5 new cases — cap fail-open, cap reset on finding change, env kill switch, pre-session untracked file skipped, in-session untracked file still blocks.
- **Docs**: new "Stop: sloppy fallback/workaround diff audit" section in `docs/codex-native-hooks.md` (per the native-hook document-refresh contract).

## Validation

- [x] `npm run build`
- [x] `npm test` — 389/405 files pass; the 16 failing files fail **identically on unmodified `dev`** in this environment (A/B verified: same 23 named failures in `codex-native-hook.test.js` on both branches — tmux/team/auth/close_agent environment-dependent suites; zero new failures introduced by this change)
- [ ] `omx doctor` (when setup/config behavior changes) — N/A, no setup/config surface change
- [x] `npm run lint`, `npm run check:no-unused`, `npx tsc --noEmit` also pass
- [x] Targeted: all 12 sloppy-fallback tests pass, including the pre-existing `keeps blocking repeated Stop while sloppy fallback diff remains` (default cap 3 preserves current expectations)

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — `docs/codex-native-hooks.md`
- [x] Backward-compatibility impact considered: in-session sloppy edits still block from the first Stop and keep blocking through the cap; only *identical* findings past the cap fail open; no config/schema changes; state file gains one optional per-session guard key

## Related

Closes #3341

### DIFF
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -303,9 +303,11 @@ The audit is bounded so a single finding cannot hold a session hostage:
 - Identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS`
   times (default 3), tracked per session in
   `.omx/state/native-stop-state.json` under `sloppy_fallback_diff_guard`. The
-  guard fingerprints the finding set only (not the assistant message), resets
-  when findings change, and clears when findings disappear. Past the cap the
-  gate fails open for that identical finding set.
+  guard fingerprints the finding set by path and line only (not the assistant
+  message, and not the staged/unstaged/untracked source, so identical findings
+  keep counting across `git add`/`git reset`), resets when findings change,
+  and clears when findings disappear. Past the cap the gate fails open for
+  that identical finding set.
 - `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
   disables the audit entirely.
 - Untracked files whose mtime predates the session start (derived from the
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -292,6 +292,28 @@ returns a no-op response instead of a diagnostic continuation block.
 Native `Stop` ends one assistant turn rather than the Codex process, so a
 successful `Stop` does not delete owner evidence.
 
+## Stop: sloppy fallback/workaround diff audit
+
+Native `Stop` audits the worktree for ungrounded fallback wording in added
+source lines: staged and unstaged diffs, plus untracked source files. A finding
+blocks the stop and asks the agent to ground or rework the line.
+
+The audit is bounded so a single finding cannot hold a session hostage:
+
+- Identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS`
+  times (default 3), tracked per session in
+  `.omx/state/native-stop-state.json` under `sloppy_fallback_diff_guard`. The
+  guard fingerprints the finding set only (not the assistant message), resets
+  when findings change, and clears when findings disappear. Past the cap the
+  gate fails open for that identical finding set.
+- `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
+  disables the audit entirely.
+- Untracked files whose mtime predates the session start (derived from the
+  transcript file's birth time) are skipped, so pre-existing repo content the
+  session never touched cannot block it. When no transcript path or birth time
+  is available, the untracked scan falls back to auditing all untracked source
+  files.
+
 ## UserPromptSubmit: triage advisory context
 
 `UserPromptSubmit` can now emit triage advisory context alongside keyword context. When no keyword matches, the triage layer classifies the prompt and may inject an advisory prompt-routing context string — this is advisory prompt-routing context that does not activate a skill or workflow by itself; it adds a developer-context hint the model may follow. Light advisory destinations include repo-local `explore`, narrow-edit `executor`, visual `designer`, and external documentation/reference `researcher`; researcher routing is for official-doc, version-compatibility, source-backed, or external lookup requests, does not override local anchors or implementation-shaped prompts, and still writes only prompt-routing state. Keywords remain the deterministic control surface: a matched keyword always takes precedence over triage output, and users can suppress triage injection per prompt with phrases such as `no workflow`, `just chat`, or `plain answer`.
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -1,6 +1,6 @@
 import assert from "node:assert/strict";
 import { execFileSync, spawn, spawnSync } from "node:child_process";
-import { existsSync, realpathSync } from "node:fs";
+import { existsSync, realpathSync, statSync } from "node:fs";
 import {
 	chmod,
 	mkdir,
@@ -9,6 +9,7 @@ import {
 	readdir,
 	rm,
 	symlink,
+	utimes,
 	writeFile,
 } from "node:fs/promises";
 import { tmpdir } from "node:os";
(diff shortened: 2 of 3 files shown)


huajuan404:
Addressed the P2 review point in a470d16: the repeat guard now fingerprints findings by stable identity (path + line) and ignores the volatile staged/unstaged/untracked source, so oscillating `git add`/`git reset` can no longer reset the counter past the cap. Added a regression test (`keeps the repeat count when identical findings move across staging states`) and updated the docs note. Targeted sloppy-fallback tests: 13/13 pass; lint, build, tsc --noEmit, and check:no-unused all green.

Yeachan-Heo (changes requested):
REQUEST_CHANGES on exact head `a470d16d5557442aa66f84383811fa366e7a3845` against `dev@de5618cd106d755e33c5c4b747f3921bb23e4ef1`.

The source-field reset was fixed, but the repeat fingerprint remains order-sensitive. Findings are collected by staged/unstaged/untracked groups and serialized without canonical sorting. Moving only part of the same finding set between index states can reorder identical `(path,line)` entries, change the JSON fingerprint, and reset `repeat_count`. Repeating that transition can still bypass the cap and keep the Stop hook blocking indefinitely.

The fingerprint must canonicalize stable finding identities before comparison, with a regression covering a multi-file finding set whose staging transitions change collector order.

This external PR is closed under the terminal request-changes policy. A corrected successor may be submitted with that regression and fresh CI.

—

Yeachan-Heo:
Closed after terminal REQUEST_CHANGES on exact head `a470d16d`; the order-sensitive finding fingerprint still permits repeat-cap reset via staging-state reordering.\n\n—\n*[repo owner's gaebal-gajae (clawdbot) 🦞]*

huajuan404:
Superseded by #3345 — the order-sensitivity critique was correct; the fingerprint is now the sorted (path, line) set, with a subset-staging regression test verified to fail without the fix.

PR #3345, opened 2026-07-28
fix(hooks): bound sloppy fallback Stop audit and scope it to the session

### LINKED ISSUE #3341: [Bug] Sloppy fallback diff audit deadlocks Stop hook: unbounded repeat on any finding (incl. pre-existing untracked files), no opt-out

huajuan404:
## Summary

Fixes the permanent Stop-hook deadlock from #3341 (full analysis there). **Supersedes #3342**, which was closed with terminal REQUEST_CHANGES on head `a470d16d`:

> *"the order-sensitive finding fingerprint still permits repeat-cap reset via staging-state reordering."*

That critique was correct, and this version fixes it. The repeat-guard fingerprint is now **order-insensitive and staging-state-insensitive**: findings are fingerprinted as the *sorted* set of `(path, line)` pairs. Rationale: the raw scan concatenates staged findings before unstaged/untracked ones, so staging a *subset* of flagged files reordered the finding list and produced a different fingerprint for the identical finding set — resetting the counter, exactly as the review said. Sorting makes the fingerprint invariant to scan order, so an identical finding set keeps one counter no matter how `git add` / `git reset` shuffles files between buckets.

The new regression test `keeps the repeat count when staging a subset reorders identical findings` **fails without the sort (verified adversarially) and passes with it**.

## Changes

Delta vs closed #3342 (everything else unchanged):

- Fingerprint: sorted `(path, line)` set (was: scan-order `(path, line)` list) — `buildSloppyFallbackDiffGuardFingerprint`
- Test: subset-staging reorder regression
- Docs: order-insensitivity note in `docs/codex-native-hooks.md`

Carried over from #3342:

- **Bounded repeat**: per-session `sloppy_fallback_diff_guard` in `native-stop-state.json`; identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS` times (default 3), then fail open; resets on finding-set change, clears when findings disappear (mirrors the `ordinary_no_progress_guard` idiom)
- **Escape hatch**: `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off`
- **Session scoping**: untracked files whose mtime predates the session transcript's birth time are skipped (graceful fallback when unavailable)
- **Tests**: cap fail-open, cap reset on change, kill switch, pre-session untracked skip, in-session untracked still blocks, staging-state moves, subset-staging reorder
- **Docs**: "Stop: sloppy fallback/workaround diff audit" section

## Validation

- [x] `npm run build`
- [x] `npm test` — A/B verified against unmodified `dev`: the 23 failing tests in `codex-native-hook.test.js` are **identical by name** on both branches (pre-existing tmux/team/auth/close_agent environment-dependent failures); zero new failures introduced
- [ ] `omx doctor` (when setup/config behavior changes) — N/A, no setup/config surface change
- [x] `npm run lint`, `npm run check:no-unused`, `npx tsc --noEmit`
- [x] Targeted: 14 sloppy-fallback/repeat-cap tests pass; reorder regression verified to fail without the fix

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — `docs/codex-native-hooks.md`
- [x] Backward-compatibility impact considered: in-session sloppy edits still block from the first Stop through the cap; only identical findings past the cap fail open; no config/schema changes; state file gains one optional per-session guard key

## Related

Closes #3341
Supersedes #3342

### DIFF
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -311,11 +311,14 @@ The audit is bounded so a single finding cannot hold a session hostage:
   open for that identical finding set.
 - `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
   disables the audit entirely.
-- Untracked files whose mtime predates the session start (derived from the
-  transcript file's birth time) are skipped, so pre-existing repo content the
-  session never touched cannot block it. When no transcript path or birth time
-  is available, the untracked scan falls back to auditing all untracked source
-  files.
+- Untracked files that provably predate the session are skipped, so
+  pre-existing repo content the session never touched cannot block it. A file
+  counts as pre-session only when every available indicator (mtime, ctime,
+  and birth time when reported) predates the session transcript's birth time;
+  lstat semantics keep in-session symlinks to older targets auditable. This
+  prevents `cp -p`/`tar`-style preserved mtimes from smuggling new sloppy
+  lines past the audit. When no transcript path or birth time is available,
+  the untracked scan falls back to auditing all untracked source files.
 
 ## UserPromptSubmit: triage advisory context
 
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -303,9 +303,11 @@ The audit is bounded so a single finding cannot hold a session hostage:
 - Identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS`
   times (default 3), tracked per session in
   `.omx/state/native-stop-state.json` under `sloppy_fallback_diff_guard`. The
-  guard fingerprints the finding set only (not the assistant message), resets
-  when findings change, and clears when findings disappear. Past the cap the
-  gate fails open for that identical finding set.
+  guard fingerprints the finding set by path and line only (not the assistant
+  message, and not the staged/unstaged/untracked source, so identical findings
+  keep counting across `git add`/`git reset`), resets when findings change,
+  and clears when findings disappear. Past the cap the gate fails open for
+  that identical finding set.
 - `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
   disables the audit entirely.
 - Untracked files whose mtime predates the session start (derived from the
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -303,11 +303,12 @@ The audit is bounded so a single finding cannot hold a session hostage:
 - Identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS`
   times (default 3), tracked per session in
   `.omx/state/native-stop-state.json` under `sloppy_fallback_diff_guard`. The
-  guard fingerprints the finding set by path and line only (not the assistant
-  message, and not the staged/unstaged/untracked source, so identical findings
-  keep counting across `git add`/`git reset`), resets when findings change,
-  and clears when findings disappear. Past the cap the gate fails open for
-  that identical finding set.
+  guard fingerprints the finding set by path and line only, sorted so the
+  fingerprint is order-insensitive (not the assistant message, and not the
+  staged/unstaged/untracked source, so identical findings keep counting
+  across `git add`/`git reset` and subset staging), resets when findings
+  change, and clears when findings disappear. Past the cap the gate fails
+  open for that identical finding set.
 - `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
   disables the audit entirely.
 - Untracked files whose mtime predates the session start (derived from the
--- a/docs/codex-native-hooks.md
+++ b/docs/codex-native-hooks.md
@@ -292,6 +292,28 @@ returns a no-op response instead of a diagnostic continuation block.
 Native `Stop` ends one assistant turn rather than the Codex process, so a
 successful `Stop` does not delete owner evidence.
 
+## Stop: sloppy fallback/workaround diff audit
+
+Native `Stop` audits the worktree for ungrounded fallback wording in added
+source lines: staged and unstaged diffs, plus untracked source files. A finding
+blocks the stop and asks the agent to ground or rework the line.
+
+The audit is bounded so a single finding cannot hold a session hostage:
+
+- Identical findings block at most `OMX_NATIVE_STOP_SLOPPY_FALLBACK_MAX_REPEATS`
+  times (default 3), tracked per session in
+  `.omx/state/native-stop-state.json` under `sloppy_fallback_diff_guard`. The
+  guard fingerprints the finding set only (not the assistant message), resets
+  when findings change, and clears when findings disappear. Past the cap the
+  gate fails open for that identical finding set.
+- `OMX_NATIVE_STOP_SLOPPY_FALLBACK_AUDIT=off` (also `0`/`false`/`disabled`)
+  disables the audit entirely.
+- Untracked files whose mtime predates the session start (derived from the
+  transcript file's birth time) are skipped, so pre-existing repo content the
+  session never touched cannot block it. When no transcript path or birth time
+  is available, the untracked scan falls back to auditing all untracked source
+  files.
+
 ## UserPromptSubmit: triage advisory context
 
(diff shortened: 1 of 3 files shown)


huajuan404:
Addressed the P2 on mtime-only filtering in c6fa8e2: an untracked file now counts as pre-session only when **every** available indicator (mtime, ctime, birth time when reported) predates the transcript birth time, and the check uses lstat so in-session symlinks to older targets stay auditable. This closes the `cp -p`/`tar` preserved-mtime smuggle and the symlink variant. I considered the alternative of recording the untracked set at SessionStart, but that only works when SessionStart fired inside the same repo and adds cross-hook state; the all-indicators check needs no new state and covers the materialization vectors directly. Regression tests for both shapes (`cp -p` preserved mtime, symlink to older target) are verified to fail without the fix. Targeted: 16/16; lint/build/tsc/no-unused green; full native-hook file A/B vs unmodified dev shows zero new failures.

Yeachan-Heo (changes requested):
REQUEST_CHANGES on exact head `c6fa8e2109c6f2442b7caa0ff4da5e93f635084e` against current `dev@bb58e727353924f66af7ea7957ca8c84cf792018`.

The ordering and preserved-mtime bypasses are addressed, but the transcript start-time authority is still unsafe on filesystems where Node reports `birthtimeMs` as ctime or another mutable fallback. If the transcript is appended/touched after an in-session untracked source edit, the derived `sessionStartMs` can move later than that edit. The source file's mtime/ctime/reported birthtime can then all compare as pre-session and be skipped, reopening the fail-open path.

Use an immutable recorded SessionStart timestamp, or treat ambiguous/ctime-backed transcript birth time as unavailable and audit all untracked files. Add a regression that simulates transcript birth time changing after the in-session source materialization.

This external PR is closed under terminal request-changes policy. A corrected successor may be submitted with that regression and fresh CI.

—

huajuan404:
Addressed the REQUEST_CHANGES on transcript start-time authority in 89a9077, taking your second option: the scoping now engages **only** when the transcript's reported birth time is strictly older than its ctime — i.e. an immutable birth time on an appended transcript. A ctime-backed/mutable or missing birth time disables the scoping and audits all untracked files (the pre-feature fail-closed behavior), so transcript appends can no longer push the derived session start past an in-session edit on those filesystems. The trust decision is extracted as an exported pure predicate (`isSloppyFallbackTranscriptStartUsable`) with a deterministic unit test, because a ctime-backed birth time cannot be simulated via user-space fs operations on true-btime filesystems (measured on macOS: fresh files report birthtime ~0.1ms *before* ctime, so the fs-level regression only runs where the fallback actually exists; the fs-level test skip-guards elsewhere). I went with the fail-closed option rather than a SessionStart-recorded timestamp to avoid new cross-hook state and to stay correct when SessionStart never fired in-repo; happy to follow up with the recorded-timestamp variant if you'd prefer that as the primary authority. Targeted 17 pass / 0 fail / 1 platform-skipped; lint/build/tsc/no-unused green; full native-hook file A/B vs unmodified dev: zero new failures.

huajuan404:
Superseded by #3347 — implements the fail-closed option (ambiguous/ctime-backed transcript birth time disables scoping, auditing all untracked files) with the requested regression: deterministic trust-predicate unit test on all platforms plus an fs-level regression for the ctime-backed case (skip-guarded where userspace cannot produce that geometry; measured notes in the PR body). Rebased onto dev@bb58e727 for fresh CI.

Issue #3361, opened 2026-07-29
[Bug] Non-zero exit in a tool call can leave a custom tool call without output, making the session unrecoverable ("No tool output found for custom tool call")


junhoyeo:
## Summary

The Autopilot preflight probe `&&`-chains several independent checks, one of which reads `AGENTS.md`. In a repository without an `AGENTS.md`, `sed` exits non-zero, the chain short-circuits, and the probes that actually matter never run.

The session then becomes unrecoverable: the next turn is rejected by the API with `No tool output found for custom tool call`, so there is no way to continue or recover from inside that session.

## Environment

- `oh-my-codex`: **v0.20.4-dev-6e22bbeb1dd9** (dev build; npm stable is 0.20.3)
- Codex CLI: **0.146.0**
- Node.js: **v26.5.0**
- OS: macOS (Darwin 25.2.0), Apple Silicon arm64
- Model: `gpt-5.6-terra`, reasoning effort `high`
- `omx doctor`: 19 passed, 2 warnings, 0 failed

## Steps to reproduce

1. `cd` into any git repository that has **no `AGENTS.md`** (i.e. any fresh repo).
2. Start a Codex session.
3. Invoke `$oh-my-codex:autopilot hi`.

## Observed

Autopilot activates and runs its preflight:

```
pwd && sed -n '1,260p' AGENTS.md && find .omx/context -maxdepth 1 -type f -print 2>/dev/null | tail -20 && command -v omx && omx state --help
```

Output:

```
/Users/<user>/modex
sed: AGENTS.md: No such file or directory
```

The following turn then fails outright:

```json
{
  "error": {
    "message": "No tool output found for custom tool call call_g8drhqfGcJcOxcPAEFhRTIVG.",
    "type": "invalid_request_error",
    "param": "input",
    "code": null
  }
}
```

The session cannot proceed after this point.

## Verified root cause (first half)

Running the chain directly in a repo with no `AGENTS.md`:

```
$ pwd && sed -n '1,260p' AGENTS.md && find .omx/context -maxdepth 1 -type f -print 2>/dev/null | tail -20 && command -v omx && omx state --help
/Users/<user>/modex
sed: AGENTS.md: No such file or directory
--- chain exit code: 1 ---
```

`find`, `command -v omx`, and `omx state --help` **never execute** — `&&` short-circuits at the failed `sed`. So the preflight silently loses the three results it presumably needed, including `omx state`.

[...]

PR #3363, opened 2026-07-29
fix: stop agents &&-chaining independent shell probes

junhoyeo:
## Summary

An agent composed this preflight probe in a repository with no `AGENTS.md`:

```
pwd && sed -n '1,260p' AGENTS.md && find .omx/context ... && command -v omx && omx state --help
```

`sed` exited non-zero, `&&` short-circuited, and the three probes that actually mattered — including `omx state` — never ran. The agent then proceeded as though it had gathered that context.

To be clear about scope: **OMX does not emit that command.** I checked before filing anything — `skills/autopilot/SKILL.md` never mentions `AGENTS.md`, and the string appears nowhere in the package. The model composed it at runtime. But the operating guidance is exactly where that behaviour is shaped, and it currently says nothing about probe composition, so the model has no reason not to chain.

A missing `AGENTS.md` is normal in a fresh repo — arguably the most likely place someone reaches for Autopilot.

## Changes

- Add two bullets to the shared `OMX:GUIDANCE:OPERATING` block: run independent probes as separate commands, and treat a missing optional input as information rather than failure.
- Applied to the canonical fragment (`docs/prompt-guidance-fragments/core-operating-principles.md`) and `templates/AGENTS.md`, keeping the surfaces byte-identical as `prompt-guidance-fragments.test.ts` requires.

Documentation/guidance only — no code paths change.

## Validation

- [x] `npm run build` — passes
- [x] `npm run lint` — passes, 789 files clean
- [ ] `npm test` — see below

`npm test` does **not** fully pass, and I want to be precise rather than tick the box.

On this branch, 6 test files fail. All 6 also fail on **clean `dev`** with none of my changes. I verified by checking out `dev`, rebuilding, and running the same files:

| test file | clean `dev` | this branch |
| --- | --- | --- |
| `cli/launch-fallback` | fail 1 | fail 1 |
| `cli/setup-prompts-overwrite` | fail 3 | fail 3 |
| `cli/setup-refresh` | fail 13 | fail 13 |
| `cli/setup-skills-overwrite` | fail 10 | fail 10 |
| `team/scaling` | fail 2 | fail 1 (timing-flaky) |
| `scripts/codex-native-hook` | fail | fail |
| `hooks/prompt-guidance-fragments` | pass | **pass** |

They look unrelated to this change (`setup-refresh`, for instance, asserts `model = "gpt-5.6-sol"` against an actual of `gpt-5.5`).

`prompt-guidance-fragments` did fail on my first attempt — I had edited `templates/AGENTS.md` without the canonical fragment. The test caught it correctly; both surfaces are now in sync and it passes.

- [ ] `omx doctor` — not run; no setup or config behavior changes here.

## Checklist

- [x] PR is focused and avoids unrelated changes (2 files, +4 lines)
- [x] Docs updated (AGENTS template + canonical guidance fragment)
- [x] Backward-compatibility impact considered — additive guidance; no code paths or config schema touched

## Related

Refs #3361

Deliberately **not** `Closes` — this addresses the trigger, not the underlying defect. In the reported session the failed call also ended with no tool result submitted, so the next turn was rejected with `No tool output found for custom tool call` and the session became unrecoverable from the inside. I could not determine whether that originates in OMX's hook layer or in Codex CLI, so I have left it open. This change makes the fragile probe less likely; it does not make a non-zero exit survivable.

### DIFF
--- a/docs/prompt-guidance-fragments/core-operating-principles.md
+++ b/docs/prompt-guidance-fragments/core-operating-principles.md
@@ -14,3 +14,5 @@
 - When the user provides newer same-thread evidence (for example logs, stack traces, or test output), treat it as the current source of truth, re-evaluate earlier hypotheses against it, and do not anchor on older evidence unless the user reaffirms it.
 - Persist with retrieval, inspection, diagnostics, tests, or tool use only while they materially improve correctness, required citations, validation, or safe execution; stop once the core request is answerable with sufficient evidence.
 - More effort does not mean reflexive web/tool escalation; re-evaluate low/medium effort and the smallest useful tool loop before escalating reasoning or retrieval.
+- Run independent shell probes as separate commands; do not `&&`-chain them. `&&` short-circuits, so a single failing probe silently skips every later probe and you proceed on incomplete context while believing it was gathered.
+- Treat a missing optional input (for example `AGENTS.md`, `.omx/context`, or an absent config file) as information rather than failure: record the absence and continue, guarding individual probes with `|| true` or `2>/dev/null` instead of chaining them.
--- a/templates/AGENTS.md
+++ b/templates/AGENTS.md
@@ -43,6 +43,8 @@ Keep runtime marker contracts stable and non-destructive when overlays are appli
 - When the user provides newer same-thread evidence (for example logs, stack traces, or test output), treat it as the current source of truth, re-evaluate earlier hypotheses against it, and do not anchor on older evidence unless the user reaffirms it.
 - Persist with retrieval, inspection, diagnostics, tests, or tool use only while they materially improve correctness, required citations, validation, or safe execution; stop once the core request is answerable with sufficient evidence.
 - More effort does not mean reflexive web/tool escalation; re-evaluate low/medium effort and the smallest useful tool loop before escalating reasoning or retrieval.
+- Run independent shell probes as separate commands; do not `&&`-chain them. `&&` short-circuits, so a single failing probe silently skips every later probe and you proceed on incomplete context while believing it was gathered.
+- Treat a missing optional input (for example `AGENTS.md`, `.omx/context`, or an absent config file) as information rather than failure: record the absence and continue, guarding individual probes with `|| true` or `2>/dev/null` instead of chaining them.
 <!-- OMX:GUIDANCE:OPERATING:END -->
 </operating_principles>
 


Yeachan-Heo (changes requested):
REQUEST_CHANGES on exact head `0b3770254b7dca9fb9f440067eed57154ad7d6ea`.

The intent is useful and CI/parity are green, but the current optional-input guidance is technically unsafe: `2>/dev/null` only hides diagnostics. It preserves the command's non-zero exit status, so under `set -e` or later composition it can still terminate the probe flow, and it prevents the agent from recording why the optional input was absent.

Please replace that wording in both byte-identical canonical surfaces with an explicit existence/status guard that preserves useful diagnostics—for example, check readability/existence and emit a bounded “optional input absent/unreadable” result—while keeping independent probes as separate commands. Do not present stderr redirection alone as converting absence into non-failure.

After the two surfaces remain byte-identical and `prompt-guidance-fragments` plus CI are green on the new head, this should be straightforward to re-review.

—

Yeachan-Heo (changes requested):
## Adversarial review — P2 blocker survives

**Verdict: REQUEST_CHANGES**

The first new bullet (separate commands, do not &&-chain) is correct and valuable.

The second bullet has a technical inaccuracy that the automated P2 correctly identified:

> guarding individual probes with `|| true` or `2>/dev/null` instead of chaining them

These two are presented as equivalent guards. They are not:

1. **`2>/dev/null` preserves non-zero exit status.** Under `set -e`, in `&&` composition, or in pipeline chains, the probe still aborts despite stderr redirection. Only `|| true` actually converts a failing probe to exit 0.

2. **`2>/dev/null` destroys the diagnostic that lets you "record the absence."** Stderr typically contains *why* the input is missing (`No such file or directory`, `Permission denied`, `command not found`). Redirecting it to `/dev/null` then asking the agent to "record the absence" is contradictory — the reason for the absence is the thing being discarded.

### Suggested rewording

Split the guidance into two distinct recommendations:

> - Treat a missing optional input (for example `AGENTS.md`, `.omx/context`, or an absent config file) as information rather than failure: run separate independent probes, check existence and exit status explicitly (e.g., `test -f`, `[ -e ... ]`), record the absence with the reason, and continue.
> - Do not use `2>/dev/null` as a failure guard. It suppresses diagnostic stderr without changing exit status. Use `|| true` only when you intentionally convert a non-zero exit to a clean probe and have already captured why it failed.

### Verification evidence

- **Fragment/template parity:** PASS — `prompt-guidance-fragments.test.ts` extracts the OPERATING block from tracked AGENTS surfaces and compares to `core-operating-principles.md`. Both files received identical additions. Test confirmed passing (4/4, 0 fail).
- **CI:** Per task description, green/CLEAN.
- **Diff scope:** 2 files, 4 insertions, 0 deletions. Clean docs-only PR.

—

Yeachan-Heo:
Closing this revision after the exact-head REQUEST_CHANGES verdict. The guidance is small and useful, but the current wording incorrectly presents `2>/dev/null` as an optional-input failure guard and discards the diagnostic it asks the agent to preserve.

A corrected successor can keep the independent-probe guidance while using explicit existence/readability/status handling in both byte-identical canonical surfaces. Please carry the review blocker and parity test into that revision.

—

PR #3365, opened 2026-07-29
fix(hooks): cross-platform process identity for session pointer and lock-owner classification (#3362)

### LINKED ISSUE #3362: [Bug] Session pointer accepts any live process on a recycled PID as owner on darwin/win32 — permanent session_pointer_owner_conflict with no recovery path
## Summary

On **darwin and win32**, session-pointer owner classification trusts *bare PID liveness*. Any live process occupying a recycled PID is accepted as the session owner, so a pointer left behind by a session that died uncleanly (SIGKILL / crash / OOM / reboot) becomes permanently "usable". Every subsequent plain `omx` launch in that checkout then aborts with `session_pointer_owner_conflict`, and the CLI advises setting a distinct `OMX_ROOT` — presenting a manual workaround for what is actually corrupt state. There is no CLI recovery path.

Linux is protected: start-ticks + cmdline-hash comparison detects the recycle and returns `stale-dead`. **darwin/win32 fail open exactly where linux fails closed.**

Sibling of #3261 / #3203 / #3215, but distinct: those all cover a **dead** PID. This is the **live-but-recycled** PID case, which none of them reach.

## Environment

- oh-my-codex 0.20.3 (npm stable); defect also present on current `main` and `dev`
- macOS (darwin arm64), Node.js v24.13.0
- `kern.maxproc: 16000`; observed PID high-water mark **99939** against the ~99999 wrap ceiling — recycling is imminent on a busy machine, not hypothetical

## Reproduction

Classify one identical pointer state, varying only `platform`. The PID belongs to a plain `sleep 300` — unambiguously not a Codex session.

```js
import { isSessionStale, isSessionStateUsable } from 'oh-my-codex/dist/hooks/session.js';

const base = {
  session_id: 'omx-1700000000000-deadbe',
  cwd: CHECKOUT,
  pid: SLEEP_PID,
  started_at: '2026-07-01T00:00:00.000Z',
};

for (const platform of ['darwin', 'win32', 'linux']) {
  const st = { ...base, platform };
  console.log(platform, isSessionStale(st, { platform }), isSessionStateUsable(st, CHECKOUT, { platform }));
}
```

Observed:

| `platform` | `isSessionStale` | `isSessionStateUsable` | |
|---|---|---|---|
| `darwin` | `false` | **`true`** | ← `sleep 300` accepted as session owner |
| `win32`  | `false` | **`true`** | ← same |
| `linux`  | `true`  | `false` | correctly reaped |

A pointer written on darwin carries no identity metadata at all, which is the root enabler:

```
pid 64804 | platform darwin | pid_start_ticks undefined | pid_cmdline undefined
```

## Root cause — five Linux-gated sites

Line numbers from `src/hooks/session.ts` on `main`:

| Line | Function | Gate |
|---|---|---|
| 864 | `sessionIdentityFor` | `return platform === 'linux' ? readLinuxProcessIdentity(pid) : null` — **writer records no identity** off Linux |
| 544 | `defaultReadProcessIdentity` | `if (platform !== 'linux') return { status: 'indeterminate' }` |
| 699 | `classifySessionProcess` | `if (platform !== 'linux') return 'usable'` — **PID-alive becomes owner-alive** |
| 641 | `isSessionStale` | `if (platform !== 'linux') return false` |
| 960 | lock-owner inspection | gated too, but returns `identity-indeterminate` |

Site 864 is the enabler and site 699 is the failure: because the writer never records `pid_start_ticks` / `pid_cmdline` on darwin, there is no recycle-detection material, and the classifier then treats bare liveness as ownership.

Note the internal inconsistency at row 5: the **lock-owner** path already has a `'reused'` status and already fails to `identity-indeterminate` off Linux, which is the safer posture. The **pointer** classifier diverges by returning `'usable'`. Part of this fix is aligning the pointer path to the posture the lock path already takes.

## A test currently pins the defect

`src/hooks/__tests__/session.test.ts` asserts the broken behavior as intended:

```js
it('falls back to PID liveness on non-Linux platforms', () => {
    const state = makeState();
    const stale = isSessionStale(state, {
        platform: 'darwin',
        isPidAlive: () => true,
        readLinuxIdentity: () => null,
    });
    assert.equal(stale, false);
});
```

This is likely why the gap survived repeated hardening of this area — it is green by design. Any fix must replace this case.

## Impact

A checkout becomes permanently unlaunchable via plain `omx`, with no CLI recovery. `omx session lock inspect|recover` targets the lock sidecar, not the pointer's owner classification, so it does not address this state. The only route is manually deleting `<root>/.omx/state/session.json` — which the error message never mentions; it points at `OMX_ROOT` instead.

## Proposed fix — versioned per-platform process identity

Replace PID-only ownership with a platform-neutral, versioned process-birth identity on Linux, Darwin, and Windows. Keep the existing directory lock for pointer mutations.

A session-lifetime OS lease (`flock` / `LockFileEx` held while Codex lives, released by the kernel on death) is the stronger end state, but it is a **separate RFC**, not this fix: the current lock is transaction-scoped and released after each mutation, and many actors rewrite the pointer (session start, native hooks, detached tmux binding, reconciliation, shutdown). A lifetime lease needs a supervisor that survives detached launches and transfers the lease before the outer launcher exits — a lifecycle redesign.

### Provider boundary

One provider used everywhere; no `platform === 'linux'` special case in the classifier.

```ts
type ProcessIdentity =
  | { scheme: 'linux-proc-start-v1';        birth: string; boot_id?: string }
  | { scheme: 'darwin-proc-bsdinfo-v1';     birth: string }
  | { scheme: 'win32-creation-filetime-v1'; birth: string };

type ProcessInspection =
  | { status: 'alive'; identity: ProcessIdentity }
  | { status: 'dead' }
  | { status: 'indeterminate'; reason: string };

inspectProcess(pid: number, platform: NodeJS.Platform): ProcessInspection;
```

- `birth` is an exact decimal **string**, never a JS number.
- Recorded platform must equal the local runtime platform before probing the PID.
- Prefer one combined liveness+identity inspection to close the current `kill(pid, 0)`-then-read-identity race; if two observations disagree because the process exited between calls, retry once, else return indeterminate.
- Command-line hashes are **diagnostic only** — they must not override a matching strong birth identity.

Use it in all five sites above. Fixing only the classifier would leave the stale/usable and lock-owner paths inconsistent.

### On-disk schema

```json
{
  "schema_version": 2,
  "session_id": "...",
  "pid": 1234,
  "platform": "darwin",
  "process_identity": { "scheme": "darwin-proc-bsdinfo-v1", "birth": "1785320123456789" }
}
```

Keep parsing v1 Linux fields, but do **not** overload `pid_start_ticks` with Darwin microseconds or Windows `FILETIME` — different clocks, different units. The lock sidecar already has a version field and should advance in parallel.

The writer should require a strong identity before publishing a new v2 pointer on a supported platform: a failed identity capture should fail the transaction explicitly rather than publish an identity-less pointer that future launches cannot classify. For detached mode, mark the outer-wrapper owner provisional, preserve identity during metadata-only updates, and replace PID + identity together only from native reconciliation that positively resolved the Codex ancestor.

### Darwin identity source

```c
proc_pidinfo(pid, PROC_PIDTBSDINFO, ...)
```
then `pbi_start_tvsec * 1_000_000 + pbi_start_tvusec`, serialized as a decimal string. Both fields are in Apple's public `proc_bsdinfo`.

Both `ps` alternatives should be rejected as the correctness boundary: `lstart` is locale-dependent and discards subsecond data, `etime` is a changing seconds-only duration, and a recycled PID can be assigned within the same displayed second — potentially with the same command line, which can also be mutated. Seconds-resolution start time plus a cmdline hash is a useful emergency mitigation but not a sound identity guarantee.

### Windows identity source

`OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` → `GetProcessTimes` → store creation `FILETIME` as an exact decimal string. Microsoft documents PID reuse and creation date as the way to distinguish reused processes.

A Darwin-first PR under this interface is defensible if native packaging makes all three too large, followed immediately by Windows — but the cross-platform defect should not be described as fixed while win32 still maps every live PID to usable.

### Classifier decision table

| Pointer evidence | Process result | Candidate | Decision |
|---|---|---|---|
| Absent | — | Any | Create v2 pointer |
| v2 | Dead | Any | `stale-dead`; replace atomically |
| v2 | Alive, birth mismatch | Any | `stale-reused`; replace atomically |
| v2 | Alive, birth match | Same verified session | Usable; reconcile/update |
| v2 | Alive, birth match | Different session | Genuine owner conflict; `OMX_ROOT` guidance appropriate |
| v2 | Query denied / error / unsupported | Any | `identity-indeterminate`; preserve and block |
| Legacy, no identity | Dead | Any | `stale-dead`; replace |
| Legacy, no identity | Alive | Verified same session lineage | Permit one guarded in-place upgrade to v2 |
| Legacy, no identity | Alive | New or unverified session | `legacy-indeterminate`; preserve and block |
| Foreign platform/host | Any | Any | Indeterminate; do not probe local PID |
| Unknown future schema | Any | Any | Preserve; report unsupported schema |
| Malformed pointer | Any | Any | Preserve/quarantine only via explicit recovery |

For legacy live pointers neither automatic reap nor blind adoption is safe, since an active old session and an unrelated process holding a recycled PID are observationally indistinguishable. A plain new launch should block with targeted wording — e.g. *"Existing legacy pointer cannot be identity-verified. Inspect or quarantine this pointer before launching a new session."* — and must **not** take the ordinary owner-conflict branch or recommend another `OMX_ROOT`, because live ownership has not been established.

### Test surface

- Table-driven classifier tests covering every row above, through both pointer-read APIs
- Provider tests: alive, dead, permission-denied, malformed response, unsupported platform
- Exact large-value test for Windows `FILETIME`
- Darwin same-second reuse fixture proving a seconds-only representation is insufficient
- Writer tests proving pointer and `owner.json` use the same provider/schema
- Writer failure test proving an identity-less v2 pointer is never published
- Cross-platform test proving a Darwin process never probes a PID recorded as Linux/Windows
- Legacy: dead replacement, verified in-band upgrade, unverified live block, future schema
- CLI regression: recycled PID leads to replacement and successful launch **without** `OMX_ROOT` guidance
- CLI control: a genuinely matching live owner still produces owner conflict **with** `OMX_ROOT` guidance
- Real child-process/SIGKILL integration tests on Linux, macOS, Windows CI
- TOCTOU retry tests where liveness and identity observations disagree
- Performance guard preventing `ps` / PowerShell / WMI subprocesses from entering every native-hook classification
- Replace the existing `falls back to PID liveness on non-Linux platforms` case

## Secondary finding — `EPERM` read as dead

`defaultIsPidAlive` treats **any** `process.kill(pid, 0)` throw as dead, so `EPERM` (live process owned by another user) classifies as dead:

```
pid 1 (/sbin/launchd, alive) -> isSessionStale: true
```

Mirror of the main failure: a live owner gets reaped. `EPERM` should map to alive-or-indeterminate, never dead. Low impact on single-user machines, but it sits in the same probe and the combined-inspection provider above is the natural place to fix it.

Yeachan-Heo:
## Summary

Closes #3362. Replaces the scattered Linux-only process identity logic with one versioned `ProcessIdentity`/`ProcessInspectionProvider` boundary covering Darwin, Windows, and Linux through the existing `omx-runtime` Rust binary.

### Security fix
Deletes the bare-liveness bug at `session.ts` line 699 (`if (platform !== 'linux') return 'usable'`) that trusted PID liveness alone on non-Linux platforms. All platforms now require positive same-lineage birth verification; identity-less live pointers are `identity-indeterminate`.

### Native identity sources
| Platform | API | Birth format |
|---|---|---|
| Darwin | `proc_pidinfo(PROC_PIDTBSDINFO)` | `pbi_start_tvsec.pbi_start_tvusec` (seconds.microseconds since epoch) |
| Windows | `GetProcessTimes` via `OpenProcess` | Creation FILETIME (100ns intervals since 1601) |
| Linux | `/proc/<pid>/stat` field 22 | Jiffies since boot |

### Fail-closed classifier
- **Pointer policy**: birth-mismatch → `stale-dead` (replaceable)
- **Lock policy**: birth-mismatch → `reused` (NOT recoverable; only confirmed-dead grants `safeToRecover`)
- EPERM never maps to dead
- TOCTOU: exactly one retry on birth disagreement, second observation authoritative
- Foreign platform, unsupported/future schema, malformed → `identity-indeterminate`

### V2 schema (v1 legacy-read-only)
- `identity_schema_version: 2` + `process_identity: {platform, birth (decimal string), cmdline_hash?}`
- Writer omits v2 fields entirely when identity capture fails (graceful degradation)
- Team/runtime.ts unified onto shared provider

### Bounded follow-up
Add `omx-runtime` to `NativeProduct`/`NATIVE_HYDRATABLE_PRODUCTS` for full binary packaging on plain `npm install`. Until then, Darwin/Win32 native identity works for development (workspace build) and release builds, but not plain npm install without the binary in PATH.

## Local verification
- ✅ `npm run build` — clean
- ✅ `npm run lint` — clean
- ✅ `npm run check:no-unused` — clean
- ✅ `cargo test -p omx-runtime` — 25 passed
- ✅ Session tests: 121 passed, 0 failed
- ✅ Team runtime tests: 196 passed, 0 failed
- ✅ Launch-fallback tests: 35 passed, 0 failed
- ✅ Terminal P0/P1 review: CLEAR, zero blockers

### Files changed (9 files, +1557 −296)
- `crates/omx-runtime/src/main.rs` — `process-identity <pid>` subcommand
- `crates/omx-runtime/Cargo.toml` — `windows-sys` target-scoped dep
- `crates/omx-runtime/tests/execution.rs` — Rust integration tests
- `Cargo.lock` — updated
- `src/hooks/session.ts` — provider boundary, classifier, writer, schema
- `src/hooks/__tests__/session.test.ts` — full decision-table test suite
- `src/hooks/extensibility/types.ts` — v2 fields on `HookPluginOmxSessionState`
- `src/team/runtime.ts` — unified onto shared provider
- `.github/workflows/ci.yml` — native-rust-darwin/windows CI jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -303,6 +303,44 @@ jobs:
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
 
+  native-rust-darwin:
+    name: Native Rust (Darwin)
+    runs-on: macos-latest
+    timeout-minutes: 30
+    needs: [changes]
+    if: ${{ needs.changes.outputs.full_suite == 'true' || needs.changes.outputs.rust_changed == 'true' || needs.changes.outputs.native_changed == 'true' }}
+    steps:
+      - uses: actions/checkout@v7
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
+      - name: Restore Rust build cache
+        uses: Swatinem/rust-cache@v2
+      - name: Build omx-runtime
+        run: cargo build -p omx-runtime
+      - name: Test omx-runtime
+        run: cargo test -p omx-runtime
+
+  native-rust-windows:
+    name: Native Rust (Windows)
+    runs-on: windows-latest
+    timeout-minutes: 30
+    needs: [changes]
+    if: ${{ needs.changes.outputs.full_suite == 'true' || needs.changes.outputs.rust_changed == 'true' || needs.changes.outputs.native_changed == 'true' }}
+    steps:
+      - uses: actions/checkout@v7
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
+      - name: Restore Rust build cache
+        uses: Swatinem/rust-cache@v2
+      - name: Build omx-runtime
+        run: cargo build -p omx-runtime
+      - name: Test omx-runtime
+        run: cargo test -p omx-runtime
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
@@ -610,7 +648,7 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     timeout-minutes: 10
-    needs: [changes, docs-check, rustfmt, clippy, rust-tests, lint, typecheck, build-dist, ralplan-preflight-macos, test, coverage-team-critical, ralph-persistence-gate, native-cache-integrity, build]
+    needs: [changes, docs-check, rustfmt, clippy, rust-tests, native-rust-darwin, native-rust-windows, lint, typecheck, build-dist, ralplan-preflight-macos, test, coverage-team-critical, ralph-persistence-gate, native-cache-integrity, build]
     steps:
       - name: Check required job results
         shell: bash
@@ -626,6 +664,8 @@ jobs:
           RUSTFMT_RESULT: ${{ needs.rustfmt.result }}
           CLIPPY_RESULT: ${{ needs.clippy.result }}
           RUST_TESTS_RESULT: ${{ needs.rust-tests.result }}
+          NATIVE_RUST_DARWIN_RESULT: ${{ needs.native-rust-darwin.result }}
+          NATIVE_RUST_WINDOWS_RESULT: ${{ needs.native-rust-windows.result }}
           LINT_RESULT: ${{ needs.lint.result }}
           TYPECHECK_RESULT: ${{ needs.typecheck.result }}
           BUILD_DIST_RESULT: ${{ needs.build-dist.result }}
@@ -680,6 +720,8 @@ jobs:
           echo "  rustfmt: $RUSTFMT_RESULT"
           echo "  clippy: $CLIPPY_RESULT"
           echo "  rust-tests: $RUST_TESTS_RESULT"
+          echo "  native-rust-darwin: $NATIVE_RUST_DARWIN_RESULT"
+          echo "  native-rust-windows: $NATIVE_RUST_WINDOWS_RESULT"
           echo "  lint: $LINT_RESULT"
           echo "  typecheck: $TYPECHECK_RESULT"
           echo "  build-dist: $BUILD_DIST_RESULT"
@@ -709,6 +751,8 @@ jobs:
           check_job rustfmt "$RUSTFMT_RESULT" "$rust_active"
           check_job clippy "$CLIPPY_RESULT" "$rust_active"
           check_job rust-tests "$RUST_TESTS_RESULT" "$rust_active"
+          check_job native-rust-darwin "$NATIVE_RUST_DARWIN_RESULT" "$rust_active"
+          check_job native-rust-windows "$NATIVE_RUST_WINDOWS_RESULT" "$rust_active"
           check_job lint "$LINT_RESULT" "$ts_active"
           check_job typecheck "$TYPECHECK_RESULT" "$ts_active"
           check_job build-dist "$BUILD_DIST_RESULT" "$dist_active"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,7 @@ dependencies = [
  "omx-mux",
  "omx-runtime-core",
  "serde_json",
+ "windows-sys",
 ]
 
 [[package]]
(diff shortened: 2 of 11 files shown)


Yeachan-Heo:
## Exact-Head Terminal CI Green — Signed Verdict

**PR:** #3365
**Exact head:** `4f6dd943`
**Base:** `origin/dev` @ `7186ee49`
**CI run:** `30488008051`

### All required checks: ✅ PASS

| Check | Status |
|---|---|
| CI Status (aggregate) | ✅ pass |
| Rust Format | ✅ pass |
| Rust Clippy | ✅ pass |
| Rust Tests + Coverage Signal | ✅ pass |
| **Native Rust (Darwin)** | ✅ pass |
| **Native Rust (Windows)** | ✅ pass |
| Typecheck | ✅ pass |
| Lint | ✅ pass |
| Build dist artifact | ✅ pass |
| Build (Full Source Build) | ✅ pass |
| Test (Node 20 / cli-core-rest) | ✅ pass |
| Test (Node 20 / hooks-notify-platform) | ✅ pass |
| Test (Node 20 / team-state-runtime) | ✅ pass |
| Test (Node 22 / smoke) | ✅ pass |
| Coverage Gate (Team Critical) | ✅ pass |
| Ralph Persistence Gate | ✅ pass |
| Ralplan preflight security (macOS) | ✅ pass |
| Native Cache Integrity (ubuntu) | ✅ pass |
| Native Cache Integrity (macos) | ✅ pass |
| Native Cache Integrity (windows) | ✅ pass |
| Docs Check | ✅ pass |

### P0/P1 Review Summary
- **P0-1**: Line 699 bare-liveness bug — DELETED ✅
- **P0-2**: Darwin uses `proc_pidinfo(PROC_PIDTBSDINFO)` `pbi_start_tvsec`/`pbi_start_tvusec` — NOT Mach absolute ✅
- **P0-3**: Pointer birth-mismatch → `stale-dead`; Lock → `reused` (NOT recoverable) ✅
- **P0-4**: EPERM → `identity-indeterminate`, never dead ✅
- **P0-5**: Unknown schema → `identity-indeterminate` block ✅
- **P1-1**: Windows FILETIME `dwHighDateTime`/`dwLowDateTime` combining ✅
- **P1-2**: RAII `CloseHandle` guard ✅
- **P1-3**: TOCTOU one-retry ✅
- **P1-4**: Writer omits v2 fields on capture failure ✅
- **P1-5**: CI native-rust-darwin/windows jobs ✅
- **P1-6**: Team/runtime.ts unified onto shared provider ✅

### Target-native evidence
- **Darwin**: compiled and tested on `macos-latest` (job `90698565225`, 19s)
- **Windows**: compiled and tested on `windows-latest` (job `90698565185`, 50s)

**Verdict: CLEAR / APPROVED FOR MERGE**

**Signed:** GJC issue-3362 lane (session 019fae5b)

PR #3377, opened 2026-07-31
test(smoke): bind packed cancel fixture and bound commands

Yeachan-Heo:
Fix the v0.20.4 packed-install smoke fixture to execute the exact cancellation probe in its own session root with matching payload identity and explicit root environment. Add a 120s timeout to generic smoke subprocesses so later hangs fail with a bounded diagnostic and teardown.

Focused checks passed: build, no-unused, lint, package dry-run, diff check. The immutable v0.20.4 tag is unchanged.

—


Yeachan-Heo:
Hostile review: exact PR head 112b075b changes only the packed-install smoke fixture and generic subprocess timeout. It does not alter product trust logic, preserves hostile environment regressions, and keeps the v0.20.4 tag immutable. Local build/no-unused/lint/package dry-run passed; exact-head CI is green.

—

PR #3385, opened 2026-07-31
test(smoke): isolate zsh startup probe environment

Yeachan-Heo:
The tagged release reached the documented zsh -f startup-control probe and the packaged hook returned an empty envelope because the smoke child inherited startup environment. Run this exact positive control with BASH_ENV, ENV, and ZDOTDIR explicitly empty; hostile startup probes remain unchanged and fail closed.

—


Yeachan-Heo:
Final hostile exact-head review: PR head 6d24668a changes only the positive zsh -f smoke probe environment, explicitly clearing BASH_ENV/ENV/ZDOTDIR. Hostile startup probes remain unchanged, and the product trust classifier is untouched. Exact-head CI run 30630639939 is terminal green.

—

PR #3389, opened 2026-07-31
fix(smoke): remove inherited env from zsh control probe

Yeachan-Heo:
The packaged zsh positive control still inherited the parent probe environment because undefined/empty overrides were not removed before spawning the hook. Normalize probe env entries so undefined values are deleted, then run the documented zsh -f control with env suppression. Hostile startup tests and product trust logic remain unchanged.

—


Yeachan-Heo:
Final hostile exact-head review for b5d169ffa87d7e3082016ed3b4f0537ff7ff7deb: the branch is reconciled to current dev b4bd6469, and full exact-head CI run 30640030921 is terminal green. The one-file smoke repair removes undefined inherited environment entries before invoking the positive env -u/zsh -f control; hostile startup probes and product trust logic remain unchanged.

—

PR #3393, opened 2026-07-31
test(smoke): remove startup env keys from zsh probe

Yeachan-Heo:
The previous fixture passed empty strings, which still counts as configured startup environment in the hook inspection. The positive zsh -f probe now passes undefined values through the normalized helper so BASH_ENV/ENV/ZDOTDIR are actually absent. Hostile startup probes and product trust logic remain unchanged.

—


Yeachan-Heo:
Final hostile exact-head review for 4f198014e5789d46c7ebc6fb196a82e5a7a069d6: the branch is current against dev e271cc59 and the full exact-head check set is terminal green. The smoke-only change removes BASH_ENV/ENV/ZDOTDIR from the positive zsh -f probe environment via normalized helper handling; hostile startup probes and product trust logic remain unchanged.

—

PR #3396, opened 2026-08-01
fix(runtime): use full Darwin process info buffer

ev78394:
## Summary

The Darwin `process-identity` provider passed `proc_pidinfo(PROC_PIDTBSDINFO)` a truncated, incorrectly laid-out local structure. macOS rejects that undersized buffer with `ENOMEM` (`errno=12`), so v0.20.4 cannot record process identity and fail-closed Stop finalization reports `session_pointer_unusable`.

This follow-up uses libc's maintained Darwin ABI definition and adds target-native regression coverage so the macOS CI job executes the provider rather than only compiling it.

## Changes

- Replace the handwritten partial `ProcTaskBsdInfo` and local FFI declaration with `libc::proc_bsdinfo`, `libc::PROC_PIDTBSDINFO`, and `libc::proc_pidinfo`.
- Require `proc_pidinfo` to return the complete `proc_bsdinfo` byte count before reading the process birth fields.
- Run the current-process identity integration test on Linux, macOS, and Windows, with platform-specific birth-format checks.

## Validation

- [x] `npm run build`
- [ ] `npm test` — attempted on macOS; existing failures in the BSD `script` attached-launch fixture and Linux-oriented session/setup fixtures reproduce on clean `upstream/dev` without this patch.
- [x] `omx doctor` — isolated Codex plugin v0.20.4 setup: 20 passed, 0 warnings
- [x] `npm run lint`
- [x] `npx tsc --noEmit`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p omx-runtime --bin omx-runtime -- -D warnings`
- [x] `cargo test -p omx-runtime` — 19 passed
- [x] Isolated Codex/OMX smoke: `omx exec --skip-git-repo-check -C . "Reply with exactly OMX-EXEC-OK"` returned the exact response, exited 0, and retained no session pointer

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — not needed; the process-identity schema and CLI contract are unchanged
- [x] Backward-compatibility impact considered — this restores the intended v0.20.4 Darwin provider contract

## Related

Follow-up to #3365 and #3362.

### DIFF
--- a/crates/omx-runtime/src/main.rs
+++ b/crates/omx-runtime/src/main.rs
@@ -413,65 +413,21 @@ fn read_linux_cmdline(pid: u32) -> Option<String> {
 // Darwin: proc_pidinfo with PROC_PIDTBSDINFO for pbi_start_tvsec/pbi_start_tvusec
 #[cfg(target_os = "macos")]
 fn process_identity(pid: u32) -> Result<ProcessIdentityResult, ProcessIdentityError> {
-    use std::ffi::c_int;
-
-    extern "C" {
-        fn proc_pidinfo(
-            pid: c_int,
-            flavor: c_int,
-            arg: u64,
-            buffer: *mut std::ffi::c_void,
-            buffersize: c_int,
-        ) -> c_int;
-    }
-
-    const PROC_PIDTBSDINFO: c_int = 3;
-    const PROC_PIDTBSDINFO_SIZE: c_int = std::mem::size_of::<ProcTaskBsdInfo>() as c_int;
-
-    // struct proc_bsdinfo (from libproc.h / proc_info.h):
-    //   pbi_name: [c_char; 16] (COMMLEN = 16)
-    //   pbi_pid: i32
-    //   pbi_ppid: i32
-    //   pbi_uid: uid_t (u32)
-    //   pbi_gid: gid_t (u32)
-    //   pbi_ruid: uid_t
-    //   pbi_rgid: gid_t
-    //   pbi_svuid: uid_t
-    //   pbi_svgid: gid_t
-    //   pbi_rfu: u32
-    //   pbi_flags: u32
-    //   pbi_start_tvsec: u64   ← birth seconds since epoch
-    //   pbi_start_tvusec: u64  ← birth microseconds
-    //   ... remaining fields not needed
-    #[repr(C)]
-    struct ProcTaskBsdInfo {
-        pbi_name: [std::ffi::c_char; 16],
-        pbi_pid: i32,
-        pbi_ppid: i32,
-        pbi_uid: u32,
-        pbi_gid: u32,
-        pbi_ruid: u32,
-        pbi_rgid: u32,
-        pbi_svuid: u32,
-        pbi_svgid: u32,
-        pbi_rfu: u32,
-        pbi_flags: u32,
-        pbi_start_tvsec: u64,
-        pbi_start_tvusec: u64,
-    }
-
-    let mut buf: ProcTaskBsdInfo = unsafe { std::mem::zeroed() };
-    let ret = unsafe {
-        proc_pidinfo(
-            pid as c_int,
-            PROC_PIDTBSDINFO,
+    // Use libc's maintained Darwin ABI definition instead of a partial local
+    // prefix. PROC_PIDTBSDINFO requires a full proc_bsdinfo-sized buffer.
+    let mut process_info: libc::proc_bsdinfo = unsafe { std::mem::zeroed() };
+    let buffer_size = std::mem::size_of::<libc::proc_bsdinfo>() as libc::c_int;
+    let bytes_read = unsafe {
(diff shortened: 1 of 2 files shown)


Yeachan-Heo (approved):
## Adversarial exact-head review — MERGE_READY

**Exact head:** `ab4b2012f8cb6f4e69bed44d78060364973160e0` (verified via `gh pr view` and local `git rev-parse HEAD`).

### Diff scope
Single commit, exactly two files: `crates/omx-runtime/src/main.rs` (-62/+50 net code) and `crates/omx-runtime/tests/execution.rs`. No scope drift; `git diff origin/dev HEAD --stat` confirms nothing else touched.

### Darwin defect — real, correctly diagnosed, correctly fixed
The old handwritten `ProcTaskBsdInfo` had **two compounding ABI bugs**:

1. **Undersized buffer.** Old struct ≈ 72 bytes; `proc_pidinfo(PROC_PIDTBSDINFO)` requires the full `sizeof(proc_bsdinfo)` = 136 bytes. Passing less causes the kernel to fail/return ENOMEM → fail-closed errors, exactly as claimed.
2. **Wrong field order/offsets.** Old layout placed `pbi_start_tvsec` at offset ~56. The real `libc::proc_bsdinfo` places it at ~120 (after `pbi_flags/status/xstatus/pid/ppid/uid/gid/ruid/rgid/svuid/svgid/rfu_1` + `pbi_comm:[c_char;16]` + `pbi_name:[c_char;32]` + `pbi_nfiles/pgid/pjobc/e_tdev/e_tpgid` + `pbi_nice`). Even with a correct buffer size, the old code read comm/name char-array bytes reinterpreted as a u64 — garbage birth timestamps.

The fix replaces the partial handwritten layout with `libc::proc_bsdinfo` (maintained Darwin ABI), eliminating both bugs. Verified against `libc-0.2.187` source: `proc_bsdinfo` field order, `PROC_PIDTBSDINFO = 3`, `MAXCOMLEN = 16`, and the `proc_pidinfo` signature all match.

### Return-size check / error mapping
- `bytes_read <= 0` → errno-mapped (`EPERM|EACCES`→Denied, `ESRCH|ENOENT`→Gone, else Error). Preserved from old code. ✓
- **New** strict `bytes_read != buffer_size` guard → Error. Matches the kernel contract (`proc_pidinfo` returns exactly `sizeof(proc_bsdinfo)` on success for `PROC_PIDTBSDINFO`); cannot falsely reject on any supported macOS. ✓
- Birth format (`"{sec}.{usec}"`) unchanged. ✓

### No behavior weakened
- Linux path (`/proc/{pid}/stat`, field 22 starttime) and Windows path (`GetProcessTimes` FILETIME) untouched.
- `cfg` gates intact for all three platforms; unsupported-platform fallback unchanged.
- `cmdline: None` on Darwin — no new info leak; birth is process start time (same surface as pre-fix).

### Cross-platform regression
Test broadened from `#[cfg(target_os = "linux")]` to `any(linux, macos, windows)` with per-platform birth-format assertions (darwin `sec.usec`, linux/win32 pure decimal). Correct for each platform's `process_identity` output.

### Commands & results (run at exact head)
```
cargo build -p omx-runtime                                    → OK
cargo check -p omx-runtime --target x86_64-apple-darwin       → OK (Darwin path type-checks against libc)
cargo check -p omx-runtime --target x86_64-pc-windows-gnu     → OK (pre-existing dead_code warning only)
cargo clippy -p omx-runtime --all-targets                     → OK
cargo clippy -p omx-runtime --target x86_64-apple-darwin      → OK
cargo fmt -p omx-runtime -- --check                          → EXIT=0
cargo test -p omx-runtime --test execution                   → 25 passed
cargo run -p omx-runtime -- process-identity $$              → {"birth":"94043245","cmdline":"…","platform":"linux"}
```

### CI verified green on exact head `ab4b2012`
Run `30678302926` (CI) + `30678302898` (PR Check) — both `success`. Key lanes:
- **Native Rust (Darwin)** ✓ — `process_identity_current_process_returns_identity ... ok` (job 91309978053, 19 passed). This is the test that actually exercises the new `libc::proc_bsdinfo` code path on macOS.
- **Native Rust (Windows)** ✓ — `process_identity_current_process_returns_identity ... ok` (job 91309978095, 19 passed).
- **Rust Tests + Coverage Signal** ✓, **Rust Clippy** ✓, **Rust Format** ✓.
- **Native Cache Integrity** (ubuntu/macos/windows) ✓ — no generated/native-cache risk.

### Verdict
**MERGE_READY.** Real ABI defect (undersized buffer + wrong field offsets → garbage birth / fail-closed errors), minimal correct fix using the maintained `libc::proc_bsdinfo` definition, strict return-size guard matching the kernel contract, no behavior weakened, no scope drift, no security/privacy regression, green CI on all three target OSes including the Darwin test that exercises the fixed path.

—

PR #3421, opened 2026-08-02
fix(hooks): silence identity-indeterminate Stop

### LINKED ISSUE #3420: [Bug] identity-indeterminate Stop rejection can loop in Codex App
## Summary

In Codex App with oh-my-codex 0.20.4, a `Stop` hook can repeatedly inject the same prompt when the selected session pointer is `identity-indeterminate`:

```text
OMX cannot authorize Stop while the selected session pointer is identity-indeterminate; repair the pointer evidence before continuing.
```

The assistant can complete the requested work, but every later Stop reaches the same rejection and the conversation does not terminate normally.

This appears to be a remaining unusable-pointer path after the bounded Stop fixes for `unmatched session id` (#3253 / #3254) and `foreign-cwd` (#3288 / #3238). Those cases now return neutral output, while `identity-indeterminate` still emits a remediation prompt that Codex App may treat as another turn.

## Environment

- OS: macOS 26.5.2 (Darwin arm64)
- Node.js version: v22.22.3
- `oh-my-codex` version: v0.20.4
- Codex CLI version: 0.146.0
- Surface: ChatGPT/Codex desktop app

## Steps to reproduce

1. Use OMX hooks in Codex App.
2. Reach a state where the selected `session.json` pointer is classified as `identity-indeterminate`.
3. Complete an ordinary assistant turn so the native `Stop` hook runs.
4. Observe the hook prompt above.
5. Let the assistant respond/finish again.
6. Observe the same Stop prompt being injected again.

### Application context

This was observed while developing and testing a local orchestration app named `divvy` inside Codex App. Given one incoming task, `divvy` assigns work between two external runners, Claude Code and Codex CLI. For the Codex-assigned portion it launches a separate `codex exec` child process, waits for that runner to finish, and then returns its result to the parent Codex App conversation for synthesis.

During repeated tests of that flow, the Stop loop appeared at the parent/child transition: the `codex exec` child had finished, control and results had returned to the parent conversation, and the parent assistant then attempted to finish its turn. At that point the native Stop hook reported the selected pointer as `identity-indeterminate`. Subsequent attempts to finish received the same injected prompt again.

This context may help reproduce the session transition, but it does not establish that `divvy` or `codex exec` caused the pointer state. The confirmed signal at the hook boundary is that OMX selected a session pointer classified as `identity-indeterminate` after control returned to the parent conversation.

## Expected behavior

An indeterminate pointer should remain fail-closed for session or workflow side effects, but the Stop path should be bounded and terminal. It should not turn the authorization diagnostic into an indefinitely repeated assistant prompt.

Possible safe outcomes include neutral Stop output for this condition, one-time deduplication, or an operator-facing recovery path that does not create another model turn.

## Actual behavior

- Stop is rejected because the pointer is `identity-indeterminate`.
- The hook asks the assistant to repair pointer evidence.
- Codex App invokes another assistant turn.
- The next Stop receives the same rejection.
- The sequence can repeat until the user interrupts or disables/patches the hook.

## Logs / screenshots

Sanitized hook payload:

```xml
<hook_prompt hook_run_id="stop:7:<user-config>/hooks.json">OMX cannot authorize Stop while the selected session pointer is identity-indeterminate; repair the pointer evidence before continuing.</hook_prompt>
```

Local diagnostic confirmation:

- A narrow local change that returns neutral output only for `identity-indeterminate` Stop pointers stops the replay.
- Existing bounded behavior for malformed pointers can remain unchanged.
- This is currently used only as a local hotfix; it is not proposed as the authoritative upstream fix because the intended recovery/security contract should be confirmed by maintainers.

## Relationship to existing reports

- #3253 / #3254 fixed repeated Stop output for unmatched session IDs.
- #3288 documents the same feedback-loop shape for `foreign-cwd`; maintainers confirmed current `dev` deliberately no-ops that Stop condition via #3238.
- This report covers the distinct `identity-indeterminate` branch on released 0.20.4.

Requested remediation: define a bounded Stop contract for `identity-indeterminate` pointers and add a regression test asserting that repeated dispatch does not emit another remediation prompt or mutate native Stop state.

Yeachan-Heo:
## Summary
- neutralize the first Stop response for `identity-indeterminate` session pointers so the diagnostic cannot become a repeated Codex App model turn
- preserve fail-closed behavior and leave malformed-pointer and authorization branches unchanged
- assert repeated Stop dispatch leaves native Stop, Ralph, and active-skill state untouched

## Verification
- `npm run build`
- `node dist/scripts/run-test-files.js dist/scripts/__tests__/codex-native-hook.test.js` (641 passed)
- `npm run check:no-unused`
- `npm run lint`
- `git diff --check`

Resolves #3420

### DIFF
--- a/src/scripts/__tests__/codex-native-hook.test.ts
+++ b/src/scripts/__tests__/codex-native-hook.test.ts
@@ -30029,6 +30029,11 @@ PY`,
           },
         },
       });
+      const ralphStatePath = join(stateDir, "sessions", "sess-dead", "ralph-state.json");
+      const skillStatePath = join(stateDir, "skill-active-state.json");
+      const ralphStateBefore = await readFile(ralphStatePath, "utf-8");
+      const skillStateBefore = await readFile(skillStatePath, "utf-8");
+
       const nativeStopStateBefore = await readFile(nativeStopStatePath, "utf-8");
       const payload = {
         hook_event_name: "Stop" as const,
@@ -30047,11 +30052,12 @@ PY`,
       );
 
       assert.equal(first.omxEventName, "stop");
-      assert.equal(first.outputJson?.decision, "block");
-      assert.equal(first.outputJson?.stopReason, "session_pointer_unusable");
+      assert.equal(first.outputJson, null);
       assert.equal(replay.omxEventName, "stop");
       assert.equal(replay.outputJson, null);
       assert.equal(await readFile(nativeStopStatePath, "utf-8"), nativeStopStateBefore);
+      assert.equal(await readFile(ralphStatePath, "utf-8"), ralphStateBefore);
+      assert.equal(await readFile(skillStatePath, "utf-8"), skillStateBefore);
     } finally {
       await rm(cwd, { recursive: true, force: true });
     }
--- a/src/scripts/codex-native-hook.ts
+++ b/src/scripts/codex-native-hook.ts
@@ -23225,10 +23225,11 @@ export async function dispatchCodexNativeHook(
       const stopHookActive = payload.stop_hook_active === true || payload.stopHookActive === true;
       const pointerCannotAuthorizeThisCwd = pointer.status === "foreign-cwd";
       const unmatchedStopSession = failure.stopReason === "session_scope_unmatched";
-      // An unmatched payload or foreign-cwd pointer cannot authorize continuation.
-      // Return no output so Codex cannot reinterpret an authorization diagnostic as
-      // a new action request. Other unusable pointers retain their bounded replay behavior.
-      if (pointerCannotAuthorizeThisCwd || unmatchedStopSession || stopHookActive) {
+      // An unmatched payload, foreign-cwd pointer, or identity-indeterminate pointer
+      // cannot authorize continuation. Return no output so Codex cannot reinterpret an
+      // authorization diagnostic as a new action request. Other unusable pointers retain
+      // their bounded replay behavior.
+      if (pointerCannotAuthorizeThisCwd || pointer.status === "identity-indeterminate" || unmatchedStopSession || stopHookActive) {
         outputJson = null;
       } else {
         outputJson = {


Yeachan-Heo (commented):
Verdict: MERGE_READY

Adversarial red-team review completed against current head `eb934817` and its `dev` parent.

- The production diff adds exactly one new predicate: `pointer.status === "identity-indeterminate"` in the existing Stop neutralization condition. No other Stop authorization path changed.
- Baseline behavior was verified from `eb934817^`: identity-indeterminate reached the first generic `session_pointer_unusable` block, while malformed pointers retained the same first-block/replay-neutralization contract.
- The regression test now proves identity-indeterminate first and replay Stop calls are both neutral and that `native-stop-state.json`, Ralph state, and active-skill state remain byte-for-byte unchanged.
- Existing malformed-pointer coverage still requires the first block and replay neutralization. Existing unmatched-session and foreign-CWD coverage remains present and passed.
- Targeted adversarial run passed: 22/22 tests covering identity-indeterminate, malformed, foreign-CWD, unmatched, and related worker-pointer boundaries.
- PR CI is green for Typecheck and Node 22 smoke; skipped jobs are platform/workflow-gated rather than failures.

No scope creep or fail-closed authorization regression was found.
—

Yeachan-Heo (commented):
Verdict: MERGE_READY

Fresh adversarial red-team review for refreshed head `e823b155` on base `234bb802`. The rebase introduced no product-scope changes: the diff remains limited to the identity-indeterminate Stop neutralization predicate and its state-preservation regression assertions.

- Malformed pointers retain first-block/replay-neutralization behavior.
- Unmatched-session and foreign-CWD Stop behavior remains neutral and fail-closed.
- No native Stop, Ralph, or active-skill state mutation occurs on the indeterminate branch.
- Authorization boundaries remain fail-closed; only the non-authorizing identity-indeterminate Stop response is newly neutralized.
- Refreshed targeted adversarial run passed 22/22; build, no-unused, lint, and diff checks passed.
- Refreshed PR CI is green and GitHub reports `CLEAN` against `dev` at `234bb802`.

No scope creep or contract regression found.
—

PR #3434, opened 2026-08-04
fix(team): explain foreign tmux pane topology

### LINKED ISSUE #3433: team: Orca shared tmux window triggers topology changed before layout mutation
# Bug: OMX Team aborts in Orca when tmux topology changes before layout mutation

## Summary

OMX Team fails reproducibly when launched from Orca's embedded terminal with:

```text
tmux window topology changed before layout mutation
```

The same project and Team workflow work from a standalone Ghostty tmux session.

## Environment

- oh-my-codex: `0.20.3`
- Orca: `1.4.167` (`com.stablyai.orca`)
- Codex standalone: `0.146.0-aarch64-apple-darwin`
- tmux: `3.6b`
- macOS: `26.5` (`25F71`)

## Reproduction

1. Open a worktree terminal inside Orca.
2. Run Codex/OMX inside its attached tmux leader pane.
3. Start an OMX Team workflow with one or more workers.
4. Team startup aborts before layout mutation with the topology-change error.
5. Run the equivalent workflow from a standalone Ghostty tmux session; Team starts normally.

## Actual behavior

`createTeamSession()` aborts because `requireFrozenWindowTopologySync()` observes a pane set different from its expected pane-ID/PID map. In the installed build, this check is called before `select-layout`, window sizing, and HUD recreation. Orca appears either to expose an existing non-Team pane in the same tmux window or to alter pane topology while Team startup is proving ownership.

The initial expected map is constructed from only the selected leader pane plus HUD panes attributed to that leader. Therefore, a stable non-HUD neighbor pane already present in Orca's shared tmux window is sufficient to trigger the same error; an actual timing race is not required.

This appears to be the intended safety guard working as written, but the integration leaves Team unusable in this environment and the current error does not identify the unexpected/missing pane.

## Expected behavior

One of the following would make Team usable and diagnosable:

1. Create/use an isolated Team-owned tmux window when the host terminal has foreign panes.
2. Treat stable host-owned panes as preserved topology rather than immediately rejecting them, while retaining ownership checks for mutations.
3. Detect unsupported embedded-terminal topology before partial startup and emit expected versus actual pane IDs, ownership tags, and recommended remediation.
4. Document Orca/embedded-terminal constraints if this topology is intentionally unsupported.

## Relevant implementation evidence

The installed `dist/team/tmux-session.js` throws the exact message in `requireFrozenWindowTopologySync()` when either pane-set cardinality or pane identity differs. `createTeamSession()` invokes the check repeatedly around initial HUD discovery, worker creation, layout selection, width mutation, and HUD creation. The existing test named `rejects an unowned pane in startup topology before any window-wide mutation` deliberately reproduces the initial shared-window/neighbor-pane case and expects this exact error.

## Privacy

No project paths, datasets, credentials, or session identifiers are included in this report.

Yeachan-Heo:
## Summary
- preserve the existing fail-closed frozen-window topology guard
- report the target plus expected, actual, unexpected, and missing pane IDs
- give actionable isolated-window remediation without classifying foreign panes as Team-owned
- strengthen the intentional unowned-neighbor regression to prove diagnostics and zero partial tmux mutation

## Reproduction
Current `origin/dev` reproduces issue #3433 with the existing `rejects an unowned pane in startup topology before any window-wide mutation` fixture: the shared window contains `%1,%2,%7,%8`, while Team owns only `%1,%2`, and startup rejects before mutation.

## Safety boundaries
- foreign panes remain host-owned
- worker/HUD ownership proofs are unchanged
- `requireFrozenWindowTopologySync()` call ordering is unchanged
- standalone tmux and existing unowned/mutating topology protections remain intact
- no broad arbitrary-topology acceptance

## Verification
- `npm run build`
- focused foreign-neighbor and later-owner-check regressions: 2 passed
- relevant tmux/guard/fixture suites: 327 passed across the selected files (290 + 19 + 18)
- full `dist/team/__tests__` suite: all 40 files passed
- `npm run lint`
- `npm run check:no-unused`
- `git diff --check`

Closes #3433

Signed-off-by: GJC <gjc@openai.com>

### DIFF
--- a/src/team/__tests__/tmux-session.test.ts
+++ b/src/team/__tests__/tmux-session.test.ts
@@ -5660,7 +5660,17 @@ esac
 
           assert.throws(
             () => createTeamSession('Owned HUD Startup', 1, cwd),
-            /tmux window topology changed before layout mutation/,
+            (error: unknown) => {
+              assert.ok(error instanceof Error);
+              assert.match(error.message, /tmux window topology changed before layout mutation/);
+              assert.match(error.message, /target=shared:0/);
+              assert.match(error.message, /expected=\[%1,%2\]/);
+              assert.match(error.message, /actual=\[%1,%2,%7,%8\]/);
+              assert.match(error.message, /unexpected=\[%7,%8\]/);
+              assert.match(error.message, /missing=\[\]/);
+              assert.match(error.message, /isolated tmux window with no host-owned foreign panes/);
+              return true;
+            },
           );
 
           const tmuxLog = await readFile(logPath, 'utf-8');
--- a/src/team/tmux-session.ts
+++ b/src/team/tmux-session.ts
@@ -775,6 +775,20 @@ function requireLiveTeamOwnedPaneSync(
   return requireLiveExactPaneSync(target, expectedPanePid);
 }
 
+function frozenWindowTopologyError(
+  teamTarget: string,
+  expectedPaneIds: ReadonlySet<string>,
+  actualPaneIds: ReadonlySet<string>,
+): Error {
+  const expected = [...expectedPaneIds].sort();
+  const actual = [...actualPaneIds].sort();
+  const unexpected = actual.filter((paneId) => !expectedPaneIds.has(paneId));
+  const missing = expected.filter((paneId) => !actualPaneIds.has(paneId));
+  return new Error(
+    `tmux window topology changed before layout mutation: target=${teamTarget}; expected=[${expected.join(',')}]; actual=[${actual.join(',')}]; unexpected=[${unexpected.join(',')}]; missing=[${missing.join(',')}]. Team requires an isolated tmux window with no host-owned foreign panes`,
+  );
+}
+
 function requireFrozenWindowTopologySync(
   teamTarget: string,
   expectedPanePids: ReadonlyMap<string, number>,
@@ -786,7 +800,7 @@ function requireFrozenWindowTopologySync(
   const actualPaneIds = new Set(topology.panes.map((pane) => pane.paneId));
   if (actualPaneIds.size !== expectedPanePids.size
     || [...expectedPanePids.keys()].some((paneId) => !actualPaneIds.has(paneId))) {
-    throw new Error('tmux window topology changed before layout mutation');
+    throw frozenWindowTopologyError(teamTarget, new Set(expectedPanePids.keys()), actualPaneIds);
   }
 
   for (const [paneId, expectedPanePid] of expectedPanePids) {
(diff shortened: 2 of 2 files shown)


Yeachan-Heo:
## Adversarial terminal review

Reviewed pushed commit `fc538b933dd1d0febd124da6327e6ad0deb7a50e` against issue #3433 and the locked ownership constraints.

### Attack matrix
- **Foreign neighbor present at startup:** rejected before any tmux mutation; regression proves no `set-option`, `kill-pane`, `split-window`, resize, layout, selection, or input effect.
- **Pane appears during later ownership proof:** existing race regression remains passing; both topology mismatch sites now emit the same bounded identity evidence.
- **Missing/replaced expected pane:** missing IDs are derived separately from unexpected IDs; PID and owner continuity checks remain unchanged and fail closed.
- **Attempted ownership broadening:** no pane set, owner tag, worker/HUD registry, or authorization predicate changed. Foreign panes remain host-owned.
- **Standalone tmux regression:** startup behavior is unchanged when actual and expected sets match; the full Team suite and selected tmux guard/fixture suites pass.
- **Information exposure:** diagnostic output is limited to tmux target and pane IDs; it does not dump commands, paths, environment, captures, or credentials.
- **Scope inflation:** two files changed; no isolated-window lifecycle, arbitrary-topology acceptance, or Team redesign was introduced.

### Verification reviewed
- focused reproduction/regressions: 2/2 passed
- selected tmux/guard/fixture suites: 327 passed
- full Team suite: all 40 test files passed
- build, lint, no-unused, and diff checks passed
- required GitHub checks completed without outstanding or failed checks
- PR targets `dev`, is mergeable, and the pushed HEAD matches the reviewed commit

**Verdict: MERGE_READY**

Signed-off-by: GJC Adversarial Review <gjc@openai.com>

Yeachan-Heo:
## Post-merge packed-smoke attribution

The blocker is **pre-existing and unrelated to #3434**.

Evidence:
- #3434 changes only `src/team/tmux-session.ts` and its test; `src/hooks` and `src/scripts` are tree-identical to merge parent `d721fe5e`.
- An isolated packed install built from `d721fe5ecd6a79378949eff6f30d327c692ec68b` fails at the identical `directive-documentation-then-command` case.
- The hook correctly selects `autopilot`, then intentionally persists `active:false`, `phase:"failed"`, `error:"documented_host_consensus_receipt_unavailable"`; the packed smoke still requires `active:true`.
- The stale assertion predates the fail-closed activation change in `aba82f18`.

Owner issue: #3435. Smallest scope is the packed-smoke case model and its helper contract test; classifier and fail-closed runtime behavior remain unchanged. The owner lane has been recovered from current `dev` for a dev-targeted fix PR.

No release workflow was rerun or cancelled, and no merge action was taken.

Signed-off-by: GJC Post-Merge Recovery <gjc@openai.com>

PR #3191, opened 2026-07-16
fix: tolerate Windows regular-file fsync EPERM

Yeachan-Heo:
## Summary
- tolerate only native Windows regular-file `fsync` failures whose structured code is exactly `EPERM`
- preserve all POSIX, non-`EPERM`, directory-sync, write, close, rename, rollback, claim-journal, and recovery failures
- aggregate degraded durability per transaction and emit exactly one post-commit stderr warning naming Windows EPERM, subsystem, and degraded durability
- cover session start/reconcile/end, setup, uninstall, doctor recovery, and claim-journal paths with scoped deterministic seams

## Verification
- `npx tsc --noEmit`
- `npm run check:no-unused`
- `npm run lint`
- `npm run build`
- focused 7-file regression suite: utility, session, claim journal, setup, uninstall, doctor, native SessionStart
- `npm run test:ci:compiled`

Fixes #3189

—

_[repo owner's gaebal-gajae (clawdbot) 🦞]_

Issue #3189, opened 2026-07-16
[Bug] v0.20.2 fails to start on native Windows when session-pointer fsync returns EPERM


SeongHo5:
## Summary

Regression in `oh-my-codex` 0.20.2 on native Windows: session-pointer and native-hook transactions now fail because `fsync()` returns `EPERM`.

This is not caused by a custom OMX state-root configuration or directory permissions:

- `OMX_ROOT`, `OMX_STATE_ROOT`, and `OMX_TEAM_STATE_ROOT` are all unset.
- The project `.omx\state` directory is writable.
- `fsync()` fails with the same `EPERM` on both an OMX state file and an unrelated regular file in the globally installed package.

The regression appears to be introduced by the 0.20.2 canonical session-pointer work (PR #3158). In 0.20.1, `src/hooks/session.ts` used plain `writeFile()` for `session.json`; 0.20.2 added `openAndSync()` / `handle.sync()` to the session-pointer transaction and treats an `EPERM` as a fatal launch failure.

## Environment

- OS: Windows 10 Pro 25H2, build 26200.8875 (native PowerShell)
- Node.js version: v24.14.1
- `oh-my-codex` version: 0.20.2 (upgraded from 0.20.1)
- Codex CLI version: codex-cli 0.144.4

## Steps to reproduce

1. On native Windows, install or update `oh-my-codex` from `0.20.1` to `0.20.2`.
2. From a normal project directory, run `omx` or trigger the native-hook refresh path.
3. Observe that OMX aborts while publishing the session-pointer lock owner.
4. The underlying failure can also be reproduced directly with Node:

```powershell
   node -e "const fs=require('fs/promises'); (async()=>{const h=await fs.open('.omx/state/native-stop-state.json','r'); try { await h.sync(); console.log('OK') } finally { await h.close() }})().catch(e=>{console.error(e); process.exit(1)})"
 ```

5. On this machine, the command prints `EPERM: operation not permitted, fsync`.

## Expected behavior

`omx` should start successfully on native Windows.

The session-pointer and native-hook transaction code should preserve atomic writes where supported, but treat Windows `EPERM` from `fsync()` as a platform capability limitation rather than aborting OMX startup and hook setup.

## Actual behavior

Both hook setup and normal OMX launch fail:

- Native hook transactions roll back with `EPERM: operation not permitted, fsync`.
- `omx` aborts with `session_pointer_io_failure`.
[...]

PR #3443, opened 2026-08-05
fix(setup): tolerate unsupported Windows directory fsync

KROU4:
## Summary

Fix native Windows setup failures in the native-hook transaction by:

- treating only Windows directory `fsync()` failures with `code === "EPERM"` as an unsupported durability capability
- aggregating regular-file and directory degradation into one accurate post-commit warning
- avoiding cross-open Windows `dev`/`ino`/mode equality as a transaction ownership token when the controlled path, single-link regular-file topology, and exact bytes still match
- applying the shared directory durability outcome to setup, uninstall, doctor recovery, and claim-journal paths

## Environment

Reproduced on native Windows 25H2 with:

- Node.js 24.19.0
- Codex CLI 0.146.0
- oh-my-codex 0.20.4
- PowerShell 5.1

## Reproduction

Run:

```powershell
omx setup --force --merge-agents
```

Before this change, the transaction rejects hash-identical temporary files and then aborts on directory synchronization:

```text
Native hook transaction read-back changed for native hook Windows shim ... temporary;
refusing to overwrite concurrent content.

Native hook transaction read-back changed for native hooks ... temporary;
refusing to overwrite concurrent content.

Native hook transaction read-back changed for notification metadata ... temporary;
refusing to overwrite concurrent content.

EPERM: operation not permitted, fsync
```

## Root cause

PR #3191 added a deliberately narrow fallback for Windows regular-file `fsync()` returning `EPERM`, but native-hook claim-journal directory synchronization still called `handle.sync()` directly and treated the same Windows capability failure as fatal.

The transaction also compared cross-open `dev`, `ino`, link-count, and Unix-mode snapshot metadata in addition to bytes. On Windows, file-index-derived identity and synthesized mode metadata are not reliable cross-open ownership tokens. That caused false rejection even when the exclusive controlled path remained a single-link regular file with exactly the expected SHA-256/content.

## Safety boundary

The durability fallback remains fail-closed except for the exact pair:

```text
process.platform === "win32" && error.code === "EPERM"
```

All non-Windows failures and all non-`EPERM` failures are rethrown.

Windows transaction validation still requires:

- controlled paths and safe ancestor topology
- regular files rather than symlinks
- exactly one link for transaction artifacts
- exact expected bytes
- capture-time identity stability
- claim-journal ownership, digest, and link-topology checks
- no-clobber copy/link behavior

Actual byte mutation, symlinks, unexpected hard links, path escape, stale content, and foreign content remain fatal. POSIX retains strict cross-open device/inode/mode equality.

## After

Setup completes and emits one diagnostic when durability is degraded:

```text
[omx] warning: Windows EPERM directory fsync unsupported in native-hook setup; operation succeeded with degraded durability.

Setup complete! Run "omx doctor" to verify installation.
```

If both regular-file and directory synchronization are unsupported, the single warning names both.

## Tests

Added or extended regression coverage for:

- Windows directory `EPERM` succeeds and emits one warning
- Windows directory non-`EPERM` fails
- POSIX directory `EPERM` fails
- hash-identical Windows temporary read-back with different file identity succeeds
- actual byte mutation fails closed
- symlinked ancestry, unexpected hard links, and path escape fail closed
- shim, `hooks.json`, notification metadata, and `config.toml` commit successfully
- `doctor` recognizes the committed config and native hooks
- claim-journal recovery keeps exact byte/link safety checks

Verification:

```text
npm run build
npm run lint
npm run check:no-unused
node dist/scripts/run-test-files.js dist/utils/__tests__/file-durability.test.js dist/cli/__tests__/native-hook-claim-journal.test.js
focused setup-install-mode Windows durability/read-back/safety tests
focused doctor setup-recognition and claim-journal recovery tests
```

## Relationship to #3189 / #3191

This is a narrow follow-up to #3189 and the regular-file fix in #3191. It extends the same exact Windows-`EPERM` capability boundary to transactional directory synchronization and preserves the original rule that every other filesystem failure remains fatal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,13 @@ omx wiki refresh --json
 Native Windows remains a secondary path, and WSL2 is generally the better choice if you want a Windows-hosted setup.
 On native Windows, OMX accepts `psmux` as the tmux-compatible binary for the existing tmux-backed paths it already uses.
 
+Native Windows filesystems may report `EPERM` when Node.js requests `fsync()`
+on an otherwise valid regular-file or directory handle. OMX treats only that
+exact Windows capability failure as degraded durability after a successful
+native-hook transaction and emits one warning. Non-Windows errors, other error
+codes, unsafe paths or links, concurrent content changes, and transaction
+validation failures remain fatal.
+
 | Platform | Install |
 | --- | --- |
 | macOS | `brew install tmux` |
--- a/src/cli/__tests__/doctor-warning-copy.test.ts
+++ b/src/cli/__tests__/doctor-warning-copy.test.ts
@@ -2839,7 +2839,7 @@ command = "node"
 							throw Object.assign(new Error("injected Windows EPERM"), { code: "EPERM" });
 						},
 					}, "win32"),
-					syncDirectory: async () => undefined,
+					syncDirectory: async () => "synced",
 				});
 				process.stderr.write = ((chunk: string | Uint8Array) => {
 					stderr.push(String(chunk));
@@ -2871,7 +2871,7 @@ command = "node"
 			const restoreDurability = setDoctorClaimJournalDurabilityForTest({
 				platform: "win32",
 				syncRegularFile: async () => { throw regularError; },
-				syncDirectory: async () => undefined,
+				syncDirectory: async () => "synced",
 			});
 			try {
 				await assert.rejects(doctor(), (error) => error === regularError);
--- a/src/cli/__tests__/native-hook-claim-journal.test.ts
+++ b/src/cli/__tests__/native-hook-claim-journal.test.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { mkdtemp, readFile, rename, rm, writeFile } from "node:fs/promises";
+import { link, mkdtemp, readFile, rename, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { test } from "node:test";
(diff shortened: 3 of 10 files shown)


Yeachan-Heo:
REQUEST_CHANGES — closing this PR.

The red-team review reproduced a PR-introduced rollback-safety regression: on the `dev` base, the setup suite passes 121/121; at this exact head, the suite gains failures and the Windows cross-open identity relaxation accepts a same-byte replacement with changed file identity during rollback. That removes an existing TOCTOU/concurrent-replacement detection boundary rather than only accommodating an unsupported Windows durability capability.

The directory-`EPERM` test also does not presently establish the intended native-Windows filesystem contract in the review environment. Keep the repair narrowly focused: preserve detection of post-write/post-rename replacement during rollback while separately defining platform-valid Windows identity evidence. Do not relax replacement ownership merely because bytes match.

The GJC review pane terminated after producing this evidence, so this signed disposition records the bounded terminal finding. Please submit a successor only with the rollback race fixed and exact-head coverage for the retained safety boundary.

—

KROU4:
Successor draft: https://github.com/Yeachan-Heo/oh-my-codex/pull/3445

The rejected Windows bytes-only identity relaxation is removed. The successor tolerates only synthesized mode drift while retaining bytes/dev/ino/nlink ownership checks, separates the directory-EPERM durability fallback, and adds exact regressions for same-byte replacement, junction escape, hardlink races, copy-fallback identity substitution, and post-open journal/artifact replacement.

PR #3476, opened 2026-08-10
fix(smoke): bind packed zsh probe to built runtime

Yeachan-Heo:
## Root cause
The packed-install actor probes intentionally removed `OMX_RUNTIME_BINARY`, so the main-root `zsh -f -c ':'` startup control could fall back to stale packaged/ambient runtime discovery instead of the exact runtime already built for the release smoke. On the tagged release runner that produced a non-empty fail-closed hook response and tripped the packed smoke assertion.

## Fix
- thread the already-validated `runtimeBinary` into `smokeInstalledNativeHookDist`
- scope `OMX_RUNTIME_BINARY` to the zsh fast-startup actor probe only
- retain the existing runtime-absent behavior for unrelated packed hook probes

## Verification
- `npm run build`
- `npm exec --yes --package=@openai/codex@0.142.5 -- sh -c 'npm run smoke:packed-install'` — PASS
- `git diff --check`

The existing `v0.20.5` tag is not mutated and npm publication remains blocked pending the orchestrated release recovery.

—


Yeachan-Heo:
MERGE_READY

Adversarial review completed against exact head `2494d792ca602d1963fcd1b0f047e582935a6a3b` and base `f585988b6d875b36a22edf7ebf6537a248431250`.

Evidence:
- Exact diff is limited to `src/scripts/smoke-packed-install.ts` (+3/-3).
- `runtimeBinary` originates from `resolvePackedSmokeRuntimeBinary(repoRoot)` after `npm run build:runtime`; validation requires an absolute executable regular file (`.exe` on Windows).
- The validated path is threaded synchronously into `smokeInstalledNativeHookDist`; there is no async yield between resolution and this probe sequence.
- Authority is scoped to only the main-root `zsh -f -c ':'` control through the per-probe inherited environment. The shared `hookEnv` still uses `{ runtimeBinary: null }`, so Team, collaboration, metadata-control, child, and all unrelated actor probes retain runtime-absent behavior.
- Ambient `OMX_*`, `CODEX_*`, `GJC_*`, shell-startup, and tmux variables remain scrubbed by `buildPackedProbeEnv`; the change does not broaden inherited environment authority.
- Missing, non-file, relative, or non-executable runtime paths fail validation. If the validated runtime becomes unavailable before dispatch, the exact zsh control cannot silently pass; the hook emits a non-empty/failing result and the smoke remains fail-closed.
- Portability is preserved through `runtimeBinaryName(process.platform)`, platform-specific executable validation, and ordinary environment transport.
- The packed artifact still runs the installed `dist/scripts/codex-native-hook.js`; only the explicitly built workspace runtime is supplied to the one regression that requires exact runtime authority.

Fresh verification:
- `npm run build` — passed.
- `node --test dist/scripts/__tests__/smoke-packed-install.test.js` — 56 passed, 0 failed.
- Exact pinned reproduction with `@openai/codex@0.142.5`: `npm run smoke:packed-install` — PASS.
- `git diff --check` — passed.
- Hosted exact-head `CI Status` — success; REST `mergeable_state` — clean.
- GitHub reviews, issue comments, and inline review comments were empty before this terminal review; no automated review blocker required reconciliation.

No correctness, trust-boundary, portability, artifact, or release-risk blocker found. Do not mutate the existing tag or publish until the owner-controlled release recovery proceeds after merge.

—

PR #3531, opened 2026-08-16
fix(team): Claude worker submits use tmux named Enter; detect Claude 2.1.x trust/bypass prompts

### LINKED ISSUE #3529: [Bug] Team Claude workers never start: OMX submits trigger with `C-m`, but Claude Code 2.1.x only submits on named `Enter` (also: trust/bypass prompt wording drift)
## Summary

On `omx team` with a Claude worker (`OMX_TEAM_WORKER_CLI_MAP=codex,claude`), the worker pane launches, is trusted, receives the typed trigger draft — and then sits idle forever. The task stays `pending`. OMX's dispatch/verification path types the trigger and presses Enter via `C-m`, but **Claude Code 2.1.233's TUI does not treat `C-m` as submit; only the tmux named key `Enter` submits.** So the draft is never sent.

This is the same class as #391 ("worker draft is unsent, manual C-m required"), but the root cause on the current Claude CLI is the opposite key: even `C-m` no longer works — it must be `Enter`.

There are two secondary wording-drift issues in the same area (trust prompt + bypass-permissions prompt) that make the readiness/auto-trust logic mis-time or miss, described at the end.

## Environment

- OS: WSL2 (Ubuntu), kernel 6.6-microsoft-standard-WSL2
- Node.js: v22.22.1
- oh-my-codex: 0.20.5 (dev branch, includes #3489)
- Codex CLI: 0.147.0
- Claude Code: 2.1.233
- tmux: 3.6
- Shell: bash

## Decisive reproduction — `C-m` vs `Enter`

Fresh trusted dir, launch Claude, type a draft, then compare keys:

```bash
tmux send-keys -t <pane> -l "reply with the single word PONG"   # type draft, do NOT submit
# Press C-m three times:
tmux send-keys -t <pane> C-m ; tmux send-keys -t <pane> C-m ; tmux send-keys -t <pane> C-m
```
Result: draft unchanged, still parked on the input line:
```
❯ reply with the single word PONG
  ⏵⏵ bypass permissions on (shift+tab to cycle)
```
```bash
# Now a single named Enter:
tmux send-keys -t <pane> Enter
```
Result: immediately submitted and answered:
```
❯ reply with the single word PONG
● PONG
✻ Cogitated for 4s
```

Three `C-m` = nothing. One `Enter` = submit. Reproducible 100%.

## Root cause (source refs, dist of 0.20.5 dev)

All worker submit paths press Enter as `C-m`:

- `src/team/tmux-session.ts` — `attemptSubmitRounds()` sends `sendKeyAsync(resolveTarget, 'C-m')` (compiled `dist/team/tmux-session.js:2693,2697`).
- `sendRobustEnter()` / trust-dismiss / ready paths also send `C-m` (`dist/...:2730, 2915, 2964-2966, 3003-3005, 3021-3023`).
- Claude workers are configured to press once: `submitKeyPressesPerRound: workerCli === 'claude' ? 1 : 2` (`dist/...:2647`).

Since Claude Code 2.1.x ignores `C-m` for submission, none of these submit the worker's initial trigger. `attemptSubmitRounds` then still sees the draft text on screen and reports failure, while higher-level dispatch may record delivery as confirmed (cf. #391), leaving the task `pending`.

## Suggested fix

For Claude workers, submit with the tmux **named key** `Enter` instead of (or in addition to) `C-m`:

```ts
// pseudo
const submitKey = workerCli === 'claude' ? 'Enter' : 'C-m';
await sendKeyAsync(resolveTarget, submitKey);
```

`Enter` also works for Codex in my testing, so a blanket switch to `Enter` on the worker-submit paths may be simplest; keeping `C-m` for Codex is fine if you prefer minimal change. Multi-line drafts (OMX's trigger wraps across several lines) may need two `Enter`s ~200ms apart (first settles, second submits) — that combination reliably submits in my tests.

## Secondary: prompt-wording drift on Claude 2.1.x

Two gate detectors use pre-2.1 wording and no longer match Claude Code 2.1.233:

1. **Trust prompt.** `paneHasTrustPrompt()` matches `Do you trust the contents of this directory?` + `Yes, continue|No, quit|Press enter to continue` (`dist/...:2489`). Claude 2.1.233 now shows:
   ```
   Quick safety check: Is this a project you created or one you trust?
   ❯ 1. Yes, I trust this folder
     2. No, exit
   ```
   So `dismissTrustPromptIfPresent()` / the readiness trust-dismiss never fire for Claude workers. (Workaround: pre-seed `~/.claude.json` `projects[<repoRoot>].hasTrustDialogAccepted=true`; Claude inherits trust to subdirectories, so trusting the repo root covers all `.omx/team/**/worktrees/*`.)

2. **Bypass-permissions prompt.** `--dangerously-skip-permissions` triggers a second confirm ("WARNING: Claude Code running in Bypass Permissions mode … 1. No, exit / 2. Yes, I accept"). It is silenced by settings `skipDangerousModePermissionPrompt: true`, not by any keypress — worth handling explicitly for Claude workers.

Adding the 2.1.x wording to the detectors (or keying off the numbered-choice layout) would restore auto-trust; but note that even after the prompt is dismissed, the **`C-m` vs `Enter`** issue above still blocks the actual task submission, so that is the primary fix.

## Impact

Claude workers in mixed `codex,claude` teams never begin their assigned task without manual intervention (`tmux send-keys -t <pane> Enter`). With the key fix, mixed teams run unattended.

Yeachan-Heo:
Fixes #3529

## Problem

Team Claude workers never started their tasks (stuck pending). All worker submit paths sent Enter as `C-m` via `tmux send-keys`, but Claude Code 2.1.x's TUI only submits on the tmux **named** key `Enter`. Reporter's decisive repro: 3× `C-m` = nothing, 1× named `Enter` = submitted, 100% reproducible. Two secondary drifts made it worse:

- `paneHasTrustPrompt()` only matched pre-2.1 wording ("Do you trust the contents of this directory?" / "Yes, continue|No, quit"), so the Claude 2.1.x quick-safety-check trust screen was never auto-dismissed.
- The bypass-permissions confirm now renders as `WARNING: Claude Code running in Bypass Permissions mode` with numbered choices (`1. No, exit` / `2. Yes, I accept`), which the detector did not match.

## Fix

1. **Submit key → named `Enter`** on every worker submit path:
   - `src/team/tmux-session.ts`: `attemptSubmitRounds()` (both queue-first Tab path and per-round presses), both `sendRobustEnter()` twins, `dismissTrustPromptIfPresent()`, `sendToWorker()` trust-guard, and all best-effort/final submit nudges.
   - `src/scripts/tmux-hook-engine.ts`: `buildSendKeysArgv()` now emits the named `Enter` and accepts a `submitKey` override; `sendPaneInput`/`queuePaneInput` (`team-tmux-guard.ts`) thread it through; `team-dispatch.ts` (the default `hook_preferred_with_fallback` dispatch transport) submits with `Enter`.
   - `src/team/scaling.ts`: scale-up final send uses `Enter`.
   - Verified empirically that `Enter` and `C-m` deliver the identical `0x0d` byte to byte-oriented pty readers on tmux 3.2a, so Codex submit behavior is unchanged; multi-line double-press semantics (~200ms apart) are preserved via the existing press counts (`claude` → 1, `codex` → 2).
2. **Trust detector**: `paneHasTrustPrompt()` also matches the Claude 2.1.x wording "Quick safety check: Is this a project you created or one you trust?" with the numbered-choice layout (`1. Yes, I trust this folder` / `2. No, exit`). Both old and new wordings require question + choices, so ordinary pane text does not trip it.
3. **Bypass-permissions detector**: `paneHasClaudeBypassPermissionsPrompt()` also matches the `WARNING: Claude Code running in Bypass Permissions mode` numbered-choice layout (keeping the existing "Press Enter to confirm" variant). Auto-accept (`OMX_TEAM_AUTO_ACCEPT_BYPASS`, default on) then presses `2` + named `Enter` as before.

## Tests

- Claude worker `sendToWorker` submits with named `Enter` — never `C-m`, never queue-first `Tab`.
- Trust detector matches both Codex/pre-2.1 and Claude 2.1.x wordings; negative cases confirm no false positives on ready panes or partial text.
- `waitForWorkerReadyAsync` auto-dismisses the Claude 2.1.x trust prompt and auto-accepts the 2.1.x bypass warning, with named `Enter`.
- Hook-path submit assertions updated (`buildSendKeysArgv` default + explicit `submitKey` override, queue-first Tab-then-Enter, isolated submit argv, leader/auto-nudge paths).

Local validation: `tsc --noEmit`, `biome lint`, `check:no-unused`, full `test:node` suite (7728 tests / 0 fail), `coverage:team-critical` (87.04% lines / 80.51% branches / 94.56% functions, above gates), `verify:plugin-bundle`, `verify:capabilities-lock`, `verify:prompt-guidance`, catalog/prompt-inventory checks.

### DIFF
--- a/src/hooks/__tests__/notify-fallback-watcher.test.ts
+++ b/src/hooks/__tests__/notify-fallback-watcher.test.ts
@@ -390,7 +390,7 @@ if [[ "$cmd" == "if-shell" ]]; then
     exit 0
   fi
   eval "set -- \$sendCommand"
-  if [[ "\$#" -lt 4 || "\$1" != "send-keys" || "\$2" != "-t" || "\$3" != "\$paneId" || ( "\$4" != "-l" && "\$4" != "C-m" ) || ( "\$4" == "-l" && "\$#" -ne 5 ) || ( "\$4" == "C-m" && "\$#" -ne 4 ) ]]; then
+  if [[ "\$#" -lt 4 || "\$1" != "send-keys" || "\$2" != "-t" || "\$3" != "\$paneId" || ( "\$4" != "-l" && "\$4" != "C-m" && "\$4" != "Enter" ) || ( "\$4" == "-l" && "\$#" -ne 5 ) || ( ( "\$4" == "C-m" || "\$4" == "Enter" ) && "\$#" -ne 4 ) ]]; then
     printf '__omx_ralph_input_denied__\n'
     exit 0
   fi
@@ -2209,9 +2209,9 @@ exit 0
       const tmuxLog = await readFile(tmuxLogPath, 'utf8');
       const typeMatches = tmuxLog.match(/send-keys -t %42 -l ping/g) || [];
       assert.equal(typeMatches.length, 1, 'fresh attempt should type once; retries with draft should be submit-only');
-      const cmMatches = tmuxLog.match(/send-keys -t %42 C-m/g) || [];
-      assert.ok(cmMatches.length > 0, 'submit should use C-m');
-      assert.ok(!/send-keys[^\n]*-l[^\n]*C-m/.test(tmuxLog), 'must keep -l payload and C-m submits isolated');
+      const enterMatches = tmuxLog.match(/send-keys -t %42 Enter/g) || [];
+      assert.ok(enterMatches.length > 0, 'submit should use the named Enter key');
+      assert.ok(!/send-keys[^\n]*-l[^\n]*Enter/.test(tmuxLog), 'must keep -l payload and Enter submits isolated');
 
       const request = await readDispatchRequest('dispatch-team', queued.request.request_id, wd);
       assert.equal(request?.status, 'pending');
--- a/src/hooks/__tests__/notify-hook-auto-nudge.test.ts
+++ b/src/hooks/__tests__/notify-hook-auto-nudge.test.ts
@@ -339,9 +339,9 @@ describe('notify-hook auto-nudge', () => {
       assert.ok(existsSync(tmuxLogPath), 'tmux should have been called');
       const tmuxLog = await readFile(tmuxLogPath, 'utf-8');
       assert.match(tmuxLog, defaultAutoNudgePattern('%99'), 'should send nudge response with injection marker');
-      // Codex CLI needs C-m sent twice with a delay for reliable submission
-      const cmMatches = tmuxLog.match(/send-keys -t %99 C-m/g);
-      assert.ok(cmMatches && cmMatches.length >= 2, `should send C-m twice, got ${cmMatches?.length ?? 0}`);
+      // Codex CLI needs Enter sent twice with a delay for reliable submission
+      const enterMatches = tmuxLog.match(/send-keys -t %99 Enter/g);
+      assert.ok(enterMatches && enterMatches.length >= 2, `should send Enter twice, got ${enterMatches?.length ?? 0}`);
     });
   });
 
--- a/src/hooks/__tests__/notify-hook-team-dispatch.test.ts
+++ b/src/hooks/__tests__/notify-hook-team-dispatch.test.ts
@@ -1574,7 +1574,7 @@ exit 0
     }
   });
 
-  it('retries submit with isolated C-m and does not retype when trigger already present', async () => {
+  it('retries submit with isolated named Enter and does not retype when trigger already present', async () => {
     const cwd = await mkdtemp(join(tmpdir(), 'omx-hook-team-dispatch-'));
     const fakeBinDir = join(cwd, 'fake-bin');
     const tmuxLogPath = join(cwd, 'tmux.log');
@@ -1612,9 +1612,9 @@ exit 0
       const tmuxLog = await readFile(tmuxLogPath, 'utf8');
       const typeMatches = tmuxLog.match(/send-keys -t %42 -l ping/g) || [];
       assert.equal(typeMatches.length, 1, 'fresh attempt should type once; retries with draft should be submit-only');
-      const cmMatches = tmuxLog.match(/send-keys -t %42 C-m/g) || [];
-      assert.ok(cmMatches.length > 0, 'submit should use C-m');
-      assert.ok(!/send-keys[^\n]*-l[^\n]*C-m/.test(tmuxLog), 'must not mix -l payload with C-m submit');
+      const enterMatches = tmuxLog.match(/send-keys -t %42 Enter/g) || [];
+      assert.ok(enterMatches.length > 0, 'submit should use the named Enter key');
+      assert.ok(!/send-keys[^\n]*-l[^\n]*Enter/.test(tmuxLog), 'must not mix -l payload with Enter submit');
 
       const request = await readDispatchRequest('alpha', queued.request.request_id, cwd);
       assert.equal(request?.status, 'pending');
@@ -2134,7 +2134,7 @@ exit 0
       const exactGlobalPaneProof = /^list-panes -a -F #\{pane_id\}\t#\{pane_dead\}\t#\{pane_pid\}$/;
       assert.match(commands[0] || '', exactGlobalPaneProof);
       assert.match(tmuxLog, /send-keys -t %42 C-u/, 'the live proof should permit the clear effect');
-      assert.doesNotMatch(tmuxLog, /send-keys -t %42 C-m/);
+      assert.doesNotMatch(tmuxLog, /send-keys -t %42 (?:C-m|Enter)/);
       const ownerProof = 'show-option -qv -p -t %42 @omx_team_pane_owner_id';
       for (const [index, command] of commands.entries()) {
         if (!/^(send-keys|paste-buffer)\b/.test(command) || !command.includes('-t %42') || /^send-keys\b.*\s-l\s/.test(command)) continue;
--- a/src/hooks/__tests__/notify-hook-team-leader-nudge.test.ts
+++ b/src/hooks/__tests__/notify-hook-team-leader-nudge.test.ts
@@ -661,9 +661,9 @@ describe('notify-hook team leader nudge', () => {
       assert.match(tmuxLog, /\[OMX\] All 2 workers idle/, 'should emit all-workers-idle nudge');
       assert.doesNotMatch(tmuxLog, /\[OMX_INTENT:/, 'should keep orchestration intent out of injected display text');
       assert.match(tmuxLog, /\[OMX_TMUX_INJECT\]/, 'should include injection marker');
-      const submitMatches = tmuxLog.match(/send-keys -t %99 C-m/g) || [];
-      assert.equal(submitMatches.length, 2, 'leader nudge should submit with isolated double C-m');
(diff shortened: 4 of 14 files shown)


Yeachan-Heo (commented):
## Adversarial review — verified at head b3b34543 vs base 5ee2193f

**Verdict: APPROVE-equivalent — MERGE_READY.** I re-derived every claim independently rather than trusting the implementation report. No correctness, security, or regression blocker.

> GitHub rejected the formal `--approve` ("Review Can not approve your own pull request") because this review lane runs under the PR author's identity. Nothing in the code blocks merge; applying the formal approval from a non-author reviewer identity is the only remaining action. OWNER_CONFIRMATION_REQUIRED

### Diff classification (all 14 files, 312+/80−)
- **Submit-key switch**: `src/team/tmux-session.ts` (all 6 paths: `attemptSubmitRounds` both branches, both `sendRobustEnter` twins, `dismissTrustPromptIfPresent`, trust-guard in `sendToWorker`, final best-effort + post-submit nudges, both `acceptClaudeBypass…` submits), `src/scripts/tmux-hook-engine.ts` (`buildSendKeysArgv` default + `submitKey` override), `src/scripts/notify-hook/team-tmux-guard.ts` (`sendPaneInput`, `queuePaneInput`), `src/scripts/notify-hook/team-dispatch.ts`, `src/team/scaling.ts`.
- **Trust/bypass detection**: `paneHasTrustPrompt` (2.1.x quick-safety-check + numbered choices), `paneHasClaudeBypassPermissionsPrompt` (2.1.x numbered layout as an OR arm, old `Enter to confirm` arm retained).
- **Tests**: 7 test files updated + 1 strengthened; `src/types/tmux-hook-engine.d.ts` (+1 param). **Zero unrelated churn** — every hunk maps to the fix or its assertions. No `dist/` committed, single-commit PR, commit message accurate.

### Checklist findings

**1. Submit-path coverage — grep-verified exhaustively.** The only remaining `C-m` key sends in the whole tree are (a) `src/question/renderer.ts` C-m into a **freshly-created question-UI shell pane** (`buildSendPaneArgvs(..., false)` then one C-m to run a shell command — not a worker submit; out of scope), (b) `src/notifications/tmux-detector.ts` `buildSendPaneArgvs` used by reply-listener / MCP hermes-bridge, which target the *leader/user* Codex pane, not Claude workers, (c) `src/hooks/extensibility/sdk/tmux.ts` plugin SDK `sendTmuxKeys`, (d) `emitRalphContinueSteer`, which **explicitly excludes team panes** (`parseRalphContinuePaneBinding` rejects `paneOwnerId.startsWith('team:')`). None can submit a Claude worker trigger. The worker-trigger surface is fully converted.

**2. Codex 2-press semantics preserved.** `buildWorkerSubmitPlan` keeps `submitKeyPressesPerRound: claude=1, else=2`; `buildSendKeysArgv` keeps the 1..4 clamp and now honors `submitKey` (unit-tested with a C-m override). `team-dispatch.ts:924` still derives `submitKeyPresses` from the resolved worker CLI. Hook-path submits remain spaced (`submitDelayMs` sleep before each press) and re-verified per press.

**3. Multi-line/wrapped drafts.** Handled by structure, not just key choice: `attemptSubmitRounds` re-captures after each round and only returns success when the normalized trigger no longer appears, so a first Enter that soft-wraps a draft keeps getting pressed; the Codex queued-banner case still fails closed (`submit_queued_after_tool_call`) and the wrapped-draft case still refuses false confirmation (both regression tests kept, now asserting ≥4 Enter presses). Acceptable residual risk, same as before.

**4. Detection regex robustness.** Detection remains a conjunction of independent line-level facts (question line + choice lines within the last 12/20 non-empty lines, `.trim()`ed, `\r`-stripped), not one anchored literal — so numbered-option layouts with `❯` glyphs or leading whitespace still match, and per-line matching sidesteps terminal re-wrap entirely. False-positive risk on Codex panes is unchanged in *shape* (both wordings are Claude-specific and require the exact "…trust this folder" + "No, exit" pair; the added negative test `'Quick safety check: unrelated banner…'` proves the question fragment alone is insufficient). Note `paneHasTrustPrompt`/`paneHasClaudeBypassPermissionsPrompt` are CLI-agnostic by design — they run before CLI resolution in `evaluateStartupDirectTriggerSafetyCapture`; the Claude-specific *actions* are gated on actual capture content.

**5. Bypass auto-accept is capture-gated and idempotent.** Every accept path (`waitForWorkerReady[Detailed]Async`, `sendToWorker`) only presses `2`+Enter **after** a fresh capture proves the prompt is on screen, with pane re-proof (`verifyForEffect`/pinned PID) between the literal and the submit. If the prompt isn't present, no keys are sent at all — no stray input on older Claude versions. Double-accept while the prompt is still rendered just re-selects choice 2; the loop exits once the capture stops matching. Opt-outs preserved (`OMX_TEAM_AUTO_ACCEPT_BYPASS=0`, `OMX_TEAM_AUTO_TRUST=0`).

**6. Tests assert the named key specifically.** New/updated tests assert `send-keys … Enter` **and** `doesNotMatch … C-m` (Claude submit also bans `Tab`), plus a tab-preserving ordering assertion (`Tab` before `Enter`) and the `submitKey` override unit test. Notably, two negative guards were *strengthened* from `doesNotMatch C-m` to `doesNotMatch (?:C-m|Enter)` (`tmux-session.test.ts:8792`, `notify-hook-team-dispatch.test.ts:2137`) — nothing was weakened or deleted to pass.

**7. Regression surface.** For non-Claude teams the only behavioral delta is the byte stream `Enter` vs `C-m` (both 0x0d to a byte-oriented reader); press counts, rounds, delays, queue-first gating, and strict/adaptive semantics are untouched. `scaling.ts` change is its one scale-up nudge. The fallback-watcher fake-tmux allowlist correctly keeps both `C-m` (ralph steer still uses it) and `Enter` permitted — it's a test double allowlist, not production policy, so no inconsistency.

**8. Verified locally** (fresh `npm install` + build): `npm run build` clean; `biome lint` clean on all changed files; the 8 changed suites — `tmux-session`, `tmux-hook-engine`, `notify-hook-team-tmux-guard`, `notify-hook-team-dispatch`, `notify-fallback-watcher`, `notify-hook-auto-nudge`, `notify-hook-team-leader-nudge`, `codex-native-hook` — **1330 tests / 0 failures**. CI at head: Typecheck, Lint, Test (Node 20 cli-core-rest, hooks-notify-platform, team-state-runtime; Node 22 smoke), Coverage Gate (Team Critical), Ralph Persistence Gate, CI Status — all green.

### Minor observations (non-blocking)
- `src/notifications/tmux-detector.ts:104` comment still says "Codex CLI uses raw input mode; send C-m …" — its callers are leader-pane/notifications only, so behavior is correct, but the comment now reads stale relative to this PR's model. A one-line follow-up comment touch-up would be nice.
- `paneHasTrustPrompt`'s Claude arm requires both the "Quick safety check" question *and* the numbered pair; if a future Claude build drops the numbering, detection will miss again — inherent to wording-matching, and the same exposure existed pre-PR.

MERGE_READY · OWNER_CONFIRMATION_REQUIRED (self-approval blocked by GitHub)

—
*[Yeachan-Heo's gaebal-gajae (clawdbot) 🦞]*

PR #3533, opened 2026-08-18
fix(macos): unbreak setup on darwin, retire deprecated skills on update, refuse working-tree self-update

Yeachan-Heo:
## Why

At `dev` HEAD `3e162d1a` the repo is **not green on macOS**, and CI could not see it: the only macOS job is `ralplan-preflight-macos`, while the grouped CLI suite is Ubuntu-only.

| Suite | Before (darwin) | After |
|---|---|---|
| `setup-install-mode` | 69/129 | **129/129** |
| `setup-agents-overwrite` | 12/26 (one 180s hang) | **26/26** |
| `session.test.js` | 94/149 | **127/149** (residual tracked, see below) |
| `update` | 82/82 | 82/82 |
| `setup-skills-overwrite` | 12/12 | **14/14** (+2 new) |

Every "before" number was reproduced at HEAD in a clean `git worktree`, so none of it is caused by local changes.

## What

1. **`fix(update)`** — a linked/bare checkout has no package-manager owner, so the launch check printed *"Unable to determine whether this global install is owned by npm or Bun"* on **every run** (the early return preceded `writeUpdateState`, so the 24h cadence was never stamped). Working-tree builds are now classified explicitly: passive checks stay silent and stamp the cadence; `omx update` refuses with an accurate message instead of advising a reinstall that would overwrite the developer's tree.
2. **`fix(setup)` skill retirement** — cleanup enumerated candidates from the manifest plus a hardcoded `HARD_DEPRECATED_SKILL_NAMES` list, so a skill deleted from the catalog outright was unreachable *even with `--force`*, and plugin mode never swept the legacy directory at all. Ownership is now proven by the `[OMX] ` install badge (the only evidence that survives catalog deletion), candidates come from the destination directory, retired directories are archived then removed on every refresh, and user-authored or user-edited copies are never touched. The hand-kept list is gone.
3. **`fix(plugin)`** — the cache namespace guard rejected *any* symlinked ancestor, which macOS always has via `/var -> /private/var`. It now canonicalizes the Codex home and requires only the OMX-owned components below it to be real directories. The security property is unchanged: a symlinked `plugins/` inside the home is still refused, and that test still asserts the foreign directory stays empty.
4. **`fix(setup)` session liveness** — `isSessionStale()` maps `identity-indeterminate` to stale, and macOS records no process-birth evidence, so **OMX classified its own live session as dead** and setup would overwrite a running session's AGENTS.md overlay. The overwrite guard now asks `classifySessionStateLiveness(...) !== 'stale-dead'`. The classifier is deliberately untouched.
5. **`ci`** — registers the four platform-sensitive suites on the existing macOS job.
6. **`test(session)`** — `validLockOwner()` hardcoded a Linux v1 owner while the harness's `matchingObservation()` defaults to `process.platform`, so off Linux every owner fixture was cross-platform and degraded live/dead/reused into indeterminate. The fixture is host-aligned; the v1 Linux default is byte-identical so Linux CI cannot regress.

## Deliberately not done

`session.test.js` still has 19 darwin failures. Closing them at the primitive requires weakening a pinned safety contract, so I did not:

- mapping `isSessionStale`'s indeterminate to not-stale breaks `returns true on Linux when identity metadata is missing` / `... cannot be read`;
- returning `usable` for alive-plus-non-linux in `probeIdentitylessProcess` reaches 140/149 but violates `classifies identity-less live non-Linux pointers as identity-indeterminate`, which exists precisely so an **alive-but-reused pid cannot inherit session authority**.

Both were measured and reverted; the rationale is recorded as comments at those call sites so they are not retried. The residual needs a per-call-site read-vs-authority audit and is tracked separately.

## Verification

All numbers above were run on darwin from a fresh `npm run build` of this branch. Also clean: `verify:native-agents`, `verify:plugin-bundle`, `verify:capabilities-lock`, `verify:prompt-guidance`, `generate-catalog-docs --check`, `prompt-inventory --check`, `uninstall` 80/80, `setup-refresh` 35/35, `codex-plugin-layout` 49/49, catalog suites 20/20, `biome lint`, `tsc --noEmit`, `tsconfig.no-unused`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -423,6 +423,12 @@ jobs:
           env -u CODEX_THREAD_ID -u OMX_SESSION_ID -u GJC_SESSION_ID -u TMUX -u TMUX_PANE \
             node --test --test-name-pattern='returns the attached client to the shell' \
             dist/cli/__tests__/launch-fallback.test.js
+      - name: Verify macOS-sensitive setup and update suites
+        run: |
+          node --test --test-force-exit dist/cli/__tests__/setup-install-mode.test.js
+          node --test --test-force-exit dist/cli/__tests__/setup-agents-overwrite.test.js
+          node --test --test-force-exit dist/cli/__tests__/update.test.js
+          node --test --test-force-exit dist/cli/__tests__/setup-skills-overwrite.test.js
 
   test:
     name: Test (Node ${{ matrix.node-version }} / ${{ matrix.lane }})
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -186,3 +186,16 @@ Adjust the terminal pattern if your client advertises a different terminfo name.
 - For explicit shell-native read-only evidence or `--tmux-pane` summaries, use `omx sparkshell -- <command>`.
 
 The earlier explore harness fallback boundaries (sparkshell-backend fallback and in-harness model fallback) no longer apply to a user-facing command, because the command no longer runs a harness. Internal harness-resolution helpers remain only to support `omx doctor`/native-asset diagnostics.
+
+## `omx update` refuses to run on a working-tree build
+
+A checkout that you run directly, or link into a global root with `npm link`, is not owned by any package manager. `omx update` refuses it explicitly:
+
+```
+[omx] This build runs from a working tree (/path/to/oh-my-codex), which no package manager owns.
+      Self-update is refused; pull and rebuild the checkout instead.
+```
+
+Update such a build with `git pull && npm run build`. The launch-time check makes the same determination silently and records the cadence, so a linked dev build never nags on startup and never overwrites your tree.
+
+This is distinct from `[omx] Unable to determine whether this global install is owned by npm or Bun`, which means the package root *is* inside a global `node_modules` but neither manager could be validated as its owner — reinstall globally with npm or Bun.
--- a/src/cli/__tests__/setup-skills-overwrite.test.ts
+++ b/src/cli/__tests__/setup-skills-overwrite.test.ts
@@ -110,7 +110,7 @@ describe('omx setup skills overwrite behavior', () => {
       await mkdir(staleWebCloneDir, { recursive: true });
       await writeFile(
         join(staleWebCloneDir, 'SKILL.md'),
-        '---\nname: web-clone\ndescription: old standalone pipeline\n---\n\nClone a target website from its URL.\n',
+        '---\nname: web-clone\ndescription: "[OMX] old standalone pipeline"\n---\n\nClone a target website from its URL.\n',
       );
       assert.equal(existsSync(staleWebCloneDir), true);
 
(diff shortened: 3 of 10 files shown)


Yeachan-Heo:
Maintainer review ownership is active for exact head `ea27659e6c284b94cd12917da8b81ce09d974d75` targeting `dev@3e162d1a10684c4062031ffe827cbac867cabb7d`. A dedicated GJC lane is attached to adversarially verify the macOS setup, skill-retirement ownership, plugin-cache namespace, session-liveness, and working-tree update contracts through terminal merge/close disposition. Exact-head CI is currently green.

—

Yeachan-Heo (commented):
## Adversarial exact-head review — MERGE_READY

Reviewed at exact head `ea27659e6c284b94cd12917da8b81ce09d974d75` against base `dev@3e162d1a10684c4062031ffe827cbac867cabb7d`. Exact-head CI run 32120723980 terminal green (CI Status, build/dist, lint/typecheck, Rust/native, coverage, smoke, all three Node lanes, cache integrity ×3 platforms, Ralph persistence, both macOS ralplan/security lanes — head confirmed unchanged since). Self-approval is blocked by GitHub for owner-authored PRs; this COMMENT review carries the signed verdict per the lane contract.

### Working-tree detection (update.ts, package-manager-ownership.ts)
- `isWorkingTreeInstall` classifies by `realpath(packageRoot)`'s parent being `node_modules`, so `npm link` (global symlink resolving to the checkout) is correctly refused even though the pre-realpath parent *is* a global `node_modules`. Legit npm (`/opt/homebrew/lib/node_modules/...`) and bun (`~/.bun/install/global/node_modules/...`) roots classify false; `win32` uses the win32 path module so `AppData\Roaming\npm\node_modules` classifies false. Unresolvable `realpath` fails open to `false` — correct, because the pre-existing ownership validator (`matchesCurrentInstall` realpaths every candidate) still gates the actual transaction; this path only picks the *refusal message*.
- The check runs only when ownership resolution already failed, so no legitimate global install can regress to the working-tree notice; the pnpm/yarn and "unable to determine owner" diagnostics are preserved verbatim.
- Passive cadence: launch path stamps `last_checked_at` only, no `fetchLatestVersion` call, no prompt, no mutation (proven by `latestCalls === 0` + `askYesNo` throwing in tests). Explicit `omx update` prints the refusal with zero network and zero mutation. `skipped` is a new distinct status consumed nowhere as failure; no caller treats unknown statuses as errors.

### Skill ownership / deletion safety (setup.ts)
- The `[OMX] ` badge is the delete authority: user-authored skills (no badge), badge-stripped/edited copies, malformed/missing SKILL.md (parse throws → not managed), and non-directory entries are all kept (`summary.skipped`), never deleted. Badge spoofing degrades only to backup-then-remove of a skill dir — and **every** removal goes through `removeDirectoryCopyAware`, which copies the full directory into `.omx/backups/setup/...` before `rm`, so even a spoofed deletion is byte-recoverable (pinned by the backup test). Deletion is opt-out-by-content, not by name list.
- Catalog-dropped skills retire on plain refresh (this is what makes `omx update` drop deprecated skills), and `--force` retains its prior semantics for catalog-known names. The legacy plugin-mode orphan loop deletes only OMX-badged dirs not in the installable set; team-skill and dry-run paths preserved. Idempotent: a second refresh finds nothing left to remove.

### Plugin cache namespace (plugin-marketplace.ts)
- Containment is now `relative(resolve(codexHome), resolve(cacheBase))` with `..`/absolute rejection — traversal outside the Codex home is refused. macOS `/var -> /private/var` TMPDIR and symlinked homes are canonicalized via `realpath(codexHomeDir)` instead of rejected (the pre-PR behavior that broke macOS setup), while symlinked *managed* components (`plugins/`, `cache/`, …) are still lstat-refused per-component, plus a post-`mkdir` re-lstat closing the create race. The original foreign-write refusal (`inspectCacheRoot` manifest check) is unchanged. Symlink refusal is pinned by the dedicated test (`readdir(foreignPlugins)` stays empty).

### Session overlay liveness (setup.ts, session.ts)
- `sessionIsActive` now requires `classifySessionStateLiveness(...) !== 'stale-dead'`: only a positively dead pid releases the overlay. On darwin a live session classifies `identity-indeterminate` and is correctly treated as active; PID reuse stays `stale-dead`-only when birth evidence proves death, and all *authority-granting* paths still require `usable` — this is a read-side call-site-only distinction with no authority escalation. The pinned conflicting suite cases (`isSessionStale` semantics) are untouched; the fixture change aligns the synthetic lock-owner platform with the host like `matchingObservation()` already did (byte-identical on Linux; explicit overrides still win).

### CI coverage
The four macOS-sensitive suites execute built `dist` after a fresh source build on the macOS runner (verified in job logs: 129/26/82/14 tests, 0 fail). Duration impact is ~20s on the macOS lane — no hidden cost.

### Docs
`docs/troubleshooting.md` matches the implementation verbatim (message text, silent launch behavior, and the distinction from the ownership-error case). No stale references to the retired hard-deprecated list remain.

No material unresolved product decision exists: the PR is owner-authored, narrowly scoped, safety-hardening, and terminal green. No blockers found.

—

Yeachan-Heo:
## Merge + dogfood evidence — terminal

**Merge SHA:** `325347e4d79291698fe11c5853aa215911022f2a` (merged 2026-08-18T10:43:30Z, PR #3533 → `dev`; exact head `ea27659e6c284b94cd12917da8b81ce09d974d75`, base `3e162d1a10684c4062031ffe827cbac867cabb7d`).

**Canonical local dev fast-forward:** `/home/bellman/Workspace/oh-my-codex` fast-forwarded `3e162d1a1 → 325347e4d` (ff-only, no new commits).

**Fresh-dev build:** `npm ci && npm run build && npm run lint && npx tsc --noEmit` — all clean (802 files linted, no errors).

**Focused suites on fresh dev dist:** setup-install-mode 129/129, setup-agents-overwrite 26/26, update 82/82, setup-skills-overwrite 14/14, session 149/149, plugin-marketplace-idempotent 3/3 — 0 failures.

**Setup/update dogfood (built dist CLI, isolated HOME):**
- Project-scope setup: 24 skills installed, exit 0.
- Skill retirement on plain refresh: badged `prometheus-strict` removed (`removed=1, backed_up=1`), bytes recoverable at `.omx/backups/setup/<ts>/.codex/skills/prometheus-strict/SKILL.md`; unbadged user skill `my-own` kept.
- Working-tree self-update refusal: `omx update` printed `[omx] This build runs from a working tree (...), which no package manager owns. Self-update is refused; pull and rebuild the checkout instead.` — no network, no mutation, correct notice (not the reinstall error).
- Passive launch cadence: `maybeCheckAndPromptUpdate` on the working-tree build produced zero log output and stamped `last_checked_at` — no prompt, no re-probe.
- Plugin cache symlink guard: symlinked `plugins/` component writes nothing through the link (foreign target untouched); `materializePackagedOmxPluginCache` with no packaged marketplace returns `unavailable` without throwing.

**Post-merge exact-dev CI:** run 32128169330 on `325347e4d` terminal green — all jobs success/skipped, including Native Rust (Darwin), both Ralplan preflight security macOS lanes, and Native Cache Integrity (macos-latest).

—

PR #3538, opened 2026-08-18
docs: fix broken star history chart across READMEs

FaintFlower:
The star history chart in the main README and the localized translations no longer renders because the upstream service is broken under current GitHub stargazer API restrictions. This change repoints the chart image and its wrapping link to a working alternative so the chart is visible again.

Updated: README.md, docs/readme/README.el.md, docs/readme/README.pl.md, docs/readme/README.uk.md, docs/readme/README.vi.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ If this happens, try:
 
 ## Star History
 
-[![Star History Chart](https://api.star-history.com/svg?repos=Yeachan-Heo/oh-my-codex&type=date&legend=top-left)](https://www.star-history.com/#Yeachan-Heo/oh-my-codex&type=date&legend=top-left)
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks for the patch. I cannot accept replacing the project's chart with an unaffiliated third-party mirror on this evidence. The current `api.star-history.com` SVG endpoint resolves and returns `200 image/svg+xml` for this repository in an independent live probe, while the proposed host introduces a new external trust/dependency surface without ownership, security, or durability evidence. The PR also targets `main`; contribution work must be based on the active development branch.

Closing rather than merging. If the official endpoint fails reproducibly, please provide the exact failing URL, response/status, time/region, and browser/render evidence against current `dev`; a future fix should prefer the official provider or remove the chart rather than silently redirecting readers to an unverified mirror.

—

FaintFlower:
If you were human you would have looked at the SVG and seen the big unmissable error

Issue #3536, opened 2026-08-18
[Bug]


dmitrii8V:
# Verified Team workers are blocked by PROVENANCE_DENIED when using an external OMX state root

## Summary

When OMX Team uses an external OMX_ROOT or OMX_TEAM_STATE_ROOT, verified Team workers cannot execute delegated product-code writes.

The Team runtime successfully creates valid worker identity, config, manifest, pane, worktree, state-root, and leader-CWD metadata. However, if the external state root has no top-level session.json containing a canonical cwd, every worker mutation is blocked with:

PROVENANCE_DENIED: the selected workflow state root has no usable canonical session cwd

This happens even though the worker passes OMX's authoritative Team-worker validation. Additionally, omx doctor --team reports that all Team checks pass.

## Environment

- OS: macOS 26.5.1 (Build 25F80), arm64
- Node.js version: v24.16.0
- oh-my-codex version: v0.20.5
- Codex CLI version: codex-cli 0.147.0

## Steps to reproduce

1. Open a Git repository inside tmux.

2. Select an isolated external OMX root:

export OMX_ROOT="$(mktemp -d /tmp/omx-team-repro.XXXXXX)"

3. Start a Team with explicit workers and delegated implementation tasks:

omx team 3:executor "1. Implement task A in repository A. 2. Implement task B in repository B. 3. Implement task C in repository C."

4. Wait for the workers to start.

5. Confirm that the generated worker environment contains values similar to:

OMX_TEAM_WORKER=<display-name>/worker-1
OMX_TEAM_INTERNAL_WORKER=<session-scoped-internal-name>/worker-1
OMX_TEAM_STATE_ROOT=<external-root>/.omx/state
OMX_TEAM_LEADER_CWD=<leader-repository>

6. Confirm that the following files contain consistent worker, pane, worktree, leader-CWD, and state-root metadata:

.omx/state/team/<internal-team>/config.json
.omx/state/team/<internal-team>/manifest.v2.json
.omx/state/team/<internal-team>/workers/worker-1/identity.json

7. Let a worker invoke any product-code mutation, such as Write, apply_patch, or a Bash command that writes a source file.

8. Run:

omx doctor --team

## Expected behavior

A worker whose internal identity, pane, worktree, Team state root, and leader CWD have been verified should be allowed to modify delegated product files.

Writes to OMX orchestration metadata, Team config, manifests, identities, and protected state files should remain blocked.

[...]

PR #3539, opened 2026-08-19
fix(cleanup): advisory workflow gates, pipeline deletion, retired keyword surfaces, state SSOT truthfulness, darwin pointer protocol

Yeachan-Heo:
Follow-up to #3533, which merged only the first six commits (the macOS setup/update fixes). This carries the remaining cleanup-epic work: 27 commits closing five of the six approved decisions from the deep-interview spec, plus a session-keying fix.

## What lands

**Advisory-but-visible workflow gates (AC-1).** The ten checks in `validateAutopilotCompletionTransition` return a structured `AutopilotCompletionAdvisory` instead of refusing. All three transports record it, a terminal state with skipped gates persists `completion_status: 'complete-with-skipped-gates'`, and `omx autopilot status` renders `complete with skipped gates` — never clean success. Identity, scope and corruption checks stay fail-closed.

**Pipeline deletion (AC-2).** `src/pipeline/**` removed (1,773 src + 2,048 test lines, zero production importers), along with its CI test root.

**Deleted-surface routing (AC-3).** `$ralph` and `$ultrawork` no longer activate: a registry trigger is what makes a token activate, so routing them contradicted their own sunset guidance and mutated workflow state on names the catalog no longer ships. They now emit non-activating diagnostics. The over-broad `parallel` trigger is gone, both two-hop sunset dead ends point at live skills, four doc surfaces and both byte-identical ralplan mirrors are corrected, and a catalog-driven contract test prevents regression. The `omx ralph` CLI and ralph persistence runtime are untouched.

**State SSOT truthfulness (AC-4).** The automatic launch-time neutralizer is deleted rather than fitted with a provenance heuristic — a real projection carries no version marker to discriminate 0.20.x state from current, so it could terminalize a live session once, silently. Retirement is now only the explicit `omx doctor --repair-state` archive path. Adds a published `src/state/namespace-owners.ts` and replaces the lexical single-writer test with a real audit.

**Darwin session pointer protocol (AC-5).** `session.test.js` goes 94/149 → 146/149 with zero failures and no suite-level skip, via one shared `isUnobservableLivePointerState` rule separating read semantics from authority. `classifySessionProcess` still returns `identity-indeterminate`, lock reaping still requires a positively dead owner, and cross-platform/malformed/foreign-cwd evidence all still fail closed.

**Session-keyed skill state.** `team` was the only stateful skill still declaring `scope: 'root'`, writing a shared unscoped projection that two sessions would clobber. It is session-keyed now; all seven stateful skills write session-keyed paths with zero unscoped projections.

**macOS launcher path aliases.** The madmax detached context key was alias-sensitive in both components, so the same directory reached as `/var/...` and `/private/var/...` hashed to different keys — defeating a forgery-proof verification design and allowing a duplicate detached launch. This net-fixes two pre-existing `launch-fallback` failures (that file is 4 failures on `dev`, 2 here).

## Review process

Seven review generations, each joining a cleaner lane, an architect lane and an executor QA/red-team lane on one frozen `sourceHash`. Generations 1–6 each returned BLOCK and each found a genuinely different defect; all were fixed with regressions. The recurring class was lossy normalization erasing "supplied but malformed" into "absent" — found in six different writers — which is now enforced at one choke point in the completion gate, with `src/state/handoff-carrier.ts` as the single definition.

## Known open, deliberately

- **Writer-audit soundness.** Detection is lexical: it sees literal projection paths, the suffix constant, `getStateFilename`/`resolveSeedStateFilePath`, and file handles opened directly on such a path. It does **not** follow a path bound to an intermediate variable. A real AST pass is the correct fix, but TypeScript 7 exposes no AST at its main entry (only `typescript/unstable/*`, where `NodeHandle` has just `constructor`/`resolve`), so it needs a parser dependency. The gap is asserted as a test case so it reads as a boundary, not as coverage.
- **Native pointer adoption.** `reconcileNativeTransition` can adopt a live pointer whose `native_session_id` is absent. Verified to pre-exist on `dev` for `usable` pointers; this branch makes darwin behave like Linux rather than opening a new hole. Closing it needs a native-to-wrapper binding, which would put an authority token in the launched environment and so needs a threat model.
- **Two `launch-fallback` failures.** Pre-existing (4 on `dev`). `handleMadmaxDetachedReuse` looks the active record up by the supplied `OMX_MADMAX_DETACHED_CONTEXT`, while the protocol mints a context per boxed run and propagates it inward, so an outer launch's supplied key never matches. Honouring the supplied key makes the first run abort, because `publishActiveRecord` verifies rather than writes the record — which is the invariant that test's name protects. Which side is wrong is a product-intent call.

## Verification (darwin)

27 suites, 1,723 passing. `session.test.js` 146/149 (3 per-case `t.skip()` with stated reasons, no suite-level skip), `codex-native-hook` 664/664, `keyword-detector` 179/179, `setup-install-mode` 129/129 (was 69/129), `setup-agents-overwrite` 26/26 (was 12/26 plus a 180s hang), `update` 82/82, `team/runtime` 196/196. `tsc`, `tsconfig.no-unused`, `biome` (794 files), `verify:generated` and `git diff --check` all clean.

`npm run test:recent-bug-regressions` exits 1 on the two pre-existing `launch-fallback` failures described above, and on nothing else.


Yeachan-Heo:
Maintainer review has started for exact head `b204cba04c4f9f5d4cc5cf54a6a048fd4b0ad6b8` against `dev@0555eaf86e4422161b9146c8dc0ae39c89a2137e`. This is a large, behavior-changing cleanup set, so it is under dedicated exact-head adversarial review while the remaining CI jobs finish. No merge-readiness claim is being made yet; the disclosed baseline failures and open contract boundaries are part of the review scope.

—

Yeachan-Heo:
Verdict: REQUEST_CHANGES

Reviewed the current exact PR head `59d1eef94a7f68697d1cb408d02dc927877b1c75` against base `0555eaf86e4422161b9146c8dc0ae39c89a2137e`. The fix-forward commit correctly closes three earlier findings: canonical artifact containment now rejects external symlink targets, Darwin identity-indeterminate pointers are not treated as usable authority, malformed handoff carriers reject non-plain objects, and `dist/autopilot/__tests__` is included in the `cli-core-rest` lane.

Blocking findings remain:

1. **Workflow single-writer/transaction contract is still false.** `src/hooks/keyword-detector.ts:3597-3602` writes the supervised Autopilot parent projection through `writeStateFile` directly. `src/state/operations.ts:87-114` shows that `writeStateFile` is only an atomic write; it does not acquire the `withStateWriteLock` queue or perform the state-operation revalidation. The new `src/state/namespace-owners.ts` workflow owner map does not declare `hooks/keyword-detector.ts`, and the audit does not catch this path. Two concurrent child reconciliations can therefore race on a read-modify-write projection, and a transition can leave the parent projection changed without the corresponding state-operation transaction. Route this authorizing path through the sanctioned locked/revalidated writer, or explicitly redesign and test its transaction boundary; do not leave the owner map claiming a single workflow writer while this path bypasses it.

2. **Skill-retirement ownership proof is forgeable by the same user who owns the skills directory.** `src/cli/setup.ts:2323-2346,2381-2434` trusts `.omx-installed-skills.json` beside the skills and compares its digests, while `src/cli/setup.ts:5977-5999,6095-6127` uses that receipt plus the badge as deletion authority. Both are user-writable and neither is integrity-bound to an installation event or trusted source. A user/process can create or replace the receipt and badge with digests matching a directory containing user files, then trigger ordinary cleanup and have that directory archived/deleted as “unmodified OMX.” `--force` being intentionally destructive does not address the ordinary path. The receipt needs an authority that an untrusted user cannot mint, or cleanup must be made non-destructive when ownership cannot be proven.

Validation evidence:

- Exact-head `npm run build`, `npm run check:no-unused`, `npm run verify:generated`, and focused completion/session/carrier tests passed; the fix-forward commit was pushed as `59d1eef94`.
- Exact-head `npm run test:recent-bug-regressions` ran serially and reproduced exactly the two disclosed launch-fallback failures: `does not let the outer launcher fabricate leader-owned active records` and `does not reuse the same active-detached lock for independent --madmax --high launches`; the other five files passed. The exact-base run reproduced the same two failures, so these are baseline and not attributed to this PR.
- Terminal CI run `32237681120` is green across build, typecheck, lint, generated/dist checks, Rust, both Darwin preflight jobs, coverage, Ralph, smoke, hooks, team-state, and `cli-core-rest`.
- The PR is still reported `BEHIND` base; no merge was performed.

Please resolve the two state/cleanup blockers and rerun the exact-head review.

—

Yeachan-Heo:
Exact-head maintainer review for `0bce731e6f9662476d1fd2d36b65c6f78203f33f` against `dev@0555eaf86e4422161b9146c8dc0ae39c89a2137e`.

Review scope covered the two prior blockers and the surrounding state, security, lifecycle, generated-artifact, CI, and platform contracts.

- The supervised Autopilot parent projection now uses `executeStateOperation('state_write', ...)`, so the mutation is serialized through the canonical state-write lock and revalidated operation path. The writer preserves concurrent fields by submitting only changed fields; the new concurrent Autopilot read-modify-write regression passes.
- The state-writer audit passes with one sanctioned workflow namespace writer and no undeclared mode projection writer.
- Ordinary skill retirement now uses a setup-owned receipt outside the skills directory and requires a complete unmodified-install digest match. A forged skills-directory receipt/badge plus a user-owned file is preserved, and the setup skill suite passes.
- The earlier artifact-containment, Darwin identity-authority, malformed handoff-carrier, and CI test-root findings remain covered by the focused tests and generated checks.
- Exact-head recent regressions pass: 196 tests, 7 files. Terminal CI run `32240956969` is green across all required lanes, including full source build, typecheck, lint, generated/dist checks, Rust, coverage, Ralph, smoke, hooks, team-state, and platform preflight lanes.

Verdict: `MERGE_READY`

—

Yeachan-Heo:
Fresh exact-head maintainer review for rebased head `8c2c0a0c03afc2b5cad3b30ee8983650c9e99b9d` against `dev@0555eaf86e4422161b9146c8dc0ae39c89a2137e`.

The branch was reconciled with the current `dev`; the reviewed fix-forward remains present, and the rebased tree also passes the shared Team-worker provenance/policy-root review for issue #3536.

- Supervised Autopilot parent writes use the canonical locked/revalidated state operation, with changed-field merging and a deterministic concurrent regression.
- Ordinary skill retirement requires the setup-owned receipt and complete unmodified-install digest proof; forged skills-directory receipt/badge content and user-owned files are preserved.
- Shared Team-worker provenance verifies internal identity, external state root, leader CWD, worker worktree, manifest/config agreement, and optional pane identity before an external-root policy binding is accepted. Mismatches remain fail-closed, and protected Team metadata is not granted write authority.
- Focused rebased provenance/doctor/native-hook tests pass: 27 tests. Exact-head terminal CI run `32242380358` is green across required build, typecheck, lint, generated/dist, Rust, coverage, Ralph, smoke, hooks, team-state, and platform lanes.

Verdict: `MERGE_READY`

—

PR #3565, opened 2026-08-24
feat: add --provider launch shorthand with orcarouter as a known provider

kuswardhanietidims-svg:
## Summary

oh-my-codex describes itself as "a workflow layer for OpenAI Codex CLI" that makes it easy to "start a stronger Codex session by default." Today, choosing a non-default model provider means hand-editing `config.toml` — `model_provider = "..."` plus a `[model_providers.<name>]` table and the matching env key. This PR adds a first-class launch shorthand so that same selection can be made from the CLI, and registers **OrcaRouter** as a named provider OMX knows about.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means OMX users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## Changes

- `src/config/providers.ts` (new): a small known-provider registry. `orcarouter` is registered with its env key (`ORCAROUTER_API_KEY`) and base URL (`https://api.orcarouter.ai/v1`), alongside `openai` and `openai-chatgpt`.
- `src/cli/index.ts`: `normalizeCodexLaunchArgs` now consumes a `--provider <name>` / `--provider=<name>` shorthand before the end-of-options marker and emits `-c model_provider="<name>"`, mirroring how `--high` maps to `-c model_reasoning_effort="high"`. Unknown provider names pass through as opaque `model_provider` values owned by `config.toml` — the same way OMX treats any non-alias model name.
- `src/cli/index.ts`: new `omx providers` list command and help-text entries for `--provider` and `omx providers`.
- `src/cli/constants.ts`: new `PROVIDER_FLAG = "--provider"`.
- Tests for the new shorthand (known provider, `=` form, opaque passthrough, combination with `--xhigh`/`--model`, end-of-options preservation, missing value) and for the provider registry.

Because the shorthand normalizes into `-c model_provider=...`, it inherits through the existing team-worker path: `collectInheritableTeamWorkerArgs` already forwards `model_provider` overrides to workers, so `omx --provider orcarouter team ...` selects the same provider for worker panes.

## Validation

- `npm run lint` — pass
- `npm run build` — pass
- `npx tsc --noEmit` — pass
- `npm test` — exit 0 (the 9 pre-existing runner timeouts are tmux/MCP-server integration files that require tmux, which is not installed in this container; none touch the changed files)
- Targeted tests: `dist/config/__tests__/providers.test.js dist/cli/__tests__/index.test.js` — 350/350 pass
- Live check against the registered OrcaRouter endpoint with `ORCAROUTER_API_KEY`:
  - `GET https://api.orcarouter.ai/v1/models` → HTTP 200 (203 models)
  - `POST https://api.orcarouter.ai/v1/chat/completions` (`orcarouter/auto`) → HTTP 200

## Checklist

- [x] Docs updated (CLI help text) when behavior changed
- [x] Backward-compatibility impact considered (unknown provider names remain opaque passthrough; `--provider` was already a value-taking flag in `omx ralph`)

## Related

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -914,6 +914,58 @@ describe("normalizeCodexLaunchArgs", () => {
     ]);
   });
 
+  it("maps --provider orcarouter to a model_provider override", () => {
+    assert.deepEqual(normalizeCodexLaunchArgs(["--provider", "orcarouter"]), [
+      "-c",
+      'model_provider="orcarouter"',
+    ]);
+  });
+
+  it("maps --provider=orcarouter to a model_provider override", () => {
+    assert.deepEqual(normalizeCodexLaunchArgs(["--provider=orcarouter"]), [
+      "-c",
+      'model_provider="orcarouter"',
+    ]);
+  });
+
+  it("passes unknown provider names through as opaque model_provider values", () => {
+    assert.deepEqual(normalizeCodexLaunchArgs(["--provider", "acme-gateway"]), [
+      "-c",
+      'model_provider="acme-gateway"',
+    ]);
+  });
+
+  it("combines --provider with --xhigh and --model", () => {
+    assert.deepEqual(
+      normalizeCodexLaunchArgs([
+        "--provider",
+        "orcarouter",
+        "--xhigh",
+        "--model",
+        "o4-mini",
+      ]),
+      [
+        "--model",
+        "o4-mini",
+        "-c",
+        'model_reasoning_effort="xhigh"',
+        "-c",
+        'model_provider="orcarouter"',
+      ],
+    );
+  });
+
+  it("preserves a literal provider selector after the end-of-options marker", () => {
+    assert.deepEqual(
+      normalizeCodexLaunchArgs(["--provider", "orcarouter", "--", "--provider", "literal"]),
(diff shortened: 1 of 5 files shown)


Yeachan-Heo:
Thanks for the submission. We’re not accepting this provider integration in its current form.

The change adds a new root CLI contract (`--provider`), a new command surface (`omx providers`), provider-specific product messaging, environment-variable semantics, and routing behavior. Those are maintainer-owned product decisions rather than a narrow compatibility fix. The implementation also selects `model_provider="orcarouter"` without creating or owning the corresponding Codex `[model_providers.orcarouter]` configuration, so the shorthand does not by itself deliver the claimed no-manual-configuration workflow.

We are closing this PR rather than leaving a product-direction proposal in review limbo. No maintainer implementation or provider commitment is implied.

—

PR #3625, opened 2026-09-06
fix(doctor/setup): stop relying on removed Codex plugin_hooks flag

Yeachan-Heo:
## Summary

OMX 0.21.3+ probes Codex plugin-scoped hook support by feature name only. Codex CLI 0.153.x still lists `plugin_hooks` in `codex features list` but marks it `removed false`, so:

- `supportsCodexPluginScopedHooks` treated the retired feature as supported,
- plugin-mode setup replaced canonical `[features].hooks = true` with obsolete `plugin_hooks = true` (which Codex reports as removed/false even when explicitly enabled),
- `omx doctor` only recognized `plugin_hooks` in config, so plugin mode was not inferred and plugin-hook diagnostics did not run under canonical `hooks` configuration — producing false "missing global hooks" / "legacy native hook fallback" warnings (issue #3623).

## Fix

- `src/config/codex-feature-flags.ts` — `supportsCodexPluginScopedHooks` now evaluates the full feature row: a `removed` lifecycle never counts as support; any non-removed lifecycle does; absent row means unsupported. No speculative compatibility layers.
- `src/cli/doctor.ts` — `configEnablesPluginScopedHooks` additionally accepts canonical `hooks = true` (root or `[features]`), so plugin mode is inferred and plugin-scoped hook checks run without the removed flag; stale "upgrade Codex to plugin_hooks support" copy replaced.
- Setup on current Codex now keeps the native `hooks.json` fallback contract instead of generating the removed flag. Explicit user configuration and disabled/missing-hook diagnostics are preserved (covered by new tests).

## Testing

- `npm run build`
- `node --test dist/config/__tests__/codex-feature-flags.test.js` (6/6, incl. new removed-flag matrix)
- `node --test dist/cli/__tests__/doctor-warning-copy.test.js` (77/77, incl. new canonical-hooks inference + disabled-hook regression)
- `node --test dist/cli/__tests__/setup-install-mode.test.js` (131/131, incl. new setup flag-generation regression)
- `node --test dist/cli/__tests__/setup-hooks-trust-e2e.test.js`, `plugin-launcher-provenance-3558.test.js`, `dist/config/__tests__/codex-hooks.test.js`, `generator-idempotent/notify` — all green
[...]

PR #3627, opened 2026-09-06
fix: correct Codex hook capability and session diagnostics

### LINKED ISSUE #3623: [Bug] doctor and setup use removed plugin_hooks flag on current Codex
## Summary

OMX 0.21.3 still uses the removed Codex `plugin_hooks` flag to detect and configure plugin hooks. On Codex CLI 0.153.4, this causes misleading doctor warnings and obsolete generated configuration.

The issue surfaced when running `omx doctor` after an Astra-era configuration update. The OMX plugin was already installed and enabled, with `[features] hooks = true` and plugin-bundled hook definitions. No global `hooks.json` was present. Doctor reported missing global hooks, missing prompts, and insufficient skills because it did not recognize the plugin configuration.

## Verified findings

- **Codex recognizes the installation.** Native `hooks/list` returned all seven OMX plugin hooks with no discovery errors or warnings: six enabled and trusted, and Stop trusted but intentionally disabled.
- **Doctor checks the wrong flag.** Using the same plugin cache and its matching Node executable, the exported native-hook checker warns with the canonical configuration. Adding only `plugin_hooks = true` to a temporary config makes the plugin-cache smoke check pass.
- **Setup generates obsolete configuration.** Its feature probe reads `plugin_hooks removed false` as supported because it checks only the feature name. The installed config generator consequently replaces `hooks = true` with `plugin_hooks = true`. Codex reports the latter as removed and false even when explicitly enabled.

Environment: macOS arm64, OMX CLI/plugin 0.21.3, Codex CLI 0.153.4. The controlled checker comparison used the cache's pinned Node 26.5.0 to exclude a separate Node-version mismatch from the result.

## Cause and expected correction

At upstream commit `2da36489cfa07ef1df802f01865e7d959d36f236`:

- [`doctor.ts`](https://github.com/Yeachan-Heo/oh-my-codex/blob/2da36489cfa07ef1df802f01865e7d959d36f236/src/cli/doctor.ts#L2129-L2138) recognizes only `plugin_hooks`, blocking plugin-mode inference and plugin-hook validation under canonical `hooks` configuration.
- [`codex-feature-flags.ts`](https://github.com/Yeachan-Heo/oh-my-codex/blob/2da36489cfa07ef1df802f01865e7d959d36f236/src/config/codex-feature-flags.ts#L58-L63) treats presence of the removed feature name as capability support.

Recognize current plugin-hook support independently of the removed flag, retain/generate the canonical `hooks` setting, and run plugin-aware diagnostics. Regression coverage should include the configuration above and genuinely disabled or missing plugin hooks.

**Conclusion:** this is a confirmed diagnostic/configuration compatibility bug. Runtime hook failure has not been demonstrated. Global `hooks.json` remains supported; its absence alone does not indicate missing plugin hooks. See the [official hook discovery documentation](https://learn.chatgpt.com/docs/hooks#where-codex-looks-for-hooks).

ev78394:
## Summary

Current Codex consolidates plugin hooks under `hooks`, but OMX still checks and generates the removed `plugin_hooks` flag. This causes missing-hooks/prompts/skills warnings for an installed plugin. Doctor also reports a failure when an ordinary Codex session has no OMX session pointer.

## Changes

- Select the effective plugin-hook flag from Codex capabilities; preserve older supported flags, explicit disablement, and canonical/alias precedence.
- Infer plugin installation from trusted enabled plugin registration and generate canonical hook configuration on current Codex.
- Report unavailable runtime binding as a warning only for an absent default-workspace pointer with a lone valid `CODEX_SESSION_ID`. Explicit OMX selectors and unsafe/conflicting pointers still fail.

## Validation

- [x] `npm run build`
- [x] `npm run lint` and `npm run check:no-unused`
- [x] 363 tests passed across the affected doctor, setup, feature-probe, and configuration suites; no skips
- [ ] Full `npm test` suite (not run)
- [x] Installed the packed build and ran user-scoped plugin setup with `--merge-agents --mcp none`
- [x] Installed `omx doctor`: 20 passed, 2 warnings, 0 failures. Remaining warnings are the fresh workspace's absent state directory and explicitly unavailable runtime binding.
- [x] Fresh `omx exec` returned `OMX-EXEC-OK`; SessionStart and UserPromptSubmit hooks completed

Validated on macOS arm64 with Codex CLI 0.153.4. Existing `hooks = true` and intentional Stop-hook disablement were preserved. The local launcher’s separate Node provenance mismatch was resolved by refreshing the plugin under the workspace’s Node executable.

## Checklist

- [x] Changes are limited to hook compatibility and session diagnostics, with regression coverage
- [x] Backward compatibility and explicit disablement considered
- [x] Independent source review completed

Closes #3623

### DIFF
--- a/src/cli/__tests__/doctor-state-root-binding.test.ts
+++ b/src/cli/__tests__/doctor-state-root-binding.test.ts
@@ -29,6 +29,54 @@ function syntheticSnapshot(
 }
 
 describe('doctor state-root/session binding diagnostics', () => {
+  it('reports unavailable runtime binding for only an ambient Codex session in an uninitialized workspace', () => {
+    const check = checkStateRootSessionBinding(
+      syntheticSnapshot('absent', { rootSource: 'cwd-default' }),
+      { CODEX_SESSION_ID: 'codex-session' },
+    );
+    assert.equal(check.status, 'warn');
+    assert.match(check.message, /runtime binding unavailable/);
+    assert.doesNotMatch(check.message, /bad_selectors|clear|relaunch/);
+  });
+
+  it('leaves a fresh workspace untouched when inspecting an ambient Codex session', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-doctor-ambient-session-'));
+    try {
+      const env = { CODEX_SESSION_ID: 'codex-session' };
+      const snapshot = await readCanonicalSessionBindingSnapshot(cwd, env);
+      assert.equal(snapshot.status, 'absent');
+      assert.equal(snapshot.rootSource, 'cwd-default');
+      assert.equal(checkStateRootSessionBinding(snapshot, env).status, 'warn');
+      assert.deepEqual(await readdir(cwd), []);
+    } finally {
+      await rm(cwd, { recursive: true, force: true });
+    }
+  });
+
+  it('keeps explicit runtime selectors and unsafe pointers fail-closed', () => {
+    for (const env of [
+      { CODEX_SESSION_ID: 'codex-session', OMX_SESSION_ID: 'omx-session' },
+      { CODEX_SESSION_ID: 'codex-session', SESSION_ID: 'session' },
+      { CODEX_SESSION_ID: 'codex-session', OMX_ROOT: '/explicit' },
+      { CODEX_SESSION_ID: 'codex-session', OMX_STATE_ROOT: '/explicit' },
+      { CODEX_SESSION_ID: 'codex-session', OMX_TEAM_STATE_ROOT: '/explicit' },
+      { CODEX_SESSION_ID: 'invalid/session' },
+    ]) {
+      assert.equal(checkStateRootSessionBinding(
+        syntheticSnapshot('absent', { rootSource: 'cwd-default' }), env,
+      ).status, 'fail');
+    }
+    for (const status of ['foreign-cwd', 'malformed', 'identity-indeterminate', 'usable'] as const) {
+      assert.equal(checkStateRootSessionBinding(
+        syntheticSnapshot(status, { rootSource: 'cwd-default' }),
+        { CODEX_SESSION_ID: 'unverified-codex' },
+      ).status, 'fail');
+    }
+    assert.equal(checkStateRootSessionBinding(
+      syntheticSnapshot('absent'), { CODEX_SESSION_ID: 'codex-session' },
+    ).status, 'fail');
+  });
+
   it('prefers the winning env root over OMX_SESSION_ID during resolution failure', async () => {
     const env = {
       OMX_ROOT: '/winning-root',
--- a/src/cli/__tests__/doctor-warning-copy.test.ts
+++ b/src/cli/__tests__/doctor-warning-copy.test.ts
@@ -12,8 +12,8 @@ import {
 	symlink,
 	writeFile,
 } from "node:fs/promises";
-import { existsSync, mkdirSync, readFileSync, renameSync, writeFileSync } from "node:fs";
-import { join, dirname } from "node:path";
+import { chmodSync, existsSync, mkdirSync, readFileSync, renameSync, writeFileSync } from "node:fs";
+import { join, dirname, delimiter } from "node:path";
 import { tmpdir } from "node:os";
 import { spawnSync } from "node:child_process";
 import { fileURLToPath } from "node:url";
@@ -29,6 +29,7 @@ import {
 	checkNativeHookDistSmoke,
 	checkNativePostCompactHookRuntime,
 	checkNativeHooks,
+	configEnablesPluginScopedHooks,
 	classifyPostCompactHookStdout,
 	doctor,
 	setDoctorClaimJournalDurabilityForTest,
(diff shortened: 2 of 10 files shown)


Yeachan-Heo:
This PR is being assessed alongside the existing #3625 fix for #3623. The shared hook-capability/configuration changes need one consistent integration path; your additional session-diagnostic behavior is also part of the review and will not be dismissed as a duplicate without checking it.

Review target: `68727391be368fe96096df24e6b8d50af823187e` against `dev` at `0d5568565d1786d7215afc9c050fa73c42bb2741`. Verification includes explicit disablement, capability precedence, duplicate global/plugin hook registration, and retaining failures for unsafe or conflicting session selectors.

No merge or release inclusion is claimed yet.

—

Yeachan-Heo:
Overlap note from the parallel #3625 lane (reviewer/reviewer coordination; no ownership change here):

We independently reproduced the same root cause your PR addresses: with `plugin_hooks` listed as `removed false` alongside canonical `hooks`, OMX 3625 heads before `78465597` emitted global `hooks.json` wrappers on plugin setup while the enabled plugin manifest registers the same hooks — the artifact-level contradiction is confirmed by us too, and we agree with treating manifest registration as the hook surface on removed-flag Codex.

Your PR's direction (effective canonical/alias capability selection driving the surface decision, trusted plugin-registration inference) matches the contract our review converged on. A few concrete observations from our exact-head diff review of `68727391`:

- Your `resolveCodexPluginHookFeatureFlag` returning `null` when only a `hooks removed` row (no canonical row) exists is a reasonable strictness refinement over our three-state classifier; worth keeping explicit test coverage for that degenerate listing.
- Your doctor-side warning-only unavailable-binding change for lone `CODEX_SESSION_ID` is genuinely outside our scope and should not be lost in supersession discussions.
- If #3627 lands first, our lane will rebase and reconcile; our uniquely remaining delta would be limited to uninstall-time `plugin_hooks` cleanup/migration coverage and legacy `codex_hooks`-only fallback regressions — happy to contribute either to your branch if the lane owners prefer.

CI on this branch is fully green (19 success / 9 skipped) as of this comment.

—

Yeachan-Heo:
## Adversarial review: fix: align hook diagnostics with current Codex capabilities

**Reviewed head:** `68727391be368fe96096df24e6b8d50af823187e` (`fix/doctor-hooks-session-diagnostics`, fork `ev78394`)
**Base:** `dev` @ `0d5568565d1786d7215afc9c050fa73c42bb2741`
**Diff:** 10 files, +297/−20, diff digest `83e9a44db9a4c3a0ad021e3617fec1efd38878c40ad750dbcdef85a4013774fb`

### Independent verification at exact head (isolated temp `HOME`/`CODEX_HOME`/`XDG_CONFIG_HOME` throughout; no host configuration touched)

- `npm run build`, `biome lint` (836 files), `tsc --noEmit`, and `check:no-unused` are all clean.
- Focused suites re-run: `codex-feature-flags` 6/6, `generator-notify` 25/25, `doctor-state-root-binding` 11/11, `doctor-warning-copy` 79/79, `setup-install-mode` 135/135 — consistent with the reported 363-test run across the affected suites.
- CI refreshed at this head: 19 success / 9 skipped (publish + platform-native + docs lanes) / 0 failed; the previously pending Node and coverage lanes are now green. Mergeable/CLEAN. Any push from either branch invalidates this review.

### Feature capability resolution (20+ adversarial probe shapes)

- `resolveCodexPluginHookFeatureFlag` verified against CRLF, aligned columns, `experimental`/`stable`/`deprecated`/`removed` lifecycles, `removed true`/`false`, both-rows-removed, name collisions (`xplugin_hooks`), spinner-prefixed rows, and null/empty output. A removed `plugin_hooks` row selects canonical `hooks` only when the `hooks` row exists and is itself not removed; otherwise the resolver returns null and setup falls back to the hooks.json path (fail-safe). Builds still reporting a live `plugin_hooks` lifecycle keep the proven legacy contract.
- Lifecycle-less `plugin_hooks true` rows no longer count as support (the stage regex requires the lifecycle column), closing the over-report gap flagged in the #3625 review thread.
- `supportsCodexPluginScopedHooks` is now a thin mirror of the resolver; uninstall and other consumers use the unchanged canonical-flag surface, so no stale call sites.

### Enablement inference and dual-dispatch surface (live E2E, not name-only)

- `configEnablesPluginScopedHooks` under canonical mode requires a trusted, enabled `oh-my-codex` plugin registration with hooks not explicitly disabled (absent key = default-enabled, matching the `hooks stable true` row); `hooks = false` still warns as legacy fallback and is never claimed as a plugin-scoped pass. `hooks` > `codex_hooks` precedence matches `resolveCodexHookFeatureFlag`; foreign plugin registrations do not leak; broken TOML returns false (fail-safe).
- Live E2E against the built CLI with a removed-flag probe: a stale OMX-only global wrapper is removed leaving a single plugin-scoped surface with exactly one `hooks = true` and zero `plugin_hooks`/`codex_hooks` lines; a foreign hooks group plus a stale wrapper keeps the foreign entries verbatim while removing the wrapper; the issue #3623 configuration (no hooks.json) gets the plugin-scoped surface only. Double setup is idempotent in all cases and user `model`/`[profiles.*]` entries survive. The pre-existing `unsafe_managed_removal` coordinate guard still refuses coordinate-shifting removals fail-closed.
- Doctor's plugin-scoped branch validates real plugin-cache content (manifest, `./hooks/hooks.json` pointer, pinned launcher compatibility, packaged match) — the runtime installation surface is genuinely validated, and the legacy surface (`plugin_hooks experimental` probe → `plugin_hooks = true`) remains test-locked.

### Session-binding downgrade (unique scope) — live boundary probes

Warn-only requires all of: absent pointer, cwd-default root, no `OMX_ROOT`/`OMX_STATE_ROOT`/`OMX_TEAM_STATE_ROOT`, exactly one nonblank selector equal to a normalized-valid `CODEX_SESSION_ID`. Verified live: a lone valid ambient ID warns with honest "runtime binding unavailable / no runtime authority verified" copy; adding `OMX_SESSION_ID` fails; a path-shaped invalid ID fails; adding `OMX_ROOT` fails with the unsafe-root disclosure fields; a true foreign-cwd pointer fails; a malformed pointer fails; a stale pointer with a matching dead session keeps unchanged warn semantics; a fresh workspace is left byte-for-byte untouched. Nothing invalid becomes valid — the change strictly converts one false-fail (an ambient Codex session in an uninitialized workspace) into a warning and never below warn.

### Overlap with #3625 (same base, same issue)

#3627 and #3625-final (`0935f1283f845a84016fa095d74b9e49672fa260`) touch the same feature-flag, generator, doctor, and setup files plus the same tests, but are not interchangeable:

- #3625 deletes the legacy generation path entirely (always emits the canonical flag); #3627 capability-matches the emitted flag to the installed Codex, keeping the proven `plugin_hooks` contract for builds that still report it.
- #3627's resolver is strictly stricter on ambiguous probe rows, its canonical-enablement inference is trust-gated, and its generator keeps the legacy dual-flag option parameterized with a guard against duplicate identical flag lines.
- #3625's initial MERGE_READY missed the dual-dispatch class; #3627 handles the wrapper-removal/foreign-preservation surface correctly, re-verified E2E above.

**Overlap/salvage recommendation:** adopt #3627 as the canonical fix for #3623 and close #3625 in its favor rather than merging both — merging both would double-rewrite the same generator contract and strand one of the two legacy-flag strategies. #3625's fix-forward deserves explicit credit: it identified the dual-dispatch class this review independently re-verified against #3627.

### Non-blocking notes

- Parse-broken legacy configs (`plugin_hooks = true` plus a syntax error) lose plugin-scoped diagnostics (the function now returns false instead of the old regex fallback); the direction is fail-safe since the legacy hooks.json checks still run.
- The warn-only binding line intentionally omits the extended static diagnostic fields; it is report-only and test-locked, and future consumers keying on `fix=`/`no-mutation` fields should treat it as a distinct minimal form.

No blocking defects found. No merge, close, or release action was taken from this lane.

MERGE_READY

—

PR #3622, opened 2026-09-04
feat(models): default all OMX agents to Astra, preserving effort and overrides

ev78394:
## Summary

Default OMX leaders, specialists, fast agents, low-complexity workers, and Sparkshell summaries to `gpt-6-astra`. Preserve existing reasoning-effort defaults and explicit model configuration, profiles, per-agent overrides, and CLI/environment choices.

Normal contribution targeting `dev`, per the repository contribution guide.

## Changes

- Set frontier, standard, spark, team-child, exact planning/research, new-agent, and subscription model defaults to Astra. Keep GPT-5.6 alternatives recognized and preserve opaque provider-specific overrides.
- Seed a missing root model during plugin setup, as legacy setup already does. Preserve existing model/profile/provider values and custom multiline instructions; keep reasoning-effort behavior unchanged.
- Keep cheap-model planner detection tied to Terra/Luna rather than default-model constants, so Astra is not classified as a cheap model.
- Default Sparkshell primary/fallback models to Astra while preserving low reasoning effort and configurable alternatives. A distinct fallback remains opt-in.
- Update routing documentation and regressions for all 30 built-in roles, effort defaults, overrides, plugin setup, and worker identity. Include captured status/output in raw-byte test failures without weakening assertions.

## Default behavior changes and rationale

This intentionally changes OMX's default model policy while preserving its orchestration structure, role responsibilities, reasoning-effort defaults, and explicit model overrides.

| Behavior | Before | After |
| --- | --- | --- |
| Frontier / standard / fast model defaults | Sol / Terra / Luna | Astra / Astra / Astra, including fast agents and low-complexity workers |
| Sparkshell primary / fallback | Luna / Terra | Astra / Astra; an alternate-model fallback requires explicit configuration |
| Plugin setup with no root `model` | Leaves the root model unset | Seeds `gpt-6-astra`; existing model and profile choices retain precedence |

**Why choose these defaults?** Astra across roles makes the default model selection consistent and gives delegated work access to the same model family as the leader. [OpenAI's model guidance](https://learn.chatgpt.com/docs/models) positions Astra for demanding workflows involving sustained reasoning, multiple steps, and tools. That supports Astra as a reasonable starting point when output quality is the priority. Seeding a missing plugin root model also makes fresh setup reproduce the declared OMX default instead of depending on the client's recommended model.

[...]

PR #3631, opened 2026-09-06
fix(setup): preserve user-owned root model_reasoning_effort across setup/update

### LINKED ISSUE #3630: [Bug] omx setup resets model_reasoning_effort to "medium", overwriting user config and omx reasoning
## Summary

`omx setup` rewrites the root `model_reasoning_effort` in `config.toml` back to
`"medium"`, discarding whatever value is configured — including a value that was
just set through the documented command `omx reasoning <mode>`.

Because `omx update` runs setup, there is currently no supported way to keep a
root reasoning effort other than `medium`: it survives ordinary use, but is
silently reverted by the next setup or update.

### Relation to #3447 / #3448

This is the `model_reasoning_effort` half of #3447. PR #3448 ("preserve user
notify/effort and project scope during launch repair", merged 2026-08-06) made
`repairConfigIfNeeded` surgical, and I can confirm that path is fixed: ordinary
runs (`omx doctor`, `omx exec`) leave a non-default value untouched on 0.21.3.

The **setup path** was not covered. `omx setup` still rewrites the key, so the
value is lost on the next `omx update` even though launch repair now preserves
it. This report is about that remaining path, not a re-report of #3447.

## Environment

- OS: Ubuntu 26.04.1 LTS (x86_64)
- Node.js: v22.22.1
- oh-my-codex: 0.21.3
- Codex CLI: codex-cli 0.153.4
- Scope: user (`omx setup --scope user --merge-agents`)

## Steps to reproduce

1. `omx reasoning high`
   -> prints `Set model_reasoning_effort="high" in ~/.codex/config.toml`
2. `omx doctor`, `omx exec ...`  (any normal use)
   -> value is still `"high"`
3. `omx setup --scope user --merge-agents`
   -> setup summary reports `config: updated=1, unchanged=1`
4. `grep model_reasoning_effort ~/.codex/config.toml`

## Expected behavior

Setup preserves a configured reasoning effort, the same way it preserves other
root keys. `omx reasoning` should not be undone by `omx update`.

## Actual behavior

```
before setup : model_reasoning_effort = "high"
after  setup : model_reasoning_effort = "medium"
```

## Evidence

Same config file, same shell, values restored between runs:

| step | model_reasoning_effort |
|---|---|
| `omx reasoning high` | `"high"` |
| `omx doctor` | `"high"` (untouched) |
| `omx exec ...` | `"high"` (untouched) |
| `omx setup --scope user --merge-agents` | **`"medium"`** |

The rewrite is specific to this key. A control run with a deliberately
non-default `approval_policy = "untrusted"` came through the same setup
unchanged, so setup is not simply resetting the whole managed block.

## Impact

- `omx reasoning` is the documented way to set this value, and its effect does
  not survive an update.
- The failure is silent: the setup summary reports `config: updated=1` without
  naming the key it overwrote.
- It is invisible to anyone whose chosen value already is `medium` — the
  overwrite and the preservation look identical.

## Suggested fix

1. Treat an existing root `model_reasoning_effort` as user-owned and leave it
   alone, as setup already does for other root keys.
2. If setup must normalize the value, report it explicitly in the summary
   ("model_reasoning_effort: high -> medium") instead of the generic
   `config: updated=1`.
3. Alternatively, persist the value chosen via `omx reasoning` in a place setup
   honors.

Yeachan-Heo:
## Summary

`omx setup` rewrote the root `model_reasoning_effort` in `config.toml` to the
managed default `"medium"` on every legacy-scope config refresh, silently
discarding values set through direct edits or the documented
`omx reasoning <mode>` command. Because `omx update` runs setup, no
non-default root reasoning effort could survive an upgrade.

Fixes #3630.

## Root cause

In `buildMergedConfig` (`src/config/generator.ts`), the managed refresh
unconditionally strips the OMX top-level keys (`notify`,
`model_reasoning_effort`, `developer_instructions`) and re-emits
`model_reasoning_effort = "medium"` in `getOmxTopLevelLines`. `notify` had a
preserve path; this key did not.

## Fix

- `MergeOptions` gains `preserveReasoningEffort`. When set, an existing root
  `model_reasoning_effort` is captured before the managed block is stripped
  and re-emitted verbatim (byte-for-byte, including the raw TOML spelling).
- When the key is absent, the managed `"medium"` default is still seeded, so
  fresh installs keep the intended default.
- Setup passes `preserveReasoningEffort: true` in `planManagedConfig`
  (legacy managed path). Plugin mode never wrote this key and is unchanged.
- Invalid values are passed through unvalidated and remain user-owned;
  setup already refuses to write a config that fails TOML parsing
  (`Refusing to write invalid planned config.toml`). This keeps honest
  invalid-value handling: setup neither silently "fixes" the value nor
  crashes on it beyond existing parse validation.

The existing Astra default agent-level reasoning overrides (#3622) are
untouched: this change only covers the root key in `config.toml`.

## Verification

All in isolated temp `HOME` / `XDG_CONFIG_HOME` / `CODEX_HOME`, no host
config touched:

- Reproduced on current dev (`c81edfa5`): `omx reasoning high` → `omx setup`
  → reverted to `"medium"` with `config: updated=1` (silent). Same after
  `omx doctor`/normal use left it untouched.
- After the fix: `omx reasoning xhigh` survives repeated setup; direct
  hand-edit to `"low"` survives setup; `omx doctor` preserves; fresh setup
  still seeds `"medium"`; project-scope setup leaves the user config alone;
  `omx reasoning` (status) still reports the preserved value.
- `[profiles.*].model_reasoning_effort` values remain intact (no root-level
  key leakage into tables; exactly one root key after refresh).
- New tests: `src/cli/__tests__/setup-reasoning-preserve.test.ts` (9 tests)
  pass; generator suites, native-config, setup-install-mode,
  setup-prompts-overwrite, uninstall, index/exec/launch-fallback suites pass;
  the only failing tests in the full local run (tmux/team lifecycle) fail
  identically on pristine `dev` in this environment and are untouched by
  this change.

Document-refresh: not-needed | no documented setup behavior contract changed; `omx reasoning` semantics preserved and extended to survive updates.

*Signed-off-by: Yeachan-Heo <Yeachan-Heo@users.noreply.github.com>*

### DIFF
--- a/src/cli/__tests__/setup-reasoning-preserve.test.ts
+++ b/src/cli/__tests__/setup-reasoning-preserve.test.ts
@@ -1,0 +1,248 @@
+import assert from "node:assert/strict";
+import { after, describe, it } from "node:test";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { existsSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import TOML from "@iarna/toml";
+import { buildMergedConfig } from "../../config/generator.js";
+import { setup } from "../setup.js";
+import { upsertTopLevelTomlString } from "../../utils/toml.js";
+
+/**
+ * Regression coverage for issue #3630:
+ * `omx setup` must preserve an existing root `model_reasoning_effort`
+ * (explicit user edit or `omx reasoning <mode>`) while still seeding the
+ * `medium` default when the key is absent.
+ */
+
+interface IsolatedHome {
+	wd: string;
+	home: string;
+	codexHome: string;
+	configPath: string;
+	restore: () => void;
+}
+
+
+async function createIsolatedHome(prefix: string): Promise<IsolatedHome> {
+	const wd = await mkdtemp(join(tmpdir(), `${prefix}-`));
+	const home = join(wd, "home");
+	const codexHome = join(home, ".codex");
+	await mkdirRecursive(codexHome);
+	const previous = {
+		HOME: process.env.HOME,
+		CODEX_HOME: process.env.CODEX_HOME,
+		XDG_CONFIG_HOME: process.env.XDG_CONFIG_HOME,
+	};
+	process.env.HOME = home;
+	process.env.CODEX_HOME = codexHome;
+	process.env.XDG_CONFIG_HOME = join(home, ".config");
+	return {
+		wd,
+		home,
+		codexHome,
+		configPath: join(codexHome, "config.toml"),
+		restore: () => {
+			if (previous.HOME !== undefined) process.env.HOME = previous.HOME;
+			else delete process.env.HOME;
+			if (previous.CODEX_HOME !== undefined) {
+				process.env.CODEX_HOME = previous.CODEX_HOME;
+			} else {
+				delete process.env.CODEX_HOME;
+			}
+			if (previous.XDG_CONFIG_HOME !== undefined) {
+				process.env.XDG_CONFIG_HOME = previous.XDG_CONFIG_HOME;
+			} else {
+				delete process.env.XDG_CONFIG_HOME;
+			}
+		},
+	};
+}
+
+async function mkdirRecursive(dir: string): Promise<void> {
+	const { mkdir } = await import("node:fs/promises");
+	await mkdir(dir, { recursive: true });
+}
+
+const MANAGED_EFFORT_DEFAULT = "medium";
+
+describe("buildMergedConfig reasoning effort ownership (issue #3630)", () => {
+	it("does not preserve by default (legacy callers keep the managed default)", () => {
+		const existing = [
+			"model_reasoning_effort = \"high\"",
+			"",
+			"[features]",
+			"hooks = true",
+			"",
+		].join("\n");
+		const merged = buildMergedConfig(existing, "/tmp/omx-test-root", {
+			includeTui: false,
+		});
+		const parsed = TOML.parse(merged) as { model_reasoning_effort?: string };
+		assert.equal(parsed.model_reasoning_effort, MANAGED_EFFORT_DEFAULT);
+	});
+
+	it("preserves an existing non-default root effort when asked", () => {
+		const existing = [
+			"model_reasoning_effort = \"xhigh\"",
+			"",
+			"[features]",
+			"hooks = true",
+			"",
+		].join("\n");
+		const merged = buildMergedConfig(existing, "/tmp/omx-test-root", {
+			includeTui: false,
+			preserveReasoningEffort: true,
+		});
+		const parsed = TOML.parse(merged) as { model_reasoning_effort?: string };
+		assert.equal(parsed.model_reasoning_effort, "xhigh");
+	});
+
+	it("preserves the exact raw TOML value, including invalid values (honest hand-off)", () => {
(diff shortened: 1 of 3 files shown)


Yeachan-Heo:
Independent review assigned at `5ea7f5ed85252511661f4b7ded60b7b390cd0a17` against `dev@c81edfa599f099bb91fe80e93c6d41497c349338`. Review scope includes setup/update and reasoning-command round trips, root/profile boundaries, raw TOML preservation, missing defaults, and unrelated configuration stability. No merge-readiness claim yet.

—

Yeachan-Heo:
**MERGE_READY** — independent adversarial review of head 5ea7f5ed against base c81edfa5 passed; no regression found. (Review verdict posted as an issue comment because GitHub blocks review submission from the PR author account.)

- Scope: 3 files, +272/−1 (generator managed-merge option, setup opt-in wire, regression tests). No unrelated contracts touched.
- Root cause in `buildMergedConfig` (`getOmxTopLevelLines` unconditionally re-emitting `"medium"`) is correctly diagnosed; fix captures the raw root `model_reasoning_effort` before `stripOmxTopLevelKeys` and re-emits it byte-for-byte only when `preserveReasoningEffort: true` (setup-only). Absence still seeds `"medium"`; plugin mode and all other `buildMergedConfig` callers unchanged; `repairConfigIfNeeded` stays surgical.
- Independently verified against the built head, isolated HOME/XDG_CONFIG_HOME/CODEX_HOME only: repeated setup after `omx reasoning xhigh` keeps `xhigh`; hand-edited `"high"` survives; fresh install seeds `medium`; `[profiles.deep]` effort and model intact; managed refresh stays single-key idempotent (re-setup on managed output keeps exactly one effort line); user `notify` untouched; base-vs-head comparison confirms base collapsed `xhigh`→`medium` and head preserves it.
- Adversarial probes all behaved honestly: duplicate root keys dedupe to the last entry (no invalid duplicate emitted); multiline basic strings, literal-quote spellings, comments, non-string (integer) values pass through verbatim; unterminated multiline string survives the merge and is refused by setup's `TOML.parse(finalConfig)` gate — invalid TOML is never silently normalized; quoted-key (`"model_reasoning_effort" =`) spellings are intentionally not recognized and still seed `medium` (canonical spellings only — acceptable, not a regression vs base); CRLF inputs merge cleanly; `--disable-hooks` and plugin-mode install never clobber a user effort; uninstall still removes the key as OMX-owned (pre-existing contract, unchanged).
- Independent runs: new regression suite 9/9 pass; generator-idempotent 103, generator-notify 25, setup-refresh 35, setup-install-mode 135, setup-scope 14, update 82, uninstall 80, models 36, codex-feature-flags 6, setup-hooks-shared-ownership 7 — all pass; setup-hooks-trust-e2e 2 skipped (missing installed-Codex binary, pre-existing lane condition, same on base). Typecheck/Lint/Build and all CI lanes green at head except one lane still in progress at review time (Test Node 20 / cli-core-rest); GitHub reports the branch MERGEABLE.
- Digests: head 5ea7f5ed85252511661f4b7ded60b7b390cd0a17, base c81edfa599f099bb91fe80e93c6d41497c349338, diff sha256 7e99c6a56bcdcdac54cc2454e24792c7c8cc74f2aaae44e785663c0ddb3507ca.

Issue #3600, opened 2026-08-31
[Bug] macOS arm64: omx-runtime is still unavailable after installing oh-my-codex 0.21.0


kaidomo:
## Summary

I installed `oh-my-codex@0.21.0` globally on macOS arm64 and re-tested
stale session pointer recovery using a completely fresh temporary Git
repository.

The `omx-runtime` native binary is still unavailable after installation.
As a result, stale pointer recovery returns `unsupported`, although the
underlying APFS volume supports the required exclusive rename operation.

The v0.21.0 release notes state that macOS arm64 `omx-runtime` hydration
and discovery were fixed, but the following clean installation still
exhibits the problem. This may indicate an incomplete fix or a problem
in the published npm installation path.

## Environment

- OS: macOS
- Architecture: arm64
- oh-my-codex: 0.21.0
- Node.js: 22.22.3
- Installation: global npm installation
- Filesystem: APFS

## Reproduction

Use a fresh temporary directory so that no project-specific files,
configuration, or dependencies are involved.

```sh
mkdir omx-runtime-repro
cd omx-runtime-repro
git init

npm install -g oh-my-codex@0.21.0
omx setup --plugin --force
omx --version
omx doctor
```

Then:

1. Start an OMX session in the fresh repository.
2. Terminate the session owner process abruptly.
3. Run:

```sh
omx session pointer recover --cwd "$PWD" --json
```

## Actual result

`omx doctor` reports failures including:

```text
native process identity unavailable
state root/session binding: ptr=stale-dead
```

Pointer recovery returns:

```json
{
  "status": "unsupported",
  "recovered": false,
  "reason": "Atomic no-replace recovery rename is unsupported on this platform."
}
```

The native runtime is also not executable:

```sh
which omx-runtime
omx-runtime --help
```

`omx-runtime` is not found, and the managed native cache does not
contain the expected v0.21.0 macOS arm64 runtime binary.

## Filesystem verification

The APFS volume was checked independently:

- macOS Foundation reports support for exclusive renaming.
- The volume supports the `RENAME_EXCL` operation used by the native
  runtime.

This suggests that the error is not caused by the filesystem lacking
atomic no-replace rename support.

## Expected result

After a clean global npm installation:

- the matching macOS arm64 `omx-runtime` should be hydrated or made
  available;
- the bridge should discover and execute it;
- `omx doctor` should pass the native process identity check;
- stale session pointer recovery should succeed, or report the actual
[...]

PR #3527, opened 2026-08-14
fix(session): recover verified-dead pointers

Yeachan-Heo:
## Summary
- add the official `omx session pointer recover [--cwd <path>] [--json]` command
- quarantine only an authoritative, positively dead selected pointer while preserving exact bytes
- keep live, reused, identity-indeterminate, malformed, foreign-lineage/root, I/O, collision, unsupported, and concurrent-change states fail-closed
- guide stale-pointer recovery through doctor/help/troubleshooting without changing lock recovery or native Stop contracts

## Verification
- `npm run build`
- `npm run check:no-unused`
- `npx biome lint` on all changed TypeScript files
- `node --test dist/hooks/__tests__/session.test.js` — 142 passed on the final full focused run; final recovery subset 20/20 after the last race hardening
- `node --test dist/cli/__tests__/session-search-help.test.js dist/cli/__tests__/doctor-state-root-binding.test.js` — 9 passed
- `node --test dist/cli/__tests__/launch-fallback.test.js` — 36 passed
- `node --test --test-name-pattern='Stop transcript-backed recovery for a stale-dead selected pointer' dist/scripts/__tests__/codex-native-hook.test.js` — 20 passed
- `node --test dist/cli/__tests__/package-bin-contract.test.js` — 1 passed
- native runtime CLI smoke proved exact archive bytes and idempotent repeat
- final architect review: CLEAR/CLEAR/CLEAR, APPROVE
- final executor red-team QA: passed

Closes #3525

—
*[repo owner's gaebal-gajae (clawdbot) 🦞]*

PR #3638, opened 2026-09-07
fix(session): portable pointer recovery without native runtime (#3636)

### LINKED ISSUE #3636: [Bug] omx-runtime still not hydrated on macOS arm64 in 0.21.3 — `session pointer recover` returns `unsupported` (follow-up to #3600)
## Summary

`omx-runtime` is still not hydrated after a clean global npm install on macOS arm64 in
**0.21.3**, so `omx session pointer recover` returns `unsupported` and the recovery path
documented in `docs/troubleshooting.md` cannot be used on this platform.

This is a follow-up to **#3600** (same symptom, reported on 0.21.0, closed 2026-08-31). It
still reproduces two releases later. Because the missing runtime is what blocks the recovery
command, the resolution of **#3525** — whose fix (PR #3527) shipped exactly this command — does
not hold on macOS.

## Environment

- OS: macOS 26.6.2, arm64 (APFS)
- Node.js v24.19.0, npm 11.17.0
- oh-my-codex **0.21.3** (= current `latest` on the registry), global npm install
- codex-cli 0.153.0

## What I measured

`omx doctor`:

```
[XX] Process identity: native process identity is unavailable (provider unavailable);
     reinstall or update OMX and rerun doctor
```

The advice cannot help: the package is already at the head of the registry, and a reinstall
does not change the outcome (this is what #3600 established on a fresh install).

`omx session pointer recover`, on a stale-dead pointer that satisfies **every** precondition the
documentation requires — owner PID definitively dead (`ps -p <pid>` → rc=1, positive control on a
live PID → rc=0), PID not reused, pointer `cwd` containing the target directory, pointer
`state_root` canonical-matching the selected state root:

```console
$ omx session pointer recover --cwd /Users/drg --json ; echo "rc=$?"
{
  "status": "unsupported",
  "pointerPath": "/Users/drg/.omx/state/session.json",
  "action": "none",
  "recovered": false,
  "reason": "Atomic no-replace recovery rename is unsupported on this platform.",
  "sessionId": "01a06685-…"
}
rc=1
```

The underlying volume is APFS and does support an exclusive rename (`renameatx_np` with
`RENAME_EXCL`), so this is a missing-runtime problem rather than a filesystem capability
problem — consistent with #3600.

## What the installed package contains

```console
$ ls ~/.npm-global/lib/node_modules/oh-my-codex/crates/
omx-runtime  omx-runtime-core   # Rust sources

$ find ~/.npm-global/lib/node_modules/oh-my-codex -name '*omx-runtime*' -type f
# no compiled binary

$ jq '.optionalDependencies' ~/.npm-global/lib/node_modules/oh-my-codex/package.json
null
```

So no per-platform prebuilt package is pulled in, and nothing in the install path hydrates the
archive that `dist/cli/__tests__/native-assets.test.js` describes
(`omx-runtime-<target>.zip` → `binary: omx-runtime`). The only route that mentions building it is
`npm run build:runtime` (`cargo build -p omx-runtime`), which a user of the published package has
no reason to run.

## Impact

`docs/troubleshooting.md` § *Exact-session reconciliation (stale-dead pointer)* presents
`omx session pointer recover --cwd <path>` as the official way out and states that a successful
recovery "leaves the canonical pointer absent so an ordinary relaunch can publish a fresh
pointer". On macOS that command cannot run, so the only way out is the manual file surgery the
same document advises against. Every `omx` launch in the affected cwd keeps aborting with
`session_pointer_unusable` / `stale-dead` until the user removes the pointer by hand.

## Expected

Either the macOS arm64 `omx-runtime` asset is hydrated by the published install path, or
`session pointer recover` falls back to a portable implementation (for example
`link()` + `unlink()`, which gives the same no-clobber guarantee without `renameat2`/
`renameatx_np`), so the documented recovery works on the platforms the docs address.

## Workaround currently in use

Reproducing by hand what the command is documented to do — an adjacent, no-clobber quarantine
that preserves the forensic bytes rather than deleting them:

```zsh
S=<root>/.omx/state/session.json
mv -n "$S" "$S.recovered-$(date +%Y%m%dT%H%M%S)"
```

Oracle afterwards: `omx doctor` reports `ptr=absent` on the *State root/session binding* line.

Yeachan-Heo:
Fixes #3636 (supsersedes the #3600 symptom: same unhydrated-runtime failure, two releases later). Also invalidates the premise that PR #3527's `session pointer recover` holds on macOS — it could not run there.

## Root cause

1. **Recovery dead-ends without the native binary.** `moveRecoveryPathNoReplace` (`src/hooks/session.ts:1591`) returned `{ unsupported: true, reason: 'Atomic no-replace recovery rename is unsupported on this platform.' }` whenever `defaultRecoveryRenameNoReplace` (`src/hooks/session.ts:579`) could not spawn `omx-runtime fs-rename-no-replace` — and `recoverSessionPointerLock` (`src/hooks/session.ts:1771`, since removed) probed the same capability and refused before writing any checkpoint. On the reporter's machine the binary was never hydrated, so `omx session pointer recover` returned `status: "unsupported"` on a pointer satisfying every documented precondition.
2. **The published tarball can never contain a macOS arm64 `omx-runtime`.** `package.json` `files` (`package.json:87`) ships `crates/` as Rust sources only; the binary hydrates at install/update time from the GitHub release manifest into the version-keyed native cache (`src/cli/native-assets.ts:685` `hydrateNativeBinary`). The `aarch64-apple-darwin` asset exists in the v0.21.3 release, so hydration works when it runs — I hydrated it from Linux as proof (`Mach-O 64-bit arm64 executable`).
3. **`omx update` was a no-op for this failure mode.** The up-to-date branch (`src/cli/update.ts:843`) returned before `hydrateRuntime`, so reinstalling 0.21.3 over 0.21.3 never repaired the empty version-keyed cache — exactly the reporter's comment #1.
4. **Doctor prescribed the no-op.** `checkProcessIdentityReadiness` (`src/cli/doctor.ts:1388`) told every macOS/Windows user with an unavailable provider to "reinstall or update OMX and rerun doctor", which changes nothing when hydration never ran.

## Fix

- **Portable fallback (reporter's `link()`+`unlink()` suggestion, credited):** `moveRecoveryPathNoReplace` now falls back to `moveRecoveryPathPortableNoReplace` (`src/hooks/session.ts:1548`) when the native provider reports `unsupported`. Hard-linking to the token-bound quarantine path fails with `EEXIST` on collision — the same no-clobber guarantee — then unlinking the source leaves the canonical pointer absent with forensic bytes preserved. Directory-incarnation lock recovery still fails closed before checkpoint residue (probe at `src/hooks/session.ts:1753`), since `link()` cannot move directories.
- **Update repairs the cache:** the up-to-date branch now calls `hydrateRuntime` on both immediate and passive paths (`src/cli/update.ts:861`).
- **Truthful doctor:** the `unsupported` verdict now says `run `omx update` (repairs the native runtime cache) or `omx setup`, then rerun doctor; session pointer recovery works without the native runtime` (`src/cli/doctor.ts:1390`).

## Verification

```
npm run build                              # clean
npx tsc --noEmit                           # clean
npx tsc -p tsconfig.no-unused.json         # clean
npx biome lint src docs                    # 838 files, no findings
node --test dist/hooks/__tests__/session.test.js                          # pass 152 fail 0
node --test dist/cli/__tests__/update.test.js                             # pass 83 fail 0
node --test dist/cli/__tests__/doctor-process-identity-readiness.test.js  # pass 6 fail 0
node --test dist/cli/__tests__/native-assets.test.js                      # pass 18 fail 0
```

New regression tests: portable recovery succeeds with the native provider forced to `unsupported`; existing quarantine destination is never overwritten (`collision`); lock-directory quarantine fails closed without residue; passive up-to-date `omx update` hydrates the runtime cache. End-to-end manual run with `atomicRenameNoReplace: () => 'unsupported'`: `status: recovered`, canonical absent, quarantine bytes exact, fresh relaunch publishes.

## Platform-unverified

No macOS arm64 host was available: the `renamex_np(RENAME_EXCL)` native path and the real `omx update` hydration on macOS are unverified. The portable fallback is platform-agnostic Node (`fs.link`/`fs.unlink`) and covered by the regression tests above, but a Mac run of `omx session pointer recover --cwd <path> --json` and `omx doctor` after `omx update` is still wanted.

—

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -99,7 +99,7 @@ To remove a verified-dead selected pointer and preserve its exact forensic bytes
 omx session pointer recover --cwd /path/to/project
 ```
 
-The command serializes with the canonical pointer lock and moves only an authoritative, positively dead `session.json` to an adjacent no-clobber quarantine path. It refuses live or usable owners, uncertain or reused identities, malformed or foreign pointers, root/lineage mismatches, concurrent changes, I/O failures, and existing quarantine destinations. A successful recovery leaves the canonical pointer absent so an ordinary relaunch can publish a fresh pointer.
+The command serializes with the canonical pointer lock and moves only an authoritative, positively dead `session.json` to an adjacent no-clobber quarantine path. It refuses live or usable owners, uncertain or reused identities, malformed or foreign pointers, root/lineage mismatches, concurrent changes, I/O failures, and existing quarantine destinations. A successful recovery leaves the canonical pointer absent so an ordinary relaunch can publish a fresh pointer. No native runtime is required: when the atomic no-replace rename provider is unavailable (for example an unhydrated `omx-runtime` on macOS arm64), recovery falls back to a portable `link()` + `unlink()` quarantine with the same no-clobber guarantee, so the documented command works on every platform the docs address.
 
 When the pointer's recorded PID is definitively dead, bind the exact current session explicitly and retry the same command — no manual file surgery:
 
--- a/src/cli/__tests__/doctor-process-identity-readiness.test.ts
+++ b/src/cli/__tests__/doctor-process-identity-readiness.test.ts
@@ -60,7 +60,6 @@ describe('doctor native process identity readiness', () => {
     assert.equal(check, null);
     assert.equal(called, false);
   });
-
   it('fails safely when the native runtime is missing or unsupported', () => {
     const check = checkProcessIdentityReadiness({
       platform: 'win32',
@@ -70,7 +69,7 @@ describe('doctor native process identity readiness', () => {
     assert.equal(check?.status, 'fail');
     assert.equal(
       check?.message,
-      'native process identity is unavailable (provider unavailable); reinstall or update OMX and rerun doctor',
+      'native process identity is unavailable (provider unavailable); run `omx update` (repairs the native runtime cache) or `omx setup`, then rerun doctor; session pointer recovery works without the native runtime',
     );
   });
 
@@ -85,7 +84,6 @@ describe('doctor native process identity readiness', () => {
       assert.match(check?.message ?? '', /reinstall or update OMX and rerun doctor$/);
     }
   });
-
   it('rejects foreign-platform identity evidence without trusting or repairing it', () => {
     const check = checkProcessIdentityReadiness({
       platform: 'darwin',
--- a/src/cli/__tests__/update.test.ts
+++ b/src/cli/__tests__/update.test.ts
@@ -498,6 +498,54 @@ describe('maybeCheckAndPromptUpdate', () => {
     }
   });
 
+  it('repairs the native runtime cache on the passive up-to-date path (issue #3636)', async () => {
+    const cwd = await mkdtemp(join(tmpdir(), 'omx-update-hydrate-'));
+    const originalLog = console.log;
+    const originalMode = process.env.OMX_AUTO_UPDATE;
+    const logs: string[] = [];
+    let hydrateCalls = 0;
+    let deferredCalls = 0;
+    console.log = (...args: unknown[]) => {
+      logs.push(args.map((arg) => String(arg)).join(' '));
+    };
+    process.env.OMX_AUTO_UPDATE = 'defer';
+
+    try {
+      await withInteractiveTty(async () => {
+        await maybeCheckAndPromptUpdate(cwd, {
+          getCurrentVersion: async () => '0.21.3',
+          fetchLatestVersion: async () => '0.21.3',
+          runGlobalUpdate: () => {
+            throw new Error('no install should run when already up to date');
+          },
+          runDeferredGlobalUpdate: () => {
+            deferredCalls += 1;
+            return { ok: true, stderr: '', logPath: join(cwd, 'update.log') };
+          },
+          runSetupRefresh: async () => {
+            throw new Error('no setup refresh should run when already up to date');
+          },
+          hydrateRuntime: async () => {
+            hydrateCalls += 1;
+            return undefined;
+          },
+        });
+      });
+
+      assert.equal(hydrateCalls, 1);
+      assert.equal(deferredCalls, 0);
+    } finally {
+      if (typeof originalMode === 'string') {
+        process.env.OMX_AUTO_UPDATE = originalMode;
+      } else {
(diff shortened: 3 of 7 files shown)


Yeachan-Heo (commented):
Reviewed head 2b05cf95 against base dev @ b9129c51.

What I verified:
- Read the full diff (7 files). `recoverDeadSessionPointer` no longer returns `unsupported`: on native `unsupported` it falls back to a portable link()+unlink() no-clobber quarantine for regular files (`moveRecoveryPathPortableNoReplace`), with identity checks before/after and EEXIST mapped to collision. Lock-directory quarantine stays fail-closed on unsupported (portable path is files-only) with no checkpoint residue.
- `SessionPointerRecoveryStatus` drops the `unsupported` variant; the refusal site now maps to race/collision. CLI (`session pointer recover`) and types need no further change (generic status rendering).
- Update up-to-date path now calls `hydrateRuntime` (non-fatal) so `omx update` repairs the version-keyed native cache; doctor's unsupported remedy points at `omx update`/`omx setup` and notes pointer recovery works without native. Docs updated.
- No silent success fabrication: portable move verifies destination identity, unlinks source, re-verifies both ends; ambiguous outcomes stay `recovery-required` with residue preserved. No stub/no-op, no dead shim (`probeRecoveryRenameCapability` + tmpdir/mkdtemp imports removed).
- Tests are real assertions: portable recover success (bytes match, canonical gone), portable collision (both files preserved), lock-dir fail-closed with reason match, update hydrate-repair on up-to-date path. No tautologies.

Commands and results (this worktree @ 2b05cf95):
- `npm install` then `npm run build` — success.
- `node --test dist/hooks/__tests__/session.test.js` — 152 pass, 0 fail.
- `node --test dist/cli/__tests__/update.test.js dist/cli/__tests__/doctor-process-identity-readiness.test.js` — 89 pass, 0 fail.
- `npx biome lint` on the 6 touched files — clean. `npx tsc --noEmit` — clean.
- Functional no-native check (atomicRenameNoReplace forced to `unsupported` on a dead-owner fixture): status=recovered, canonical gone, quarantine bytes match — OK.
- PR state: mergeable=true, mergeable_state=clean; CI Status success context present.

Verdict: MERGE_READY.

—

PR #3639, opened 2026-09-07
fix(runtime): resolve omx-runtime from native cache via createRequire

### LINKED ISSUE #3636: [Bug] omx-runtime still not hydrated on macOS arm64 in 0.21.3 — `session pointer recover` returns `unsupported` (follow-up to #3600)

AmatsuZero:
## Summary

`omx doctor` reports `Process identity: native process identity is unavailable (provider unavailable)` on macOS arm64, and `omx session pointer recover` returns `unsupported`. Root cause (with repro detail in #3636): the verified native-cache lookup in `resolveRuntimeBinaryPath()` is silently skipped.

`src/runtime/bridge.ts` lazily `require()`s `../cli/native-assets.js` to compute the cached `omx-runtime` path. The package is published as ESM (`"type": "module"`), so this `require` throws `ReferenceError: require is not defined in ES module scope`. The surrounding `catch { /* best-effort */ }` swallows it, so resolution falls through to the bare `omx-runtime` string (which is not on `PATH`), and `execFileSync` fails → `{ kind: 'unsupported' }`.

The native binary itself is hydrated correctly by `postinstall`; the failure is purely in how it is resolved. (Worth flagging for the discussion in #3636: `find <global install> -name omx-runtime` not finding the binary is expected — it lives under `~/.cache/oh-my-codex/native/<version>/...`, not the install tree — so that observation was a red herring.)

## Changes

- `src/runtime/bridge.ts`: replace the bare `require()` with `createRequire(import.meta.url)`, preserving the lazy-require intent that avoids a static import cycle with `native-assets`.
- `src/runtime/__tests__/bridge.test.ts`: add a regression test that exercises the production branch. Every existing case injects `exists`, and the cache branch is gated on `!options.exists`, so the branch had zero coverage and could regress silently.

## Validation

- [x] `npm run build`
- [x] `npm test` (focused: `node --test dist/runtime/__tests__/bridge.test.js` → 17/17 pass; the new case fails against the unpatched source and passes with the fix)
- [x] `npm run lint`
- [x] `npx tsc --noEmit`
- [x] `omx doctor` on the affected host now reports `[OK] Process identity: native process identity provider is ready`

> Note: I did not run the full `npm test` target end-to-end because it pulls in the heavy `verify:native-agents` / `verify:plugin-bundle` / `verify:prompt-guidance` steps, which are unrelated to this change and are slow in this environment. The unit tests and type checks relevant to the change pass.

## Checklist

- [x] Docs updated (README/DEMO/COVERAGE/AGENTS template) when needed — not needed; change is confined to `src/runtime/**`
- [x] Backward-compatibility impact considered — none: behavior is identical for users who already had a working `PATH` `omx-runtime`; this restores correct behavior for the npm-cache install path.

## Related

Closes #3636


Yeachan-Heo:
Thank you for this fix, @AmatsuZero — you found the real root cause behind #3636.

## SHAs reviewed
- head: `f993c4dc` (branch `fix/runtime-esm-create-require`, after merge of current `dev`)
- base: `dev` @ `070451ee` (includes #3638)
- original head at launch: `01071d70` — I re-fetched and reviewed the current head; the diff is unchanged apart from the `dev` merge.

## Independent reproduction (not trusting the framing)
Our package is `"type": "module"` (`package.json:5`), compiled with `module: NodeNext`. In `src/runtime/bridge.ts`, `resolveRuntimeBinaryPath()` called a bare `require('../cli/native-assets.js')` inside a `try { … } catch { /* best-effort */ }`. I confirmed against a build of the base code that this throws `ReferenceError: require is not defined in ES module scope` (verified directly under `--input-type=module`), which the catch swallows — so the verified native-cache lookup is silently skipped and resolution falls through to the bare `'omx-runtime'` PATH fallback:
- base code + seeded verified cache (`OMX_NATIVE_CACHE_DIR` with matching `.sha256`): resolved to `"omx-runtime"` → **CACHE MISSED (bug reproduced)**
- patched code, same setup: resolved to the seeded cache path → **CACHE HIT (fix works)**

On the shape: I checked for a canonical native-path resolver to reuse and for an import cycle — `src/cli/native-assets.ts` does not import `bridge.ts` (only `platform-command`, `package`, `native-assets/*`), so the lazy boundary is precautionary rather than a live cycle. `createRequire(import.meta.url)` is the minimal correct fix in our ESM conventions, and I kept your patch shape as-is. The already-merged #3638 (portable link/unlink recovery in `session.ts`) is complementary, not redundant: `session.ts:584,633` still calls `resolveRuntimeBinaryPath()` for the fs-rename and process-identity probes, so the cache lookup this PR repairs is still on the live path. No diff reduction needed.

## Regression test
Your new test `prefers the verified native cache on the production path` is a real regression test: it calls `resolveRuntimeBinaryPath()` without an injected `exists` so the production cache branch runs. I proved it fails without the fix (temporarily restored the bare `require`: **16 pass / 1 fail**, the new test failing) and passes with it (**17/17** in `bridge.test.js`, **40/40** across `dist/runtime/__tests__/*.test.js`).

## Commands run and results
- `npm run build` — clean
- `node --test dist/runtime/__tests__/*.test.js` — 40 pass / 0 fail
- `npx biome lint src/runtime/bridge.ts src/runtime/__tests__/bridge.test.ts` — clean
- `npx tsc --noEmit` — clean
- CI on latest head: full `CI` workflow `success`, `CI Status` check `success` (all TS lanes green; Rust/native lanes correctly skipped for a TS-only diff)

## Verdict
**MERGE_READY** — no further changes required. Merging into `dev` now.

—

PR #3653, opened 2026-09-10
fix: harden auth storage, hotswap redaction, and TOML updates

NagyVikt:
## Summary

Forward the seven reproduced correctness/security fixes from NagyVikt/oh-my-codex#1 onto current upstream `dev`. This branch contains only the five fix commits, not the fork's unrelated team-launch commits.

## Changes

- Prevent auth slot names from colliding with `slots.json`, while preserving existing case-distinct accounts whose file identity is demonstrably separate.
- Serialize auth mutation transactions across CLI processes, including quota updates and live account selection. Validate credential file identity inside the lock.
- Redact stderr across arbitrary chunk boundaries, preserving UTF-8 and multiline token fields. After a 64 KiB redaction-buffer overflow, suppress the remaining child stderr rather than risk exposing a token.
- Recover the actual rollout session ID from `session_meta.payload.id`, with a filename UUID fallback.
- Preserve bracket-containing quoted TOML tables and multiline instruction text when reading/updating top-level settings. Reuse the existing generator lexical scanner without changing its behavior.
- Ignore non-executable POSIX PATH candidates, preserving upstream's absent/empty/whitespace PATH behavior and Windows discovery.

## Validation

- Targeted regression tests cover all seven defects, legacy account compatibility, cross-process writes, and redaction overflow.
- Auth tests use temporary homes and synthetic credentials; hotswap checks use fake executables.
- [x] `npm run build`
- [x] `npm run lint`
- [x] 650 focused auth/config/platform/TOML/tmux tests via the repository test runner (0 failures)
- [x] 16 focused CLI reasoning/TOML tests
- [ ] Full `npm test` (not run locally; focused checks and upstream CI are reported separately)
- No real OAuth login, native Windows, or case-insensitive-filesystem validation claimed.

## Compatibility and operational boundaries

- Auth mutex acquisition times out after 10 seconds. Locks left by a killed process are not stolen by age; remove one only after confirming no writer remains.
- Overflow suppression intentionally hides subsequent diagnostics/quota messages for that child.
- No global OMX installation, setup configuration, or dependency changes.
- Original fork PR is retained for traceability; this contribution targets upstream `dev`.

### DIFF
--- a/src/auth/__tests__/config-sessions.test.ts
+++ b/src/auth/__tests__/config-sessions.test.ts
@@ -4,7 +4,7 @@ import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { readAuthConfig } from "../config.js";
-import { findLatestRolloutSession } from "../sessions.js";
+import { extractRolloutSessionId, findLatestRolloutSession } from "../sessions.js";
 
 describe("auth config", () => {
   it("merges project keys over user keys per absent key fallback", async () => {
@@ -27,6 +27,20 @@ describe("auth config", () => {
 });
 
 describe("Codex rollout session heuristic", () => {
+  it("uses session_meta payload identity instead of the timestamped filename", async () => {
+    const home = await mkdtemp(join(tmpdir(), "omx-auth-session-id-"));
+    try {
+      const id = "019c6e27-e55b-73d1-87d8-4e01f1f75043";
+      const file = join(home, `rollout-2026-09-10T10-00-00-${id}.jsonl`);
+      await writeFile(file, JSON.stringify({ type: "session_meta", payload: { id } }) + "\n");
+      assert.equal(await extractRolloutSessionId(file), id);
+      await writeFile(file, "invalid metadata\n");
+      assert.equal(await extractRolloutSessionId(file), id);
+    } finally {
+      await rm(home, { recursive: true, force: true });
+    }
+  });
+
   it("returns newest rollout id by mtime", async () => {
     const home = await mkdtemp(join(tmpdir(), "omx-auth-sessions-"));
     try {
--- a/src/auth/__tests__/hotswap.test.ts
+++ b/src/auth/__tests__/hotswap.test.ts
@@ -79,6 +79,69 @@ describe("auth hotswap pointer abort lifecycle", () => {
     }
   });
 
+  it("still detects a quota line that follows an oversized stderr record", { skip: process.platform === "win32" }, async (t) => {
+    const cwd = await mkdtemp(join(tmpdir(), "omx-hotswap-oversized-quota-"));
+    const home = join(cwd, "home");
+    const runtimeHome = join(cwd, "runtime-codex-home");
+    const binDir = join(cwd, "bin");
+    const codexLog = join(cwd, "codex.log");
+    try {
+      await writeAuthSlot(home);
+      await writeFile(join(home, ".omx", "auth", "second.json"), '{"access_token":"second-secret"}\n');
+      await writeFile(
+        join(home, ".omx", "auth", "slots.json"),
+        JSON.stringify({
+          version: 1,
+          currentSlot: "first",
+          slots: [
+            { slot: "first", createdAt: "now", updatedAt: "now" },
+            { slot: "second", createdAt: "now", updatedAt: "now" },
+          ],
+        }),
+      );
+      await mkdir(binDir, { recursive: true });
+      // A verbose upstream payload overflows the redaction buffer before the
+      // quota marker. Suppressing the rest of stderr would hide the rotation
+      // signal and turn a recoverable quota exit into a plain failure.
+      await writeFile(
+        join(binDir, "codex"),
+        `#!/bin/sh\nprintf 'spawned\\n' >> ${JSON.stringify(codexLog)}\nawk 'BEGIN { while (i++ < 70000) printf "x" ; printf "\\n" }'  >&2\necho 'error: 429 quota exceeded' >&2\nexit 1\n`,
+      );
+      await chmod(join(binDir, "codex"), 0o755);
+      let captured = "";
+      const capture = t.mock.method(process.stderr, "write", (chunk: string | Uint8Array) => {
+        captured += String(chunk);
+        return true;
+      });
+
+      try {
+        await runAuthHotswap({
+          cwd,
+          home,
+          env: { CODEX_HOME: runtimeHome, PATH: `${binDir}:/usr/bin:/bin` },
+          argv: [],
+          lifecycle: lifecycle({
+            prepareCodexHomeForLaunch: async () => ({ codexHomeOverride: runtimeHome }),
+          }),
+        });
+      } finally {
+        capture.mock.restore();
+      }
+
+      assert.equal(await readFile(codexLog, "utf-8"), "spawned\n");
+      // Suppressing the rest of stderr would classify this as a plain failure,
+      // leaving the slot unmarked and skipping rotation entirely.
+      assert.match(captured, /quota detected/);
+      const metadata = JSON.parse(await readFile(join(home, ".omx", "auth", "slots.json"), "utf-8")) as {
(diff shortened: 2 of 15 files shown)


Yeachan-Heo (changes requested):
Independent review at head `9af9f690` (base `dev@00d31f78`). The auth-storage locking, reserved-slot identity recheck inside the lock, quoted/multiline TOML boundary handling, and the POSIX executable-bit check all look correct and well covered by the added tests. One blocking defect in the stderr redactor.

## Blocking: permanent stderr suppression breaks quota rotation

`src/auth/redact.ts` — `createAuthStderrRedactor`:

```ts
const accept = (text: string) => {
  if (dropping) return;
  ...
  if (pending.length > 64 * 1024) {
    emit("[omx auth] oversized stderr record; remaining stderr suppressed\n");
    pending = "";
    dropping = true;
    return;
  }
```

Once `dropping` flips true it is never reset, so every subsequent byte of that child's stderr is discarded for the remainder of the process.

That is not only a logging concern. In `src/auth/hotswap.ts`, `runCodexDirect` builds `stderr` from this same emit callback, and the rotation decision consumes it:

```ts
const authInvalid = isAuthInvalidationError(result.stderr);
const quota = isQuotaError({ status: result.status, signal: result.signal, stderr: result.stderr }, config);
```

So a single oversized record — a large stack trace, a verbose upstream JSON error payload, a debug dump — permanently blinds quota and auth-invalidation detection for that run. Hotswap then reports a plain non-zero exit instead of rotating to the next slot and resuming the rollout session. The failure is silent from the user's perspective beyond one suppression line, and it is triggered by untrusted upstream output, not by anything the user controls.

Safety-wise the trade is also not required. The stated reason for never resuming is that no later newline is provably safe without the buffered prefix. That prefix does not need to be 64 KB: keep a bounded trailing window of the dropped record (e.g. the last 1–2 KB) and evaluate the existing `incompleteSecret` pattern against `window + part` at each newline. When the window ends at a boundary that does not match `incompleteSecret`, resumption is provably as safe as the normal path, and the redactor recovers instead of going permanently deaf.

Please either implement bounded-window resumption or, if permanent suppression is intentional, make `runCodexDirect` mark the captured stderr as truncated and treat quota/auth classification as unknown rather than negative, so rotation is not silently skipped. Also add a test covering "oversized record followed by a quota error line still rotates".

## Non-blocking

- `withAuthStorageLock` uses a bare `mkdir` lockdir with no owner record. A crash between `mkdir` and `rmdir` leaves every later auth operation failing after a 10 s timeout, and recovery is manual. The error message does explain the removal condition, so this is acceptable for now, but writing the owning pid/hostname into the lock directory would let a future check distinguish a dead owner from a live one without stealing on age.
- `resolveTmuxBinaryForPlatform` now takes `existsImpl?: ExistsSyncLike` and forwards `undefined` so the platform-dependent default applies. Correct, but the signature change is easy to misread; a short comment stating that omitting the argument is intentional would help.

Head is currently BEHIND `dev`; a rebase is needed before merge regardless of the fix above.

—

Yeachan-Heo:
Fix-forward pushed to this branch as `94eda9f4` rather than leaving the PR on hold — the implementation is worth keeping and only the redactor blocker stood in the way.

## What changed

`createAuthStderrRedactor` no longer suppresses permanently after a 64 KB record. The reasoning in the original comment was sound — no later newline is safe without the buffered prefix — but the prefix does not need to be the whole record. The boundary pattern is anchored with `\s*$`, so it only ever inspects the tail of the buffer.

- A 256-char `suppressedBoundaryWindow` retains the trailing bytes of the suppressed record.
- Whitespace runs are collapsed when appending to that window, so an unbounded blank tail cannot evict a token-field prefix out of it. `key\s*$` matches a collapsed run identically, which closes the one hole a raw fixed-size window would have had.
- Suppression ends at the first newline whose retained tail does not match `INCOMPLETE_SECRET_PATTERN`. No suppressed byte is ever released, and the notice text is now `oversized stderr record suppressed` since the remainder is no longer discarded.

## Why it mattered

`runCodexDirect` builds `result.stderr` from the same emit callback, and rotation consumes it:

```ts
const authInvalid = isAuthInvalidationError(result.stderr);
const quota = isQuotaError({ status: result.status, signal: result.signal, stderr: result.stderr }, config);
```

So a single verbose upstream error payload — attacker-uncontrolled but also user-uncontrolled — turned a recoverable quota exit into a plain failure: no `markSlotQuota`, no rotation, no resume.

## Tests

Four cases added; all three new behavioral tests fail against the previous redactor and pass now:

- resumption after an oversized record, with the following `quota exceeded` line emitted
- a multiline `refresh_token` field arriving *after* resumption is still redacted to `"[REDACTED]"`
- a blank tail longer than the boundary window never looks safe, so the trailing secret stays suppressed
- `hotswap.test.ts`: codex emits a 70 KB stderr line then `error: 429 quota exceeded`; the run must still log `quota detected` and mark slot `first` exhausted in `slots.json`

`src/auth/__tests__` and `src/utils/__tests__` full sweep: all green, 1 pre-existing skip. `tsc --noEmit` and `biome lint src/auth src/utils` clean.

## Still open

The branch is still **BEHIND** `dev`. I rebased it locally and it applies cleanly with no conflicts, but pushing the rebase to this fork is rejected — it carries `.github/workflows/ci.yml` changes from `dev` and my token has no `workflow` scope. So the rebase has to come from your side or from a base merge at merge time. Nothing in the code needs it; it is purely branch freshness.

The two non-blocking notes from my earlier review are untouched and remain your call: the `.metadata-lock` directory carrying no owner identity, and the `existsImpl?` signature change in `resolveTmuxBinaryForPlatform` reading as accidental.

—

Yeachan-Heo:
이 PR은 제 권한으로는 terminal 상태에 도달할 수 없어 in-repo 브랜치로 이어갑니다: **#3662**.

이유는 코드 문제가 아니라 권한 문제입니다. 이 브랜치는 fork에 있고 `dev` 대비 BEHIND인데, 리베이스를 push하는 것도 GitHub의 `update-branch` API를 호출하는 것도 동일하게 거부됩니다:

```
refusing to allow an OAuth App to create or update workflow
`.github/workflows/ci.yml` without `workflow` scope
```

`dev`가 워크플로 파일을 바꿨기 때문에, fork 브랜치를 최신화하는 모든 경로가 제 토큰 스코프에 막힙니다. 구현 자체는 가치가 있어서 닫아두지 않고 이어갑니다.

**#3662 구성:** NagyVikt 님의 원본 커밋 5개를 authorship 그대로 유지한 채 `dev` 위로 리베이스했고(충돌 없음), 리뷰에서 지적한 redactor 블로커 fix-forward 커밋 1개만 제 것으로 올렸습니다. 구현 공로는 @NagyVikt 님입니다.

여기 head `94eda9f4`에도 동일한 fix-forward가 들어가 있고 CI는 성공 19·건너뜀 9로 통과했습니다. 최신화만 불가능한 상태입니다.

—

Yeachan-Heo:
이 작업은 **#3662로 `dev`에 머지 완료**됐습니다. 머지 커밋 `57d596de`, 09:32 UTC (18:32 KST). CI 성공 19·건너뜀 9로 전부 통과했습니다.

원본 커밋 5개는 authorship 그대로 들어갔습니다. 구현 공로는 @NagyVikt 님이고, 제 커밋은 리뷰에서 지적한 redactor 블로커 fix-forward 하나뿐입니다.

이 PR을 직접 머지하지 못한 이유는 코드가 아니라 권한입니다. fork 브랜치가 `dev` 대비 BEHIND인데, `dev`가 워크플로 파일을 변경한 탓에 리베이스 push도 GitHub `update-branch` API도 `refusing to allow an OAuth App to create or update workflow .github/workflows/ci.yml without workflow scope`로 거부됐습니다. 그래서 in-repo 브랜치로 옮겨 최신화·검증·머지했습니다.

중복 방지를 위해 이 PR은 닫습니다. 내용은 전부 반영됐고 다음 릴리즈에 포함됩니다. 좋은 패치 감사합니다 — 특히 락 안에서 legacy 파일 identity를 재검증한 부분과 multiline TOML 경계 처리가 정확했습니다.

—

PR #3660, opened 2026-09-11
fix(hud): skip stale Team leader pane during reconciliation

wangxingzhen:
## Problem
Team setup/teardown and tmux layout changes can leave a resize/layout hook carrying a leader pane ID that no longer exists. `hud --reconcile-tmux` then targets the stale pane, producing `can't find pane` errors and potentially recreating or resizing HUD panes against invalid topology.

## Fix
- Reuse the existing scoped pane snapshot and verify that the hook's leader pane is still live.
- When the scoped snapshot is unavailable, perform only an ID-only target existence probe; avoid changing the normal full-geometry query sequence.
- Return `skipped_stale_leader_pane` without any HUD split/resize/kill mutation when the leader target has disappeared.
- Add a regression test proving stale hooks do not create, resize, or kill panes.

The fix intentionally leaves hook registration, ownership matching, and argv-separator behavior unchanged.

## Validation
- `npm run build` ✅
- `npm run lint` ✅
- `npm run check:no-unused` ✅
- `node dist/scripts/run-test-files.js dist/hud/__tests__/reconcile.test.js` ✅ (59/59)
- `node dist/scripts/run-test-files.js dist/hud/__tests__/tmux.test.js dist/team/__tests__/tmux-session.test.js dist/hud/__tests__/tmux-split-realtmux.test.js` ✅ (combined HUD/Team suites passed; Team suite 303/303)
- Full local Node suite has unrelated macOS native mutex failures in existing non-HUD tests; CI's focused `cli-core-rest` failure was addressed by preserving the existing normal tmux query sequence.


wangxingzhen:
@codex review

Yeachan-Heo (changes requested):
Independent review at head `df9e59f8` (base `main@cb955b0d`, checks 27 success / 1 skipped). The core idea is right: a layout hook that outlives its Team leader pane must not mutate tmux geometry, and returning a distinct `skipped_stale_leader_pane` status keeps that observable. Two issues before this should land on `main`.

## Blocking: an empty snapshot from a failed tmux call still proceeds to mutate

```ts
const leaderPaneIsMissing = leaderPane === undefined
  && (
    panes.length > 0
    || (panes.length === 0 && targetPaneProbe !== undefined && !targetPaneProbe(currentPaneId))
  );
```

`listCurrentWindowPanes` returns `[]` on three distinct conditions: a `parseCanonicalPaneIdSnapshot` miss, a `parseExactTmuxAuthorityLines` rejection, and the `catch` around the `tmux list-panes` calls. A live tmux window always has at least one pane, so `panes.length === 0` in practice means *the snapshot could not be trusted*, not *the window is empty*.

In that state the guard defers entirely to `targetTmuxPaneExists`, which is a separate `tmux list-panes -t <pane>` call and is also window-agnostic — it resolves the pane wherever it lives. If that one call happens to succeed (or the pane now lives in another window) the probe returns true, `leaderPaneIsMissing` is false, and reconciliation continues with an empty pane list into the create/resize/kill path. That is exactly the layout mutation on untrustworthy state this PR is meant to prevent.

Treat an unusable snapshot as non-actionable on its own: when `panes.length === 0`, skip regardless of the probe. The probe then only needs to answer the narrower question it is good at, and the failure mode collapses to "do nothing, let the next live leader hook reconcile" — the behavior the PR comment already commits to.

## Blocking: production behavior keyed on the presence of a test seam

```ts
const targetPaneProbe = deps.targetPaneExists
  ?? (deps.listCurrentWindowPanes ? undefined : targetTmuxPaneExists);
```

Whether the real tmux probe runs depends on whether a caller injected `listCurrentWindowPanes`. That is a test-only concern leaking into the decision path. Any future caller that injects a pane lister for a legitimate reason — a cached snapshot, a Team-side wrapper — silently loses the probe and changes the staleness verdict, with no test covering the regression.

Default `targetPaneProbe` to `targetTmuxPaneExists` unconditionally and let tests inject `deps.targetPaneExists` explicitly when they want a fake. The new test already passes `listCurrentWindowPanes` with a non-matching pane, so it lands in the `panes.length > 0` branch and does not need the probe at all.

## Test coverage gap

The added test only covers `panes.length > 0` with a non-matching leader. Please also cover:

- empty snapshot from a failed tmux call → `skipped_stale_leader_pane`, no mutation
- `leaderPane.paneDead === true` → `skipped_stale_leader_pane`
- live leader present → normal reconciliation still runs and is not regressed by the new early return

The third one matters most: the guard sits ahead of the orphaned-HUD reclamation block, so a false positive does not just skip one resize, it stops dead-leader HUD cleanup for that submit.

## Nit

The unrelated blank-line removal above `isPaneLiveInStrictGlobalProbe` is unrelated churn in a `main`-targeted fix; dropping it keeps the diff to the behavior change.

`main` merges are owner-controlled, so this is review feedback only — no merge action taken.

—

wangxingzhen:
已按 review 修正并推送 commit `771445f`：

- 空的 pane 快照现在直接按不可验证状态 fail-closed，绝不继续进行 HUD 布局变更；
- 删除跨窗口的 target pane 探针及其测试 seam，避免在不可信快照上继续操作；
- 补充空快照、死亡 leader、正常 live leader 三类回归测试；
- 删除无关空行变更。

本地验证：`npm run build`、`npm run check:no-unused`、`npm run lint`、HUD reconcile 62/62、HUD/Team/tmux 相关测试 303/303 全部通过。

wangxingzhen:
CI 新一轮发现 `cli-core-rest` 中两个原生 hook 测试夹具仍返回不完整的 pane 快照，触发了新的 fail-closed 保护；已在 commit `f8fbcae` 修正测试夹具：

- 为 leader/HUD 返回完整几何与 incarnation 字段；
- 让 `-t %1` 的 scoped list-panes 只返回 leader，避免 authoritative snapshot 误判；
- 保留原有 HUD split 断言。

本地 `build`、`check:no-unused`、`lint`、reconcile 62/62 通过；macOS 原生 mutex 相关测试仍是本机既有环境限制。

wangxingzhen:
Updated the PR as an English-only history and addressed the review feedback.

- Squashed the branch into commit `ca0ccc8`: `fix(hud): skip stale Team leader panes during reconciliation`.
- Empty or untrusted pane snapshots now fail closed without layout mutations.
- Removed the cross-window target-pane probe and its test seam.
- Added regression coverage for empty snapshots, dead leaders, live leaders, and trusted native-hook pane fixtures.
- Local build, no-unused check, lint, and HUD reconciliation tests pass; the focused Team/HUD suites pass locally.

Yeachan-Heo (approved):
Both blockers are resolved at head `ca0ccc86`. Approving; the merge itself stays with the repo owner since this targets `main`.

## Blockers closed

- **Untrusted empty snapshot.** `panes.length === 0` now skips unconditionally, ahead of the missing/dead leader checks. The window-agnostic probe is gone entirely, so a failed `list-panes` can no longer be rescued into a layout mutation by a second lookup that resolves the pane in another window. `targetTmuxPaneExists` was removed rather than left unused — correct call.
- **Production behavior keyed on a test seam.** The `deps.listCurrentWindowPanes ? undefined : targetTmuxPaneExists` branch is gone, so the staleness verdict no longer depends on whether a caller injected a pane lister.

## Test coverage

All four cases I asked for are present in `src/hud/__tests__/reconcile.test.ts`: stale leader, unusable empty snapshot, dead leader, and — the one that matters most — `continues normal reconciliation when the scoped leader is live`, which pins the guard against false positives that would otherwise silently stop dead-leader HUD cleanup.

## On the two CI failures at `771445f0`

`codex-native-hook.test.ts` tests 213 and 215 were fixture artifacts, not a product regression, and the fix on this head is the right one: emit a real full-field pane row instead of a 2-field row. Worth recording why they were green before — the fixture's `[[ "$panes" == *'%9'* ]] && printf ...` exits 1 when no HUD pane exists, and under `set -euo pipefail` that made the whole `list-panes` call exit nonzero, so every snapshot read failed and the old probe silently covered for it. Wrapping it in `if ... fi` is what actually makes those fixtures behave like tmux. I had prepared the same change locally and dropped it once this head landed; your version also scopes `-t %1` correctly, which mine did not.

Verified locally against `main` as the baseline: `reconcile.test.ts` 62/62 pass, and `codex-native-hook.test.ts` shows zero new failures relative to `main` (the remaining local failures are environment-dependent and fail identically on the base commit). CI on this head is 27 success / 1 skipped / 0 failures.

The unrelated blank-line churn near `isPaneLiveInStrictGlobalProbe` is still in the diff; not worth another round.

—

Yeachan-Heo:
Terminal disposition: base를 `main` → `dev`로 전환했습니다. `main` 직행 머지는 레포 오너 권한이라 이 PR이 `CLEAN·APPROVED` 상태로 27시간 이상 정지해 있었고, 이 레포의 정상 통합 경로는 `dev`입니다.

- head `ca0ccc86` 기준 CI 전 lane 통과(실패/타임아웃 0), review `APPROVED` 유지
- latest `dev` 기준 branch update 후 재검증 → `dev` 머지로 닫습니다
- `main` 반영은 이후 오너 릴리즈 경로에서 진행됩니다

Thanks for the fix, @wangxingzhen.

—

PR #3668, opened 2026-09-15
fix(notifications): strip complete ANSI CSI sequences

hiSandog:
## Summary

Notification tails retain CSI sequences such as `ESC[?25l` (hide cursor), and those sequences also prevent metadata-only tmux status lines from being filtered. Both notification matchers currently accept only digits/semicolons followed by a letter.

## Changes

- Match CSI parameter, intermediate, and final byte ranges using the pattern already present in the sidecar renderer, retaining existing two-character escape handling.
- Add regression coverage for private modes, colon-separated colors, cursor-style sequences, and bracketed-paste markers, including metadata filtering that preserves real failure output.

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] `node --test dist/notifications/__tests__/formatter.test.js dist/notifications/__tests__/tmux.test.js` — 63 tests pass; the five new cases fail before the fix.
- [x] `git diff --check`
- [ ] `npm test` — full suite not run; validation is scoped to the affected notification helpers.

## Checklist

- [x] Documentation impact considered

Document-refresh: not-needed | Corrects the existing ANSI-stripping and metadata-filtering behavior; no configuration or public interface changes.


Yeachan-Heo (approved):
**MERGE_READY** — approved at head \`107761440cfc8d705a67ddb816ee588b27e8a60a\` (base \`dev\`).

Reviewed:
- \`ANSI_RE\` in \`src/notifications/formatter.ts\` and \`src/notifications/tmux.ts\` now follows the ECMA-48 CSI grammar (\`\x1b[\` + params \`[0-?]*\` + intermediates \`[ -/]*\` + final \`[@-~]\`), so private-mode (\`?25l\`/\`?25h\`), colon-subparameter SGR (\`38:2::255:0:0m\`), intermediate-byte (\`2 q\`) and bracketed-paste (\`200~\`/\`201~\`) sequences are stripped instead of leaking into notification text.
- Regression coverage added in both \`__tests__/formatter.test.ts\` and \`__tests__/tmux.test.ts\` exercising exactly those four previously-leaking forms.
- No behavior change for the already-handled \`[0-9;]*[A-Za-z]\` subset; the new class is a strict superset.

CI at this head: 26 success / 2 skipped, 0 failing. Merge state \`clean\`. Merging to \`dev\`.

—

PR #3682, opened 2026-09-19
fix(hud): silence reconcile jobs and reap orphan HUD panes

NagyVikt:
## Summary

Fix two tmux HUD lifecycle problems: periodic reconciliation failures print long `returned 1` messages into a user's pane, and an owned HUD can outlive the Codex/OMX session that created it.

Targets `dev`.

## Changes

- Launch periodic reconciliation as an ignored-stdio detached Node child, rather than a tmux `run-shell` job. Preserve the owner environment, leader pane, and working directory.
- Keep best-effort POSIX layout-hook reconciliation failures quiet, without changing the explicit `omx hud --reconcile-tmux` exit-status contract.
- Observe the matching session owner's process identity before rendering or reconciling, including while detached. Retain that identity across pointer deletion/replacement at shutdown.
- Stop a dead owner's watcher and atomically remove only the watcher's own pane incarnation, including with `remain-on-exit` enabled. Unknown ownership, standalone watches, foreign sessions, and shell-hosted/reused panes are preserved.
- Add unit and private real-tmux regression coverage for shutdown, pointer removal, owner aliases, indeterminate state, and reconciliation transport.

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] Targeted HUD tests: 92 passed, no skips.
- [x] `SHELL=/bin/bash node dist/scripts/run-test-files.js dist/hud`: 418 passed, no skips.
- [x] Both affected-adjacent CLI lifecycle suites under `SHELL=/bin/bash`: 47 passed, no skips (`detached-launch-diagnostics-3578.test.js` and `detached-leader-teardown.test.js`).
- [x] Red/green check against unmodified `dev` (`5a00100f`): both owner-exit integration cases fail before the fix; the reconciliation transport regression also fails before the fix.
- [ ] `npm test`: attempted, not a green full-suite run. The local Fish environment produced eight failures in `detached-launch-diagnostics-3578.test.js` and six in `detached-leader-teardown.test.js`; the run was stopped after it continued into slow doctor tests. All eight diagnostics failures and two selected teardown failures reproduce on unmodified `dev`. On that same baseline, the diagnostics file passes all 25 tests with `SHELL=/bin/bash`. Remaining full-suite validation is left to CI.
- `omx doctor`: not applicable; no setup/configuration changes.

## Checklist

- [x] Focused change; no deployment, global installation, or active-user-pane cleanup.
- [x] Backward compatibility considered: ownership uncertainty does not authorize pane deletion; explicit reconcile errors remain observable to callers.
- [x] No public configuration or documented workflow changes.

Document-refresh: not-needed | Internal HUD lifecycle correction with regression coverage; no new public configuration.


Yeachan-Heo:
Closing as **superseded by #3683**, not as a rejection — the fix itself is good and is being merged there.

Both PRs touch the identical 6 files for the same defect (orphan HUD watcher + reconcile error overlay), opened 6 minutes apart. #3683 is the later iteration and is strictly better on the real-tmux coverage: it asserts against a recorded tmux command log (`display-message` present, `run-shell` absent) instead of `capture-pane` output, which does not contain tmux's view-mode error overlay — so the assertion in this PR could pass without actually proving the overlay is gone.

Verification evidence is on #3683 (reproduced 4 failures on clean `dev` `5a00100f`, 69/69 pass at `f811a282`). Thanks for the fix — keeping one lane per defect so the dev history stays single-sourced.

—

PR #3683, opened 2026-09-19
fix(hud): stop orphan watchers and silence reconcile failures

NagyVikt:
## Summary

Fix orphaned owned HUD watchers and recurring tmux `hud --reconcile-tmux ... returned 1` error overlays. Target: `dev`.

An owned watcher could continue reconciling after its launcher exited. Redirecting a failed `tmux run-shell -b` command's stdout/stderr does not prevent tmux from displaying its nonzero exit in an active pane.

## Changes

- Check the exact session owner's process liveness before rendering, reconciling, or running authority ticks, including while detached.
- Retain the observed owner across session-pointer deletion/replacement; preserve standalone watches and fail open for missing, foreign, malformed, or indeterminate evidence.
- Stop the watch timer and remove only the watcher's own PID-fenced pane, including with `remain-on-exit` enabled.
- Spawn periodic reconciliation directly with ignored stdio, preserving its cwd/environment and leader pane; make deferred POSIX reconciliation hooks explicitly best-effort.
- Add unit and private real-tmux regressions. The reconciliation test records tmux commands because `capture-pane` alone cannot see tmux's error overlay.

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] `npm run check:no-unused`
- [x] `git diff --check`
- [x] 92 targeted HUD tests passed, with no skips.
- [x] Complete HUD suite: 418 tests across 17 files passed, with no skips (`node dist/scripts/run-test-files.js dist/hud/__tests__`, after the native runtime build).
- [x] Regression checks against unpatched `dev`: both owner-exit cases, periodic reconciliation, and deferred hook handling fail as expected; the patched code passes.
- [ ] `npm test` is not green. The full run was stopped after detached-session integration failures. Multiple failures in `dist/cli/__tests__/detached-launch-diagnostics-3578.test.js` were also reproduced in a separate clean worktree at the same `dev` commit (`5a00100f`). Additional full-suite failures were not investigated in this focused PR.
- `omx doctor`: not applicable; setup/config behavior is unchanged.

## Checklist

- [x] Focused change; no new dependency, CLI flag, config option, or installed-package modification.
- [x] Backward compatibility considered: standalone and unproven-owner watches remain running; cleanup is pane/PID-fenced.
- [x] No documentation changes needed for this internal lifecycle/error-reporting fix.

## Related

Observed two orphan HUD processes repeatedly submitting failing reconciliation jobs and flashing tmux error overlays. No session-wide cleanup is introduced.

PR #3685, opened 2026-09-20
fix(hud): close HUD when its tmux leader pane exits

NagyVikt:
## Summary

Follow up on #3683: closing a tmux leader pane can leave its HUD alive when the launcher process survives or the session pointer has been removed. Process liveness alone does not detect this case.

## Changes

- Check a bounded, server-wide tmux pane snapshot before the existing owner-process check. Close the owned HUD when its leader is missing or dead, including `remain-on-exit`.
- Preserve live leaders moved to another window. Treat empty, malformed, failed, or uncorroborated queries as unknown, not proof of exit.
- Add unit coverage and private-server tmux regressions that close the leader while its launcher stays alive and verify that an unrelated sibling pane survives.

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] `git diff --check`
- [x] Entire HUD suite in an isolated environment: 427 passed, zero failed or skipped.
- [x] Installed-runtime smoke test on a private tmux server: leader removal and `remain-on-exit` both remove only the HUD.
- [ ] `npm test`: attempted, then stopped after detached-session failures; the full suite did not finish green.
  - All four failing cases observed in `detached-launch-diagnostics-3578.test.js` also fail on an untouched `upstream/dev` checkout at `fe8a4297` (the baseline run additionally failed two other cases).
  - `detached-leader-teardown.test.js` also reported failures before the run was stopped; those have not been separately baseline-isolated. No claim is made that all full-suite failures are unrelated.
- `omx doctor`: not applicable; no setup/config changes.

## Checklist

- [x] Focused follow-up to the HUD lifecycle fix; no predictor application changes.
- [x] Documentation impact assessed: internal lifecycle correction, no new configuration or public commands.
- [x] Backward compatibility considered: standalone watchers and uncertain ownership remain protected; existing guarded self-pane removal is unchanged.

## Related

Follow-up to #3683.


Yeachan-Heo:
Reviewed at head `1d6ea9b4` against base `dev` `fe8a4297` in a dedicated clean worktree. Verdict: **APPROVE**.

Independent reproduction on an untouched `origin/dev` (`fe8a4297`) worktree with only this PR's `watch-realtmux.test.ts` copied in:
- **5 pass / 2 fail** — `removes only its HUD, including remain-on-exit (closed-pane)` and `(dead-pane)` both fail, i.e. the leader-pane-exit case this PR targets.
- The new `watch-owner.test.ts` cannot compile on `dev` (it injects the new `execTmuxSync` dep), which is expected for a new seam, so the behavioral repro above is the load-bearing evidence.
- At this PR head, `watch-owner` + `watch-realtmux` + `tmux`: **78 pass / 0 fail**. `tsc --noEmit` clean, `biome lint src/hud` clean. CI at this head: **19/19 success, 0 failure, 0 pending**.

On the change itself:
- Querying `list-panes -a` server-wide rather than the current window is the right call: moving a leader pane to another window is not an exit, and a window-scoped query would have reaped a live HUD.
- Requiring a *complete, well-formed* snapshot (`/^%\d+\t[01]$/` on every row) that also contains this HUD before acting is the correct guard — a partial or garbled tmux response can no longer be read as "leader is gone".
- Treating a failed/throwing tmux query as "no evidence" and falling through to owner-pointer state keeps the existing fail-open bias from #3683 intact. A 1s timeout with `stdio` stderr ignored keeps this off the hot path.
- Scope is still fenced on `OMX_TMUX_HUD_OWNER=1` + session id + distinct leader/hud pane ids, so standalone and manually started watchers are untouched.

Noted and agreed on your `npm test` caveat: the `detached-launch-diagnostics-3578` failures reproduce on untouched `dev`, so they are pre-existing and not a blocker for this PR. Thanks for baselining instead of just asserting it.

Merging to `dev`.

—

